PR #8, opened 2022-10-25
Archiver merge

zschira:
Initial pull request of new combined scraper/archiver repo. These repo was migrated from the [scrapers](https://github.com/catalyst-cooperative/pudl-scrapers) repo, and while some code is derived from that repo, looking at the diff probably won't be particularly helpful.

## Scope
This PR merges the scraper/archer repos into a single code base and workflow. It is capable of downloading data for all data sources, then creating/updating zenodo depositions to archive this data. There is a single CLI for controlling this entire workflow.

### Async/Await
The new implementation utilizes [asyncio](https://docs.python.org/3/library/asyncio.html) to perform much of the downloading/uploading tasks concurrently. Async programming allows for concurrency within a single thread that is quite fast for I/O bound tasks. It does this by implementing it's own run-loop and context switching, so it can run individual tasks until they reach the `await` keyword which indicates that it would normally perform some blocking I/O. Instead, by using special non-blocking I/O libraries the run loop will switch to another task without ever having to block the current thread.

To utilize async programming, this implementation has switched to using `aiohttp` instead of `requests`, which is essentially just an async version of the `requests` library. One big change is that `aiohttp` requires you to use a context manager to perform requests, so it is slightly more verbose.

### Dataset archivers
All archivers inherit from an abstract base class, `AbstractDatasetArchiver`. This class implements several helper methods that are available to individual archivers, as well as the method `create_archive` that manages coordinating the downloads and the interaction with zenodo. Each individual archiver is only required to implement a single abstract method, `get_resources`, which should be a generator that yields awaitable objects that will download a single resource each.


src/pudl_archiver/archiver/eia860.py
```diff
@@ -0,0 +1,32 @@
...
+        for link in await self.get_hyperlinks(BASE_URL, link_pattern):
+            yield self.get_year_resource(link, link_pattern.search(link))
+
+    async def get_year_resource(
+        self, link: str, match: typing.Match
+    ) -> tuple[Path, dict]:
+        """Download zip file."""
+        # Append hyperlink to base URL to get URL of file
+        url = f"{BASE_URL}/{link}"
+        year = match.group(1)
```
bendnorman, line 28: Does it make sense to add an exception handler here if the page structure changes? For example, raise an exception if `year` is None or not numeric.

src/pudl_archiver/archiver/classes.py
```diff
@@ -0,0 +1,165 @@
...
+        async with self.session.get(url, ssl=verify) as response:
+            text = await response.text()
+            parser.feed(text)
+
+        # Filter to those that match filter_pattern
+        hyperlinks = parser.hyperlinks
+        if filter_pattern:
+            hyperlinks = {link for link in hyperlinks if filter_pattern.search(link)}
+
+        return hyperlinks
```
bendnorman, line 143: Does it make sense to raise an exception or warning if the regex search doesn't find any matches? Something like "The archiver couldn't find any hyperlinks that match {pattern}. Make sure your `filter_pattern` is correct or if the structure of the {url} page changed."

src/pudl_archiver/archiver/classes.py
```diff
@@ -0,0 +1,165 @@
...
+from pudl_archiver.frictionless import ResourceInfo
+from pudl_archiver.zenodo.api_client import ZenodoDepositionInterface
+
+MEDIA_TYPES: dict[str, str] = {
+    "zip": "application/zip",
+    "xlsx": "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+    "csv": "text/csv",
+}
+
+ArchiveAwaitable = typing.Generator[typing.Awaitable[tuple[Path, dict]], None, None]
```
bendnorman, line 23: Can you include a docstring describing this type, please? Also, does it make sense to turn this into a class?

src/pudl_archiver/cli.py
```diff
@@ -0,0 +1,140 @@
...
+    """A CLI for the PUDL Zenodo Storage system."""
+    args = parse_main()
+    load_dotenv()
+
+    if args.sandbox:
+        upload_key = os.environ["ZENODO_SANDBOX_TOKEN_UPLOAD"]
+        publish_key = os.environ["ZENODO_SANDBOX_TOKEN_UPLOAD"]
+    else:
+        upload_key = os.environ["ZENODO_SANDBOX_TOKEN_UPLOAD"]
+        publish_key = os.environ["ZENODO_SANDBOX_TOKEN_UPLOAD"]
```
bendnorman, line 106: Are these vars all supposed to be the same?

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -0,0 +1,406 @@
+"""Core routines for archiving raw data packages on Zenodo."""
```
bendnorman, line 1: Can you add args and return descriptions to the method docstrings in the module, please?

bendnorman (commented):
It's looking really good, Zach! I didn't have time to do a super complete review, but I can add more when I'm back online Thursday.

setup.py
```diff
@@ -1,22 +1,21 @@
...
+    name="pudl_archiver",
     version="0.2.0",
     author="PUDL",
-    # Directory to search recursively for __init__.py files defining Python packages
-    packages=find_packages("src"),
-    # Location of the "root" package:
-    package_dir={"": "src"},
     python_requires=">=3.10,<3.11",
     install_requires=[
+        "catalystcoop.pudl @ git+https://github.com/catalyst-cooperative/pudl.git@xbrl_integration",
```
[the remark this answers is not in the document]
zschira: Right now Ferc 6,60 only have data source metadata in this branch, but we should be able to change this shortly once we get `xbrl_integration` merged into dev

src/pudl_archiver/archiver/classes.py
zschira: Good call, added the message. This should give a warning for the comment above too.

src/pudl_archiver/archiver/classes.py
zschira: I've added a docstring and used a NamedTuple to wrap the inner object. I'm not sure if there's a great way wrap the `Generator[Awaitable` thing though. It is a very confusing type to look at though

src/pudl_archiver/cli.py
zschira: No good catch, fixed now

src/pudl_archiver/zenodo/api_client.py
zschira: Oh yeah I had kinda forgotten to update all these. Fixed now

src/pudl_archiver/archiver/eia860m.py
```diff
@@ -0,0 +1,37 @@
...
+    """EIA-860M archiver."""
+
+    name = "eia860m"
+    month_map = {
+        name.lower(): number for number, name in enumerate(calendar.month_name)
+    }
+
+    async def get_resources(self) -> ArchiveAwaitable:
+        """Download EIA-860M resources."""
+        link_pattern = re.compile(r"([a-z]+)_generator(\d{4}).xlsx")
```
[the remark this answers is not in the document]
zschira: Yeah these will definitely be tested

src/pudl_archiver/archiver/classes.py
bendnorman: Ah the named tuple feels clearer.  Thanks!

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -0,0 +1,469 @@
...
+            Deposition metadata pertaining to the latest version returned by Zenodo api.
+        """
+        url = f"{self.api_root}/deposit/depositions"
+        params = {"q": f'conceptdoi:"{concept_doi}"', "access_token": self.upload_key}
+
+        async with self.session.get(url, params=params) as response:
+            # Deposition should be first element in json response
+            raw_json = await response.json()
+
+            return Deposition(**raw_json[0])
```
bendnorman, line 242: Can we raise an exception here if the first element of `raw_json` doesn't match our expectations?

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -0,0 +1,469 @@
...
+            Updated Deposition object.
+        """
+        for name, resource in resources.items():
+            if name not in self.deposition_files:
+                self.changed = True
+                logger.info(f"Adding {name} to deposition.")
+
+                with open(resource.local_path, "rb") as f:
+                    await self.upload(f, resource.local_path.name)
+            else:
```
bendnorman, line 204: I'm a little confused by this method. So at this point in `create_archive` we have a dictionary of resource names to `ResourceInfo` objects and we need to the add them to the Zenodo deposition. 

`add_files()` iterates through the resources and adds them to the deposition. I don't understand the case this `else` statement is handling. Do the archives occasionally find duplicate resources? I probably just don't understand the Zenodo workflow well enough.

bendnorman: I guess I'm confused as to what a deposition is.

src/pudl_archiver/zenodo/api_client.py
zschira: Added a check for the array length, and I think any other errors will get caught in the pydantic validation of the `Deposition` object

src/pudl_archiver/zenodo/api_client.py
zschira: I added a new docstring that hopefully makes this more clear. Basically the deposition is a versioned container for uploading files to zenodo and new versions contain all of the files from the previous version by default. So, this function will check if the file already exists, then see if it's been changed and update it if that's the case

src/pudl_archiver/zenodo/api_client.py
bendnorman: Oh that's right. Thanks pydantic.

src/pudl_archiver/zenodo/api_client.py
bendnorman: Oh that makes more sense thank you.

bendnorman (approved):

PR #17, opened 2022-12-03
Use app token for auto-merging bot PRs when CI passes.

zaneselvans:
I don't think there's a `tox-pytest` action in here yet, so this won't ever run. But it'll be good to have once that workflow is set up.


jdangerx:
the force-push was just a rebase onto `main` which should fix the pre-commit.ci failure.

jdangerx (approved):
Looks good! This will be nice eventually.

PR #20, opened 2023-01-10
Keep CLI choices in sync with AbstractDatasetArchiver.__subclasses__

jdangerx:
Just some refactoring I did while I was mucking around making sense of the code.

Basically this just looks at `AbstractDatasetArchiver` and grabs all its subclasses - they need to get `import`ed first before they become registered on the base class, though, hence that weird `__import__` thing.

That makes it so that we can just add the Archiver subclasses and not worry about updating the CLI - which is a trivial worry, but maybe a nice quality-of-life thing to have.

If this seems like not a useful thing I'm happy to close it too - this was mainly for my own learning/orientation.

### DIFF
--- a/src/pudl_archiver/cli.py
+++ b/src/pudl_archiver/cli.py
@@ -28,6 +28,7 @@ def all_archivers():
     ]
     module_names = [f"pudl_archiver.{str(p).replace('/', '.')[:-3]}" for p in pyfiles]
     for module in module_names:
+        # AbstractDatasetArchiver won't know about the subclasses unless they are imported
         __import__(module)
     return AbstractDatasetArchiver.__subclasses__()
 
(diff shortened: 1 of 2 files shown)


src/pudl_archiver/cli.py
```diff
@@ -60,36 +66,11 @@ async def archive_dataset(
...
-            case "epacems":
-                archiver = EpaCemsArchiver(session, deposition)
-            case "epacamd_eia":
-                archiver = EpaCamdEiaArchiver(session, deposition)
-            case _:
-                raise RuntimeError("Dataset not supported")
-
+        cls = ARCHIVERS.get(name)
+        if not cls:
+            raise RuntimeError("Dataset not supported")
```
zschira, line 71: Maybe add the dataset name to the message

zschira (commented):
Looks good! Just one small comment!

zschira (approved):

PR #21, opened 2023-01-11
Add dry run

jdangerx:
Add a `--dry-run` mode so that I can more safely test the automated runs.

* move "actually do archiving" logic into `__init__.py` and out of `cli.py` - in preparation for maybe doing some programmatic access later.
* add `dry_run` attribute to the `ZenodoDepositionInterface` so we can check if we're running dry or wet.
* change `ZenodoDepositionInterface` to the "collect a bunch of IO things to do, then do all the IO in one place" pattern to make dry-run checks easier

I could imagine trying to separate the IO and the logic even more, that would help the testing as well - e.g. some sort of `DepositionDiffer` that can take deposition info from anywhere and produces a changeset, then some sort of `ZenodoThingDoer` that produces deposition info and also applies a changeset. In testing you could just pass stuff in and investigate the changesets produced.

### DIFF
--- a/src/pudl_archiver/zenodo/api_client.py
+++ b/src/pudl_archiver/zenodo/api_client.py
@@ -25,6 +25,16 @@ from pudl_archiver.zenodo.entities import (
 logger = logging.getLogger(f"catalystcoop.{__name__}")
 
 
+class _UploadSpec(BaseModel):
+    """Defines an upload that will be done by ZenodoDepositionInterface."""
+
+    source: io.IOBase | Path
+    dest: str
+
+    class Config:
+        arbitrary_types_allowed = True
+
+
 class DatasetSettings(BaseModel):
     """Simple model to validate doi's in settings."""
 
@@ -85,10 +95,8 @@ class ZenodoDepositionInterface:
         self.dry_run = dry_run
         # We accumulate changes in a changeset, then apply - makes dry runs and testing easier.
         # upload takes (source: IOBase | Path, dest: str) tuples; delete just takes the str key to delete.
-        self.changeset: dict[str, tuple[io.IOBase | Path, str] | str] = {
-            "upload": [],
-            "delete": [],
-        }
+        self.uploads: list[_UploadSpec] = []
+        self.deletes: list[str] = []
 
     @classmethod
     async def open_interface(
@@ -189,9 +197,9 @@ class ZenodoDepositionInterface:
             # Ignore content type
             return Deposition(**await response.json(content_type=None))
 
-    def changed(self):
+    def changed(self) -> bool:
         """Whether there are changes between the old deposition and the new one."""
-        return self.changeset["upload"] or self.changeset["delete"]
+        return self.uploads or self.deletes
 
     async def add_files(self, resources: dict[str, ResourceInfo]):
         """Add all downloaded files to zenodo deposition.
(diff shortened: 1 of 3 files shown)


zschira:
> I could imagine trying to separate the IO and the logic even more, that would help the testing as well - e.g. some sort of DepositionDiffer that can take deposition info from anywhere and produces a changeset, then some sort of ZenodoThingDoer that produces deposition info and also applies a changeset. In testing you could just pass stuff in and investigate the changesets produced.

I'm definitely a fan of separating the logic here. I've never liked the logic flow in `api_client.py`, and I think the way it maintains state is error prone and confusing.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -80,7 +81,14 @@ def __init__(
         self.deposition: Deposition = None
         self.new_deposition: Deposition = None
         self.deposition_files: dict[str, DepositionFile] = {}
-        self.changed = False
+
+        self.dry_run = dry_run
+        # We accumulate changes in a changeset, then apply - makes dry runs and testing easier.
+        # upload takes (source: IOBase | Path, dest: str) tuples; delete just takes the str key to delete.
+        self.changeset: dict[str, tuple[io.IOBase | Path, str] | str] = {
```
zschira, line 88: Type annotation here is a little confusing at first, could maybe use a class or alias here. I could also see this being overtaken in a future round of refactoring (like the logic refactor you mentioned), so I don't think it's too big of a deal

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -178,6 +189,10 @@ async def create_deposition(self, data_source_id: str):
             # Ignore content type
             return Deposition(**await response.json(content_type=None))
 
+    def changed(self):
```
zschira, line 192: Return type annotations aren't great in this module right now, but I feel like one would be useful here

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -196,32 +211,43 @@ async def add_files(self, resources: dict[str, ResourceInfo]):
...
             if filename not in resources and filename != "datapackage.json":
-                self.changed = True
-                await self.delete_file(file)
+                self.changeset["delete"].append(file_info)
 
         await self.update_datapackage(resources)
 
+        await self._apply_changeset()
+
+    async def _apply_changeset(self):
```
zschira, line 237: Probably not too big of a deal since this isn't in the public API, but a one sentence docstring could be nice

zschira (approved):
I have a couple little thoughts/suggestions, but overall looks good, so I'm approving!

src/pudl_archiver/zenodo/api_client.py
jdangerx: Yeah, I thought about making, like, an `UploadSpec` namedtuple or something - not a lot of overhead, I'll make it happen.

src/pudl_archiver/zenodo/api_client.py
jdangerx: Def! Will add.

zschira:
> @zschira I'd be happy to pair on that if you're interested!

Yeah definitely! We should chat about design sometime soon

PR #22, opened 2023-01-12
Github Action to run `pudl_archiver` regularly

jdangerx:
This adds a github action that creates the conda env and runs `pudl_archiver`.

It allows for some parallelism in builds - we can define that via the `dataset_groups` setting on the `job.strategy`.

Some things before this becomes fully "production":
* [ ] give it the non-sandbox credentials
* [ ] add the rest of the datasets
* [ ] schedule weekly

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ conda activate pudl-archiver
 A CLI is provided for creating and updating archives. The basic usage looks like:
 
 ```
-pudl_archiver {list_of_datasources}
+pudl_archiver --datasets {list_of_datasources}
 ```
 
 This command will download the latest available data and create archives for each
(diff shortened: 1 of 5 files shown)


jdangerx:
@zschira this is ready for review, too - I think we're still seeing some funky zenodo issues but the actual GH action stuff is sorted.

Once we fix zenodo / get tests sorted / whatever, we can add tweaks here in terms of how we want to parallelize the builds.

README.md
```diff
@@ -50,20 +50,23 @@ conda activate pudl-archiver
...
-If successful, this command will automatically add the new Zenodo DOI to the
-`dataset_doi.yaml` file.
+There are also four optional flags available:
+* `--sandbox`: used for testing. It will only interact with Zenodo's
+  [sandbox](https://sandbox.zenodo.org/) instance.
+* `--initialize`: used for creating an archive for a new dataset that doesn't
+  currently exist on zenodo. If successful, this command will automatically add
+  the new Zenodo DOI to the `dataset_doi.yaml` file.
+* `--dry-run`: used for testing, it ignores all Zenodo write operations.
+* `--all`: shortcut for archiving all datasets that we have defined archivers
```
zaneselvans, line 68: Do the GitHub runners have enough disk to do all of the datasets at once? I think they only have 14GB total, some of which will be used for the OS and software packages.

I guess we could easily break the different datasets out into different scheduled processes so they don't all download everything at the same time.

zaneselvans: Or would the `job.strategy` kick off totally independent runners? In that case I guess we just break them down so each job fits within the given resources of the runner. This will be great!

jdangerx: `job.strategy` kicks off totally different runners!
In theory we could have some clever caching so we only have to spend the runner-minutes on building the env once. But seeing as we're living large on the open-source free minutes that seems kind of low on the list.

jdangerx: That being said - this `--all` functionality turned out to not be immediately useful - I'm happy to rip it out.

zschira (commented):
This all looks good to me. We'll probably want to make this a cron job to run at a desired frequency, but I'm pretty sure that's a really simple change

PR #23, opened 2023-01-13
Improve error logging and handling

jdangerx:
I was running into some Zenodo errors and decided to start sending back the `errors` field in the actual request.

I also noticed that the archiver process doesn't exit with a non-zero status if we swallow the exceptions - so I decided to collect all the exceptions and log them at the end - this lets the archiver process actually fail, basically.

Otherwise, there are a few small bug fixes in here that are basically typos.

### DIFF
--- a/src/pudl_archiver/zenodo/api_client.py
+++ b/src/pudl_archiver/zenodo/api_client.py
@@ -405,19 +405,23 @@ class ZenodoDepositionInterface:
         """
         params = {"access_token": self.upload_key}
         if self.new_deposition.links.bucket:
-            url = f"{self.new_deposition.links.bucket}"
+            url = f"{self.new_deposition.links.bucket}/{filename}"
+            logger.info(f"PUT file {filename} to Zenodo url {url}.")
+            raw_json = await self._check_resp(
+                await self.session.put(url, params=params, data=file)
+            )
+            logger.debug(f"Response json: {raw_json}")
         elif self.new_deposition.links.files:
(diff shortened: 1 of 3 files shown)


src/pudl_archiver/zenodo/api_client.py
```diff
@@ -352,14 +396,19 @@ async def upload(self, file: BinaryIO, filename: str):
...
         elif self.new_deposition.links.files:
-            url = f"{self.new_deposition.links.files}/{filename}"
+            url = f"{self.new_deposition.links.files}"
         else:
             raise RuntimeError("No file or bucket link available for deposition.")
 
-        logger.info(f"Uploading file {filename} to zenodo deposition.")
-        await self.session.put(url, params=params, data=file)
+        logger.info(f"POSTing file {filename} to Zenodo url {url}.")
+        raw_json = await self._check_resp(
```
jdangerx, line 406: This was confusingly silently succeeding before - the [`PUT`](https://developers.zenodo.org/?shell#update27) is on a specific file and renames the file only. the [`POST`](https://developers.zenodo.org/?shell#create24) is to the files/bucket and creates a new file.

This was causing the `publish` to fail because the new deposition was completely empty.

zschira: This is definitely strange, because `publish` worked for me using the `PUT`

zschira (approved):
Mostly just some thoughts to keep in mind while we continue to refactor

PR #24, opened 2023-01-13
Use pyproject.toml instead of setup.py

jdangerx:
Quick lil' PR to test out using modern Python packaging tools.

I experimented with adding `conda-lock` but it got confused by our `catalystcoop.pudl` and `catalystcoop.arelle-mirror` pins, so I abandoned it. Otherwise it's very easy to set up - just `conda-lock -f pyproject.yml` and then later `conda-lock install` will skip all the resolution and use the lockfile. That should also help us with [caching the environment](https://github.com/actions/cache) to speed up our builds later.


pyproject.toml
```diff
@@ -1,6 +1,77 @@
...
+authors = [{name = "PUDL", email = "pudl@catalyst.coop"}]
+requires-python = ">=3.10,<3.11"
+
+dependencies = [
+    "catalystcoop.pudl @ git+https://github.com/catalyst-cooperative/pudl.git@dev",
+    "coloredlogs~=15.0",
+    "factory_boy>=2.12",
+    "feedparser>=6.0",
+    "tqdm>=4.64",
+    "catalystcoop.arelle-mirror==1.3.0",
```
zaneselvans, line 13: No longer required. If we need arelle we can depend on the production release of arelle (see main PUDL repo dependencies)

pyproject.toml
```diff
@@ -1,6 +1,77 @@
+[project]
+name = "pudl_archiver"
+version = "0.2.0"
+authors = [{name = "PUDL", email = "pudl@catalyst.coop"}]
+requires-python = ">=3.10,<3.11"
+
+dependencies = [
+    "catalystcoop.pudl @ git+https://github.com/catalyst-cooperative/pudl.git@dev",
+    "coloredlogs~=15.0",
+    "factory_boy>=2.12",
```
zaneselvans, line 10: Are we still using `factory_boy` or was that only needed by the old scrapers?

pyproject.toml
```diff
@@ -1,6 +1,77 @@
+[project]
+name = "pudl_archiver"
+version = "0.2.0"
+authors = [{name = "PUDL", email = "pudl@catalyst.coop"}]
+requires-python = ">=3.10,<3.11"
+
+dependencies = [
+    "catalystcoop.pudl @ git+https://github.com/catalyst-cooperative/pudl.git@dev",
```
zaneselvans, line 8: I believe this (heinous) dependency only exists because we rely on some metadata that is stored in the PUDL repo for the construction of the metadata that's uploaded to Zenodo.  IIRC, getting that split would be both one of the complexities and one of the benefits of pulling the datastore module out of PUDL into a standalone pacakge.

environment.yml
```diff
@@ -7,6 +7,10 @@ dependencies:
   # Used to set up the environment
   - pip>=21,<23
   - python>=3.10,<3.11
-  - setuptools<66
+  - setuptools<66,>=61.0
+  # fiona is a transitive dependency which needs GDAL. so we install with conda
+  # TODO: we shouldn't have to install all this geo stuff, so once we break the
+  # dependency on `pudl` we should remove this too.
+  - fiona>=1.8
```
zaneselvans, line 14: There have been some big changes in the Python GIS stack since these dependencies were set up. Might try removing `fiona` here and see if everything works just fine without it.

jdangerx: Nah, I added `fiona` because things weren't working without it :[

pyproject.toml
jdangerx: Yeah, it seems that way. It's here for now, and one day we'll get rid of it, but this does continue to work. Once we deal with that I think `conda-lock` might even work nicely.

pyproject.toml
jdangerx: There's some `import arelle` [here](https://github.com/catalyst-cooperative/pudl-archiver/blob/daz/pyproject-toml/src/pudl_archiver/archivers/ferc/xbrl.py#L18) though I bet it's worth giving production arelle a shot.

zschira (approved):
Looks good, I created #30 to move to official arelle package. I'll quickly test that and create a PR if it works

PR #26, opened 2023-01-16
Create dedicated Zenodo interaction class

jdangerx:
This is very much cut-and-paste from the `ZenodoDepositionInterface`, and the code isn't used anywhere outside of the test I just wrote.

I decided to make One Big Test because I didn't want to make a million little depositions on the sandbox environment.

There are a couple logic changes which I'll point out in the actual PR comments.

@zschira - let me know if you'd like to walk through these changes together, +504 is pretty big. Though, again, I think most of this will be pretty familiar to you.

### DIFF
--- a/src/pudl_archiver/depositors/zenodo.py
+++ b/src/pudl_archiver/depositors/zenodo.py
@@ -1,7 +1,7 @@
 """Handle all deposition actions within Zenodo."""
 import json
 import logging
-from typing import BinaryIO
+from typing import BinaryIO, Literal
 
 import aiohttp
 import semantic_version  # type: ignore
@@ -68,7 +68,11 @@ class ZenodoDepositor:
         """Wraps our session requests with some Zenodo-specific error handling."""
 
         async def requester(
-            method: str, url: str, log_label: str, parse_json: bool = True, **kwargs
+            method: Literal["GET", "POST", "PUT", "DELETE"],
+            url: str,
+            log_label: str,
+            parse_json: bool = True,
+            **kwargs,
         ) -> dict | aiohttp.ClientResponse:
             """Make requests to Zenodo.
 
@@ -264,7 +268,7 @@ class ZenodoDepositor:
         deposition: Deposition,
         target: str,
         data: BinaryIO,
-        force_api: str | None = None,
+        force_api: Literal["bucket", "files"] | None = None,
     ) -> None:
         """Create a file in a deposition.
 
--- a/src/pudl_archiver/depositors/zenodo.py
+++ b/src/pudl_archiver/depositors/zenodo.py
@@ -36,11 +36,7 @@ class ZenodoClientException(Exception):
 
 
 class ZenodoDepositor:
-    """Deposition actions within Zenodo.
-
-    * manipulate depositions & their versions
-    * manipulate files within depositions
-    """
+    """Act on depositions & deposition files within Zenodo."""
 
     def __init__(
         self,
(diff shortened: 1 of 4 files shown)


src/pudl_archiver/depositors/zenodo.py
```diff
@@ -0,0 +1,327 @@
...
+
+        This deposition must either be un-published or in the editing state.
+
+        Args:
+            deposition: the deposition you want to discard.
+
+        Returns:
+            None if success.
+        """
+        url = deposition.links.discard
```
jdangerx, line 234: I implemented this seemingly very straightforward request, but immediately got a `500` for my trouble. Not sure if we should keep this around or not.

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -0,0 +1,319 @@
...
+        params = {"q": f'conceptdoi:"{concept_doi}"'}
+
+        response = await self.request(
+            "GET",
+            url,
+            f"Query depositions for {concept_doi}",
+            params=params,
+            headers=self.auth_write,
+        )
+        if len(response) > 1 and not self.sandbox:
```
jdangerx, line 149: Since the concept DOI is defined as "always pointing at the latest deposition" we can just always get the most recent one. We should still log if the response is plural though.

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -0,0 +1,319 @@
...
+    def _make_requester(self, session):
+        """Wraps our session requests with some Zenodo-specific error handling."""
+
+        async def requester(
+            method: str, url: str, log_label: str, parse_json: bool = True, **kwargs
+        ) -> dict | aiohttp.ClientResponse:
+            """Make requests to Zenodo.
+
+            Args:
+                method: HTTP method - "GET", "POST", etc.
```
zschira, line 76: Could be nice to use an `Enum` or `Literal` to constrain this

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -0,0 +1,319 @@
...
+        """Delete a file from a deposition.
+
+        Args:
+            deposition: the deposition you are applying this change to.
+            target: the filename of the file you want to delete.
+
+        Returns:
+            None if success.
+        """
+        candidates = [f for f in deposition.files if f.filename == target]
```
zschira, line 244: Is it possible to have more than one match here?

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -0,0 +1,319 @@
...
+            headers=self.auth_write,
+        )
+        return response
+
+    async def create_file(
+        self,
+        deposition: Deposition,
+        target: str,
+        data: BinaryIO,
+        force_api: str | None = None,
```
zschira, line 267: Another place it might be nice to constrain this option

zschira (approved):
Nothing blocking, just a couple thoughts I had

src/pudl_archiver/depositors/zenodo.py
jdangerx: I don't expect one, but I'm wary of Funky Zenodo State so I decided to make sure.

src/pudl_archiver/depositors/zenodo.py
jdangerx: Woah, didn't know about `Literal` ! Nice.

Issue #32, opened 2023-01-20
If dataset shrinks, we fail when updating datapackage


jdangerx:
I simulated "a bunch of files go offline" by only returning one hyperlink from `get_hyperlinks`:

```diff
diff --git a/src/pudl_archiver/archivers/classes.py b/src/pudl_archiver/archivers/classes.py
index 1883d0c..35b85bc 100644
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -157,7 +157,7 @@ class AbstractDatasetArchiver(ABC):
                 f"Make sure your filter_pattern is correct or if the structure of the {url} page changed."
             )

-        return hyperlinks
+        return list(hyperlinks)[:1]

     async def create_archive(self):
         """Download all resources and create an archive for upload.
```

Then when I ran  `pudl_archiver --sandbox --datasets eia861` I got:
```
[2023-01-20 14:10:49 [    INFO] catalystcoop.pudl_archiver.zenodo.api_client:317 GET https://sandbox.zenodo.org/api/deposit/depositions - get
deposition
2023-01-20 14:10:54 [    INFO] catalystcoop.pudl_archiver.zenodo.api_client:197 GET https://sandbox.zenodo.org/api/deposit/depositions/1149408/
files
2023-01-20 14:10:57 [    INFO] catalystcoop.pudl_archiver.archivers.classes:75 Archiving eia861
2023-01-20 14:11:07 [    INFO] catalystcoop.pudl_archiver.archivers.classes:176 Downloaded /var/folders/7q/28j531_d151fn0d3831ql0c00000gr/T/tmp
5w_m4p_9/eia861-1999.zip.
2023-01-20 14:11:07 [    INFO] catalystcoop.pudl_archiver.zenodo.api_client:259 Adding eia861-1999.zip to deposition.
2023-01-20 14:11:07 [    INFO] catalystcoop.pudl_archiver.zenodo.api_client:439 Creating new datapackage.json for eia861
2023-01-20 14:11:07 [    INFO] catalystcoop.pudl_archiver.zenodo.api_client:443 GET https://sandbox.zenodo.org/api/deposit/depositions/1149408/
files - getting file list for datapackage
Encountered exceptions, showing traceback for last one: ["('eia861', KeyError('eia861-2006.zip'))"]
Traceback (most recent call last):
[...]

PR #33, opened 2023-01-20
Apply changes to new deposition before datapackage update

jdangerx:
As discussed in #32 we need to apply changes to the deposition before refreshing the `datapackage.json` - this lets us have an up-to-date view of what's in the deposition, what the remote URLs are, etc.

### DIFF
--- a/src/pudl_archiver/zenodo/api_client.py
+++ b/src/pudl_archiver/zenodo/api_client.py
@@ -122,6 +122,7 @@ class ZenodoDepositionInterface:
         # upload takes (source: IOBase | Path, dest: str) tuples; delete just takes the str key to delete.
         self.uploads: list[_UploadSpec] = []
         self.deletes: list[str] = []
+        self.changed = False
 
     @classmethod
     async def open_interface(
--- a/src/pudl_archiver/zenodo/api_client.py
+++ b/src/pudl_archiver/zenodo/api_client.py
@@ -232,10 +232,6 @@ class ZenodoDepositionInterface:
             # Ignore content type
             return Deposition(**await self._check_resp(response, content_type=None))
 
-    def changed(self) -> bool:
-        """Whether there are changes between the old deposition and the new one."""
-        return self.uploads or self.deletes
-
     async def add_files(self, resources: dict[str, ResourceInfo]):
         """Add all downloaded files to zenodo deposition.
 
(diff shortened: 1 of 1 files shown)


src/pudl_archiver/zenodo/api_client.py
```diff
@@ -293,12 +290,15 @@ async def _apply_changes(self):
             return
         for file_info in self.deletes:
             await self.delete_file(file_info)
+        self.deletes = []
```
jdangerx, line 293: We have to reset these so that the changes don't get double-applied.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -232,10 +232,6 @@ async def create_deposition(self, data_source_id: str):
             # Ignore content type
             return Deposition(**await self._check_resp(response, content_type=None))
 
-    def changed(self) -> bool:
```
jdangerx, line 235: With the "clear pending changes after they're applied" update below, this became "are there outstanding changes to apply to the deposition" and not "were there ever any changes."

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -388,7 +388,7 @@ async def delete_file(self, file: DepositionFile):
             file: DepositionFile metadata pertaining to file to be deleted.
         """
         logger.info(f"DELETE file {file.filename} from {file.links.self}")
-        self._check_resp(
+        await self._check_resp(
```
jdangerx, line 391: Haha, oops. :sweat_smile:

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -388,10 +388,8 @@ async def delete_file(self, file: DepositionFile):
             file: DepositionFile metadata pertaining to file to be deleted.
         """
         logger.info(f"DELETE file {file.filename} from {file.links.self}")
-        self._check_resp(
-            await self.session.delete(
-                file.links.self, params={"access_token": self.upload_key}
-            )
+        await self.session.delete(
```
jdangerx, line 391: `_check_resp` assumes a JSON response, which this DELETE request doesn't have (it returns an empty `text/html`). Once we use the Depositor for this we can just pass `parse_json=False` and that should handle this.

zschira (approved):
Looks good!

PR #34, opened 2023-01-21
Remove Zenodo from the ZenodoDepositionInterface

jdangerx:
This removes all direct Zenodo access from `api_client.py` - which means we should probably reorganize the code now that the code in the `zenodo` directory doesn't directly deal with Zenodo anymore.

There's still plenty of clean-up to do with, e.g., collapsing `ZenodoClient` and `ZenodoDepositionInterface` into one `DepositionTwiddler` class or something better-named - but I think that's the next step here.

### DIFF
--- a/src/pudl_archiver/depositors/zenodo.py
+++ b/src/pudl_archiver/depositors/zenodo.py
@@ -156,13 +156,21 @@ class ZenodoDepositor:
             )
 
         latest_deposition = Deposition(**sorted(response, key=lambda d: d["id"])[-1])
-        full_deposition_json = await self.request(
+        return await self.get_record(latest_deposition.id_)
+
+    async def get_record(self, rec_id: int) -> Deposition:
+        """Get a deposition by its record ID directly instead of through concept.
+
+        Args:
+            rec_id: The record ID of the deposition you would like to get.
+        """
+        response = await self.request(
             "GET",
-            f"{url}/{latest_deposition.id_}",
-            log_label=f"Get freshest data for {latest_deposition.id_}",
+            f"{self.api_root}/deposit/depositions/{rec_id}",
+            log_label=f"Get deposition for {rec_id}",
             headers=self.auth_write,
         )
-        return Deposition(**full_deposition_json)
+        return Deposition(**response)
 
     async def get_new_version(self, deposition: Deposition) -> Deposition:
         """Get a new version of a deposition.
--- a/src/pudl_archiver/frictionless.py
+++ b/src/pudl_archiver/frictionless.py
@@ -1,4 +1,5 @@
 """Core routines for frictionless data package construction."""
+from collections.abc import Iterable
 from datetime import datetime
 from pathlib import Path
 
@@ -78,7 +79,10 @@ class DataPackage(BaseModel):
 
     @classmethod
     def from_filelist(
-        cls, name: str, files: list[DepositionFile], resources: dict[str, ResourceInfo]
+        cls,
+        name: str,
+        files: Iterable[DepositionFile],
+        resources: dict[str, ResourceInfo],
     ) -> "DataPackage":
         """Create a frictionless datapackage from a list of files and partitions.
 
--- a/src/pudl_archiver/zenodo/api_client.py
+++ b/src/pudl_archiver/zenodo/api_client.py
@@ -271,16 +271,18 @@ class ZenodoDepositionInterface:
             logger.info("Dry run, aborting.")
             return
         for file_info in self.deletes:
-            await self.depositor.delete_file(self.deposition, file_info.filename)
+            await self.depositor.delete_file(self.new_deposition, file_info.filename)
         self.deletes = []
         for upload in self.uploads:
             if isinstance(upload.source, io.IOBase):
                 await self.depositor.create_file(
-                    self.deposition, upload.dest, upload.source
+                    self.new_deposition, upload.dest, upload.source
                 )
             else:
                 with upload.source.open("rb") as f:
-                    await self.depositor.create_file(self.deposition, upload.dest, f)
+                    await self.depositor.create_file(
+                        self.new_deposition, upload.dest, f
+                    )
         self.uploads = []
 
     async def update_datapackage(self, resources: dict[str, ResourceInfo]):
@@ -387,9 +389,10 @@ class ZenodoClient:
             )
             raise e
 
-        deposition = await deposition_interface.depositor.publish_deposition(
-            deposition_interface.new_deposition
-        )
+        if deposition_interface.changed:
+            deposition = await deposition_interface.depositor.publish_deposition(
+                deposition_interface.new_deposition
+            )
 
         if initialize:
             # Get new DOI and update settings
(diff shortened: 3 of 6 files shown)


src/pudl_archiver/frictionless.py
```diff
@@ -78,7 +79,10 @@ class DataPackage(BaseModel):
 
     @classmethod
     def from_filelist(
-        cls, name: str, files: list[DepositionFile], resources: dict[str, ResourceInfo]
+        cls,
+        name: str,
+        files: Iterable[DepositionFile],
```
jdangerx, line 84: `mypy` was complaining about how `files.values()` wasn't a list. Instead of converting it to list I relaxed the type requirements.

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -156,13 +156,21 @@ async def get_deposition(self, concept_doi: str) -> Deposition:
             )
 
         latest_deposition = Deposition(**sorted(response, key=lambda d: d["id"])[-1])
-        full_deposition_json = await self.request(
+        return await self.get_record(latest_deposition.id_)
+
+    async def get_record(self, rec_id: int) -> Deposition:
```
jdangerx, line 161: The direct-access endpoint seems to be more reliably up-to-date than the conceptdoi query.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -25,30 +23,6 @@
 logger = logging.getLogger(f"catalystcoop.{__name__}")
 
 
-class ZenodoClientException(Exception):
```
jdangerx, line 28: Farewell, sweet prince; and flocks of angels sing thee to thy rest.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -110,18 +84,21 @@ def __init__(
...
         self.publish_key = publish_key
 
+        self.depositor = ZenodoDepositor(upload_key, publish_key, session, self.testing)
+
         # Map resource name to resource partitions
         self.resource_parts: dict[str, dict] = {}
 
-        self.deposition: Deposition = None
-        self.new_deposition: Deposition = None
+        self.deposition: Deposition | None = None
```
jdangerx, line 92: More `mypy` stuff.

Separately, I think in the end I'd like these to not be instance variables at all, if I can get away with it - certainly we don't need `self.deposition` - we only use it to get the draft deposition and then we have no use for it anymore. At least we can make these `_private_ish` variables and know that we only reach into them for tests or something.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -191,16 +159,9 @@ async def remote_fileinfo(self, deposition: Deposition):
...
-
-        logger.info(f"GET {url}")
-        async with self.session.get(url, params=params) as response:
-            raw_json = await self._check_resp(response)
-
-        return {file["filename"]: DepositionFile(**file) for file in raw_json}
+        return {f.filename: f for f in deposition.files}
 
-    async def create_deposition(self, data_source_id: str):
+    async def create_deposition(self, data_source_id: str) -> Deposition:
```
jdangerx, line 164: I think as part of the ZDI/ZenodoClient consolidation we could probably inline this.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -280,8 +222,10 @@ async def add_files(self, resources: dict[str, ResourceInfo]):
             if filename not in resources and filename != "datapackage.json":
                 self.deletes.append(file_info)
 
+        self.changed = len(self.uploads or self.deletes) > 0
```
jdangerx, line 225: This is the whole "reload the draft depo before trying to update the datapackage" thing.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -292,136 +236,19 @@ async def _apply_changes(self):
             logger.info("Dry run, aborting.")
             return
         for file_info in self.deletes:
-            await self.delete_file(file_info)
+            await self.depositor.delete_file(self.new_deposition, file_info.filename)
```
jdangerx, line 239: I spent a long time trying to work out why these were giving me 403's - turns out I was targeting the `self.deposition`, which had already been published 🪦

src/pudl_archiver/zenodo/entities.py
```diff
@@ -5,12 +5,22 @@
...
 
-from pydantic import AnyHttpUrl, BaseModel, Field, constr, validator
+from pydantic import AnyHttpUrl, BaseModel, ConstrainedStr, Field, validator
 
 from pudl.metadata.classes import Contributor, DataSource
 
-Doi = constr(regex=r"10\.5281/zenodo\.\d{6,7}")
-SandboxDoi = constr(regex=r"10\.5072/zenodo\.\d{6,7}")
+
+class Doi(ConstrainedStr):
```
jdangerx, line 13: `mypy` hates `constr` but loves `ConstrainedStr`.

jdangerx (commented):
Just some notes to hopefully shed some light on these decisions.

src/pudl_archiver/depositors/zenodo.py
zschira: This definitely seems like a good improvement

src/pudl_archiver/zenodo/api_client.py
zschira: You served us well

src/pudl_archiver/zenodo/api_client.py
zschira: I think this is a good idea, especially given the comment below about trying to delete files from `self.deposition`

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -433,19 +260,14 @@ async def update_datapackage(self, resources: dict[str, ResourceInfo]):
             Updated Deposition.
         """
         # If nothing has changed, don't add new datapackage
-        if not self.changed():
+        if not self.changed:
             return None
 
+        if self.new_deposition is None:
```
zschira, line 266: Curious under what circumstances this condition will be true?

tests/integration/zenodo_depositor_test.py
```diff
@@ -142,6 +142,11 @@ async def test_full_zenodo_flow(depositor, deposition_metadata, upload_key):
     assert latest_deposition.id_ == published_deposition.id_
 
     new_deposition = await depositor.get_new_version(published_deposition)
+    doubly_new_deposition = await depositor.get_new_version(new_deposition)
```
zschira, line 145: who wants a new deposition when you can have _doubly_ new deposition

zschira (approved):
Again, nothing blocking here that I see. Just some thoughts

src/pudl_archiver/zenodo/api_client.py
jdangerx: Mostly I did this because `mypy` was complaining "this could be None, so you should think twice about accessing `self.new_deposition.files`."

I think if we create a ZDI directly, instead of with `ZenodoClient.deposition_interface`, we could in call `update_datapackage` on it before setting `self.new_deposition`.

tests/integration/zenodo_depositor_test.py
jdangerx: > A new deposition isn't cool, you know what's cool? A doubly new deposition.

PR #36, opened 2023-01-26
ZenodoDepositionInterface / ZenodoClient Unification

jdangerx:
This puts the various initialization/cleanup code that was in `ZenodoClient` into `ZenodoDepositionInterface` and simplifies the calling process by putting the high-level business logic in `ZenodoDepositionInterface.run`.

### DIFF
--- a/src/pudl_archiver/__init__.py
+++ b/src/pudl_archiver/__init__.py
@@ -27,21 +27,6 @@ def all_archivers():
 ARCHIVERS = {archiver.name: archiver for archiver in all_archivers()}
 
 
-async def archive_dataset(
-    name: str,
-    zenodo_deposition_interface: ZenodoDepositionInterface,
-    session: aiohttp.ClientSession,
-):
-    """Download and archive dataset on zenodo."""
-    cls = ARCHIVERS.get(name)
-    if not cls:
-        raise RuntimeError(f"Dataset {name} not supported")
-    else:
-        archiver = cls(session, zenodo_deposition_interface)
-    # TODO (daz): instead - pass the archiver into the ZDI, then hit the MF run button.
-    await archiver.create_archive()
-
-
 async def archive_datasets(
     datasets: list[str],
     sandbox: bool = True,
@@ -63,8 +48,14 @@ async def archive_datasets(
         # List to gather all archivers to run asyncronously
         tasks = []
         for dataset in datasets:
-            zenodo_deposition_interface = await ZenodoDepositionInterface.create(
+            cls = ARCHIVERS.get(dataset)
+            if not cls:
+                raise RuntimeError(f"Dataset {dataset} not supported")
+            else:
+                downloader = cls(session)
+            zenodo_deposition_interface = ZenodoDepositionInterface(
                 dataset,
+                downloader,
                 session,
                 upload_key,
                 publish_key,
@@ -74,13 +65,7 @@ async def archive_datasets(
                 sandbox=sandbox,
             )
 
-            tasks.append(
-                archive_dataset(
-                    dataset,
-                    zenodo_deposition_interface,
-                    session,
-                )
-            )
+            tasks.append(zenodo_deposition_interface.run())
 
         results = zip(datasets, await asyncio.gather(*tasks, return_exceptions=True))
         exceptions = [
--- a/src/pudl_archiver/__init__.py
+++ b/src/pudl_archiver/__init__.py
@@ -31,8 +31,6 @@ async def archive_dataset(
     name: str,
     zenodo_deposition_interface: ZenodoDepositionInterface,
     session: aiohttp.ClientSession,
-    initialize: bool = False,
-    dry_run: bool = True,
 ):
     """Download and archive dataset on zenodo."""
     cls = ARCHIVERS.get(name)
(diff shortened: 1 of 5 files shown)


src/pudl_archiver/__init__.py
```diff
@@ -27,26 +27,6 @@ def all_archivers():
 ARCHIVERS = {archiver.name: archiver for archiver in all_archivers()}
 
 
-async def archive_dataset(
```
jdangerx, line 30: This and `create_archive` both got folded into `ZDI.run`.

src/pudl_archiver/__init__.py
```diff
@@ -68,23 +48,24 @@ async def archive_datasets(
...
-            zenodo_client = ZenodoClient(
-                "dataset_doi.yaml",
+            cls = ARCHIVERS.get(dataset)
+            if not cls:
+                raise RuntimeError(f"Dataset {dataset} not supported")
+            else:
+                downloader = cls(session)
+            zenodo_deposition_interface = ZenodoDepositionInterface(
+                dataset,
+                downloader,
```
jdangerx, line 58: Now we pass the downloader into `ZDI`, so all the coordination lives there. Name-wise it doesn't quite make sense with `AbstractDatasetArchiver` probably being a good place to run the orchestration - but I think what *does* make sense is having the "downloader" and the "depositor" both being controlled by some higher being.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -56,35 +55,44 @@ class ZenodoDepositionInterface:
...
         publish_key: str,
-        api_root: str,
+        deposition_settings: Path,
+        create_new: bool = False,
         dry_run: bool = True,
+        sandbox: bool = True,
     ):
         """Prepare the ZenodoStorage interface.
 
+        Do not call this directly, use .create(). There's async stuff that needs to happen
```
jdangerx, line 69: Eep - this is outdated. fixing.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -56,35 +55,44 @@ class ZenodoDepositionInterface:
...
         self.data_source_id = data_source_id
 
-        self.api_root = api_root
         self.session = session
 
         self.upload_key = upload_key
         self.publish_key = publish_key
 
-        self.depositor = ZenodoDepositor(upload_key, publish_key, session, self.testing)
+        self.depositor = ZenodoDepositor(upload_key, publish_key, session, self.sandbox)
```
jdangerx, line 94: We should initialize the Depositor elsewhere so that we remove the Zenodo dependencies in this class - I put a TODO in here in the rename PR.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -250,7 +249,33 @@ async def _apply_changes(self):
                     )
         self.uploads = []
 
-    async def update_datapackage(self, resources: dict[str, ResourceInfo]):
+    def _update_dataset_settings(self, published_deposition):
```
jdangerx, line 252: This just pulled in from the ZenodoClient.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -95,62 +103,41 @@ def __init__(
...
 
-        """
-        interface = cls(
-            data_source_id, session, upload_key, publish_key, api_root, dry_run=dry_run
-        )
+        self.async_initialized = False
 
-        if create_new:
-            interface.deposition = await interface.create_deposition(data_source_id)
+    async def _initialize(self):
```
jdangerx, line 121: This stuff pulled in from the ZenodoClient.

tests/integration/zenodo_test.py
```diff
@@ -143,9 +147,43 @@ async def test_zenodo_workflow(
     mocker,
 ):
     """Test the entire zenodo client workflow."""
-    from pudl_archiver.zenodo.api_client import ZenodoClient
 
-    client = ZenodoClient(test_settings, session, upload_key, publish_key, testing=True)
+    def verify_files(expected, deposition: Deposition):
```
jdangerx, line 151: I got tired of updating this logic in multiple places, hah!

tests/integration/zenodo_test.py
```diff
@@ -143,9 +147,43 @@ async def test_zenodo_workflow(
...
+
+    class TestDownloader(AbstractDatasetArchiver):
+        name = "Test Downloader"
+
+        def __init__(self, resources, **kwargs):
+            super().__init__(**kwargs)
+            self.resources = resources
+
+        async def get_resources(self):
+            async def identity(x):
```
jdangerx, line 182: This seems like an awkward way to `async`-ify values, but I haven't figured out anything nicer.

tests/integration/zenodo_test.py
```diff
@@ -1,7 +1,6 @@
 """Test zenodo client api."""
 import os
 import tempfile
-import time
```
jdangerx, line 4: No more sleep! Our code is woke now.

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -178,23 +165,39 @@ async def create_deposition(self, data_source_id: str) -> Deposition:
...
 
         Returns:
-            Updated Deposition object.
+            Updated Deposition object - if published, then return that; else
+            return the pending deposition version. Distinguishable by `state`
+            field being 'done' and 'unsubmitted', respectively.
         """
+        await self._initialize()
+        resources = await self.downloader.download_all_resources()
+        # TODO (daz): pass around changesets instead of persisting them on the instance, this
```
zschira, line 184: This will be a huge improvement imo!

src/pudl_archiver/zenodo/api_client.py
```diff
@@ -178,23 +165,39 @@ async def create_deposition(self, data_source_id: str) -> Deposition:
...
+        4. If there were updates, publish.
+        5. Update the dataset settings if this was a new deposition.
 
         Returns:
-            Updated Deposition object.
+            Updated Deposition object - if published, then return that; else
+            return the pending deposition version. Distinguishable by `state`
+            field being 'done' and 'unsubmitted', respectively.
         """
+        await self._initialize()
```
zschira, line 182: I also wonder if we could pass around `new_deposition` and `depostion_files` as well as the changes. It feels like with this new flow we can rip out almost all of the state from the deposition orchestrator

tests/integration/zenodo_test.py
zschira: There might be a cleaner way, but I don't really feel like it's worth the effort for a test haha

zschira (approved):
Looking so much cleaner!

PR #37, opened 2023-01-26
Rename now misleadingly-named ZenodoDepositionInterface

jdangerx:
It is the deposition orchestrator now so let's call it that.

I split this out because I was worried doing the rename in the same PR would cause mayhem in the diff view.

Also threw a bunch of TODOs in here to keep notes for the next time we jump into this - I think it's about time to focus on the Dagster stuff.

### DIFF
--- a/src/pudl_archiver/__init__.py
+++ b/src/pudl_archiver/__init__.py
@@ -5,8 +5,9 @@ import pathlib
 
 import aiohttp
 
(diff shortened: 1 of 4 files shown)


jdangerx:
Ah, I fixed the imports but not the `patch` calls. Will get to that this afternoon.

zschira (approved):
Nice!

PR #42, opened 2023-01-26
Small fixes

zschira:
Grouped together several small bug fixes in this PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ dependencies = [
     "coloredlogs~=15.0",
     "feedparser>=6.0",
     "tqdm>=4.64",
-    "catalystcoop.arelle-mirror==1.3.0",
+    "arelle-release>=2.3,<2.4",
     "python-dotenv~=0.21.0",
     "semantic_version>=2.8,<3",
 ]
--- a/src/pudl_archiver/archivers/eia860m.py
+++ b/src/pudl_archiver/archivers/eia860m.py
@@ -35,7 +35,7 @@ class Eia860MArchiver(AbstractDatasetArchiver):
         url = f"https://eia.gov/{link}"
         year = match.group(2)
         month = self.month_map[match.group(1)]
-        download_path = self.download_directory / f"eia860m-{year}-{month}.zip"
+        download_path = self.download_directory / f"eia860m-{year}-{month:02}.zip"
         await self.download_zipfile(url, download_path)
 
         return ResourceInfo(
--- a/src/pudl_archiver/archivers/eia860m.py
+++ b/src/pudl_archiver/archivers/eia860m.py
@@ -31,13 +31,13 @@ class Eia860MArchiver(AbstractDatasetArchiver):
     async def get_year_month_resource(
         self, link: str, match: typing.Match
     ) -> tuple[Path, dict]:
-        """Download zip file."""
+        """Download xlsx file."""
         url = f"https://eia.gov/{link}"
         year = match.group(2)
         month = self.month_map[match.group(1)]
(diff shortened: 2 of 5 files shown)


src/pudl_archiver/archivers/eia860m.py
```diff
@@ -31,13 +31,13 @@ async def get_resources(self) -> ArchiveAwaitable:
...
         self, link: str, match: typing.Match
     ) -> tuple[Path, dict]:
-        """Download zip file."""
+        """Download xlsx file."""
         url = f"https://eia.gov/{link}"
         year = match.group(2)
         month = self.month_map[match.group(1)]
-        download_path = self.download_directory / f"eia860m-{year}-{month}.zip"
-        await self.download_zipfile(url, download_path)
+        download_path = self.download_directory / f"eia860m-{year}-{month:02}.xlsx"
```
zschira, line 38: Was naming files incorrectly, and setting parts wrong.

src/pudl_archiver/zenodo/entities.py
```diff
@@ -13,13 +14,13 @@
 class Doi(ConstrainedStr):
```
zschira, line 14: For some reason, when creating `DatasetSettings` objects I was getting validation errors. Making this explicitly compiled regex patterns fixed the problem for me. Also added unit tests to test this.

src/pudl_archiver/zenodo/entities.py
```diff
@@ -81,6 +82,17 @@ def check_empty_string(cls, doi: str):  # noqa: N805
     def from_data_source(cls, data_source_id: str) -> "DepositionMetadata":
         """Construct deposition metadata object from PUDL DataSource model."""
         data_source = DataSource.from_id(data_source_id)
+        creators = [
```
zschira, line 85: Some new data sources don't have contributors defined in PUDL

pyproject.toml
```diff
@@ -9,7 +9,7 @@ dependencies = [
     "coloredlogs~=15.0",
     "feedparser>=6.0",
     "tqdm>=4.64",
-    "catalystcoop.arelle-mirror==1.3.0",
+    "arelle-release>=2.3,<2.4",
```
jdangerx, line 12: woo!

src/pudl_archiver/archivers/eia860m.py
jdangerx: Nice find :)

src/pudl_archiver/zenodo/entities.py
jdangerx: Makes sense, and then we can just claim credit :)

tests/unit/zenodo_api_test.py
```diff
@@ -0,0 +1,27 @@
+"""Test zenodo API client."""
+import pydantic
+import pytest
+
+from pudl_archiver.zenodo.api_client import DatasetSettings
+
+
+def test_dataset_settings():
```
jdangerx, line 8: non-blocking: These tests seem like they're mostly testing if Pydantic works the way we expect it to - which is fine, but we might not need these? Separately, I think we might want some "test to see if our dataset settings updating logic does the right thing" tests, though those seem a bit more annoying to write.

jdangerx (approved):
🎉

tests/unit/zenodo_api_test.py
zschira: My thoughts were that these sort of test that the regex is working as expected by saying "give it something that looks like a DOI and it's all good. Give it anything else and it fails". Tests that actually test that the settings are handled/updated properly would definitely be more useful though.

Issue #31, opened 2023-01-19
Archive FERC EQR


katie-lamb:
There's high demand for FERC EQR data. Even if we haven't integrated it yet, we can still archive it.

[FERC EQR data](https://www.ferc.gov/power-sales-and-markets/electric-quarterly-reports-eqr).

This is a duplicate issue of [PUDL issue 1414](https://github.com/catalyst-cooperative/pudl/issues/1414), but I'm putting it in the archiver repo.

e-belfer:
I can take a stab at this.

zaneselvans:
The EQR may be a bit weird because it's big and also comes in multiple formats. [GitHub runners only have 14GB of disk](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners) but I think the whole EQR dataset is significantly larger than that (though each individual file won't be).

They don't make the download pages easy to find. The new format (2013-present) [uses a kind of annoying javascripty interface](https://eqrreportviewer.ferc.gov/) but seems to have direct links embedded in it that are formulaic. [The older data](https://www.ferc.gov/download-database) (pre-2013) splits out transaction tables into a bunch of partitions, and then has the non-trasaction tables for all years in its own separate file.

Right now it looks like each quarterly zipped CSV is around 3GB. So even 1 year might exceed the a GitHub runner's disk space.  I wonder how hard / worthwhile it would be to make it possible to serialize the download / upload process from FERC to the runner and then up to Zenodo so the whole dataset isn't ever on disk at the same time.

At least they've finally shut down the janky FTP site and now these can all be downloaded over HTTPS!

e-belfer:
Thanks for the heads up, Zane. Just talked to Zach & Katie about how to structure this. Github runner questions outstanding, I'll start on the archiver part of this, following the FERC2 archiver for some inspiration on how to handle year-to-year file structure changes.

zaneselvans:
IIRC Zenodo also has a 50GB per deposition limit, but like AWS Open Data it's possible to request a higher quota.

jdangerx:
As a shorter workaround than #44, it should be pretty quick to add a command line flag to restrict EQR to certain partitions - and then we could just split the EQR archival process across like 4 different runners.

bendnorman:
Where can I find the archive on Zenodo?

PR #43, opened 2023-01-26
FERC EQR Archiver

e-belfer:
This script archives FERC EQR data from two sources: [2002-2013](https://www.ferc.gov/download-database) and [2014-present](https://eqrreportviewer.ferc.gov/). Technically, the code _should_ work: it produces the expected URLs for each resource, and sends them to the `download_zipfile()` method. However, two issues will frustrate the easy use of this archiver and should be dealt with more creatively.

1. Exceptionally slow download times from FERC. As FERC notes on their website:
> Due to the large size of the database, only three users are allowed to download simultaneously at any time. Because downloading may take a long time, it will likely work best if scheduled during off-peak hours, such as Sundays or at night.
Indeed, running this script generally results in timeout errors. Because of the 3-download limitation, downloads may want to be done serially, rather than in parallel.

2. Files exceed size limitations for GitHub runner, [as noted by Zane](https://github.com/catalyst-cooperative/pudl-archiver/issues/31#issuecomment-1404205318).

This issue is blocked by #44

### DIFF
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -260,13 +260,13 @@ class AbstractDatasetArchiver(ABC):
 
         # Download resources concurrently and prepare metadata
         for resource_chunk in resource_chunks:
+            for resource_coroutine in asyncio.as_completed(resource_chunk):
+                resource_info = await resource_coroutine
+                self.logger.info(f"Downloaded {resource_info.local_path}.")
+                yield str(resource_info.local_path.name), resource_info
+
             # If requested, create a new temporary directory per resource chunk
             if self.directory_per_resource_chunk:
                 tmp_dir = tempfile.TemporaryDirectory()
                 self.download_directory = Path(tmp_dir.name)
                 self.logger.info(f"New download directory {self.download_directory}")
-
-            for resource_coroutine in asyncio.as_completed(resource_chunk):
-                resource_info = await resource_coroutine
-                self.logger.info(f"Downloaded {resource_info.local_path}.")
-                yield str(resource_info.local_path.name), resource_info
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -2,6 +2,7 @@
 import asyncio
 import io
 import logging
+import math
 import tempfile
 import typing
 import zipfile
@@ -53,6 +54,7 @@ class AbstractDatasetArchiver(ABC):
     """An abstract base archiver class."""
 
     name: str
+    concurrency_limit: int | None = None
 
     def __init__(self, session: aiohttp.ClientSession):
         """Initialize Archiver object.
@@ -165,7 +167,9 @@ class AbstractDatasetArchiver(ABC):
 
         return hyperlinks
 
-    async def download_all_resources(self) -> dict[str, ResourceInfo]:
+    async def download_all_resources(
+        self,
+    ) -> typing.Generator[tuple[str, ResourceInfo], None, None]:
         """Download all resources.
 
         This method uses the awaitables returned by `get_resources`. It
@@ -174,11 +178,16 @@ class AbstractDatasetArchiver(ABC):
         # Get all awaitables from get_resources
         resources = [resource async for resource in self.get_resources()]
 
-        # Download resources concurrently and prepare metadata
-        resource_dict = {}
-        for resource_coroutine in asyncio.as_completed(resources):
-            resource_info = await resource_coroutine
-            self.logger.info(f"Downloaded {resource_info.local_path}.")
-            resource_dict[str(resource_info.local_path.name)] = resource_info
+        # Split resources into chunks to limit concurrency
+        chunksize = self.concurrency_limit if self.concurrency_limit else len(resources)
+        resource_chunks = [
+            resources[i * chunksize : (i + 1) * chunksize]
+            for i in range(math.ceil(len(resources) / chunksize))
+        ]
 
-        return resource_dict
+        # Download resources concurrently and prepare metadata
+        for resource_chunk in resource_chunks:
+            for resource_coroutine in asyncio.as_completed(resource_chunk):
+                resource_info = await resource_coroutine
+                self.logger.info(f"Downloaded {resource_info.local_path}.")
+                yield str(resource_info.local_path.name), resource_info
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -186,12 +186,18 @@ class AbstractDatasetArchiver(ABC):
             for i in range(math.ceil(len(resources) / chunksize))
         ]
 
+        if self.concurrency_limit:
+            self.logger.info("Downloading resources in chunks to limit concurrency")
+            self.logger.info(f"Resource chunks: {len(resource_chunks)}")
+            self.logger.info(f"Resources per chunk: {chunksize}")
+
         # Download resources concurrently and prepare metadata
         for resource_chunk in resource_chunks:
             # If requested, create a new temporary directory per resource chunk
             if self.directory_per_resource_chunk:
                 tmp_dir = tempfile.TemporaryDirectory()
                 self.download_directory = Path(tmp_dir.name)
+                self.logger.info(f"New download directory {self.download_directory}")
 
             for resource_coroutine in asyncio.as_completed(resource_chunk):
                 resource_info = await resource_coroutine
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -55,6 +55,7 @@ class AbstractDatasetArchiver(ABC):
 
     name: str
     concurrency_limit: int | None = None
+    directory_per_resource_chunk: bool = False
 
     def __init__(self, session: aiohttp.ClientSession):
         """Initialize Archiver object.
@@ -187,6 +188,11 @@ class AbstractDatasetArchiver(ABC):
 
         # Download resources concurrently and prepare metadata
         for resource_chunk in resource_chunks:
+            # If requested, create a new temporary directory per resource chunk
+            if self.directory_per_resource_chunk:
+                tmp_dir = tempfile.TemporaryDirectory()
+                self.download_directory = Path(tmp_dir.name)
+
             for resource_coroutine in asyncio.as_completed(resource_chunk):
                 resource_info = await resource_coroutine
                 self.logger.info(f"Downloaded {resource_info.local_path}.")
--- a/src/pudl_archiver/archivers/ferc/ferceqr.py
+++ b/src/pudl_archiver/archivers/ferc/ferceqr.py
@@ -14,6 +14,7 @@ class FercEQRArchiver(AbstractDatasetArchiver):
     """FERC EQR archiver."""
 
     name = "ferceqr"
+    concurrency_limit = 1
 
     async def get_resources(self) -> ArchiveAwaitable:
         """Download FERC EQR resources."""
--- a/src/pudl_archiver/archivers/ferc/ferceqr.py
+++ b/src/pudl_archiver/archivers/ferc/ferceqr.py
@@ -1,4 +1,6 @@
 """Download FERC EQR data."""
+import asyncio
+import ftplib  # nosec: B402
 from pathlib import Path
 
 from pudl_archiver.archivers.classes import (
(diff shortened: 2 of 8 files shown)


jdangerx:
Like, a global 3-user throttle? Oy vey. Can you link to the specific docs where they mention that? Not that I don't believe you, I just want to see it with my own eyes. I think we could specifically limit the # of concurrent connections to `1` for EQR - maybe via some sort of command-line flag. I'm thinking something like this in `src/pudl_archiver/__init__.py`:

```diff
diff --git a/src/pudl_archiver/__init__.py b/src/pudl_archiver/__init__.py
index 4eb4439..cf9d20c 100644
--- a/src/pudl_archiver/__init__.py
+++ b/src/pudl_archiver/__init__.py
@@ -47,13 +47,19 @@ async def archive_dataset(
         await archiver.create_archive()
 
 
+DATASETS_LIMITED_CONCURRENCY = {"ferc_eqr"}
+
+
 async def archive_datasets(
     datasets: list[str],
     sandbox: bool = True,
     initialize: bool = False,
     dry_run: bool = True,
+    conns_per_host: int = 20,  # could be parsed as a CLI option
 ):
     """A CLI for the PUDL Zenodo Storage system."""
+    if set(datasets).intersection(DATASETS_LIMITED_CONCURRENCY):
+        conns_per_host = 1
     if sandbox:
         upload_key = os.environ["ZENODO_SANDBOX_TOKEN_UPLOAD"]
         publish_key = os.environ["ZENODO_SANDBOX_TOKEN_PUBLISH"]
@@ -61,7 +67,7 @@ async def archive_datasets(
         upload_key = os.environ["ZENODO_TOKEN_UPLOAD"]
         publish_key = os.environ["ZENODO_TOKEN_PUBLISH"]
 
-    connector = aiohttp.TCPConnector(limit_per_host=20, force_close=True)
+    connector = aiohttp.TCPConnector(limit_per_host=conns_per_host, force_close=True)
     async with aiohttp.ClientSession(
         connector=connector, raise_for_status=False
     ) as session:
@@ -75,7 +81,6 @@ async def archive_datasets(
                 publish_key,
                 testing=sandbox,
             )
-
             tasks.append(
                 archive_dataset(
                     dataset,

```

Re: the file size, if we were to upload individual files to Zenodo instead of downloading the whole dataset and then uploading everything, would that work?

src/pudl_archiver/archivers/ferc/ferceqr.py
```diff
@@ -0,0 +1,68 @@
...
+        """Download all 2002-2013 non-transaction data."""
+        url = "https://eqrdds.ferc.gov/eqrdbdownloads/eqr_nontransaction.zip"
+        download_path = self.download_directory / "ferceqr_nontrans.zip"
+        await self.download_zipfile(url, download_path)
+
+        return ResourceInfo(
+            local_path=download_path, partitions={"data_type": "non-transaction"}
+        )
+
+    async def get_year_dbf(
```
jdangerx, line 42: non-blocking: It might be a little cleaner to split this into two separate functions, one for yearly and one for quarterly - then you aren't, like... checking if certain params were `None`, you can just call the right function. IDK if that feels better to you.

src/pudl_archiver/archivers/ferc/ferceqr.py
```diff
@@ -0,0 +1,68 @@
...
+
+BASE_URL = "https://eqrreportviewer.ferc.gov/"
+
+
+class FercEQRArchiver(AbstractDatasetArchiver):
+    """FERC EQR archiver."""
+
+    name = "ferceqr"
+
+    async def get_resources(self) -> ArchiveAwaitable:
```
jdangerx, line 18: It *might* be easier to test this if you have something that generates all the links, then something else that downloads all of them. Then you could have a test that just checks that all the links that were generated can be downloaded with a `HEAD` request [or something](https://stackoverflow.com/questions/16778435/python-check-if-website-exists).

jdangerx (commented):
Hey, that was quick! Of course there are, well, the other issues you mentioned 😓  - but overall this looks good. One non-blocking code structure comment.

I'm trying to figure out what testing for this might look like - any thoughts @zschira ?

zschira:
Testing the individual archivers is definitely an important/unanswered question. This archiver is following the general pattern of having a method that gets the url for/downloads a single file. We could always mock out `download_zipfile` and check that the URL can be downloaded from there, or we could go with the strategy you presented of splitting it out into two parts.

In general I'd like to have testing that would actually check that the file names and partitions are structured in the way that PUDL expects, so we could catch issues like the one I found with `eia860m`. I'm not sure of an obvious way to get there though.

zschira:
As for the rate limiting and file size issues, I like the idea of being able to set the number of concurrent connections per dataset. I think this will be useful in tuning the rate limiting more generally. I also think downloading one file at a time seems like a good approach. We could probably even do something where we download a file, go through the whole deposition update process (uploads/deletes and update datapackage), but just don't publish the new version, and continue this until we're through the whole process. This might be an easy approach without having to change very much.

jdangerx:
Sounds like we should make a separate ticket for "streaming deposition update in archiver" - I can do that.

In terms of testing the actual data - how did you find that `eia860m` issue?

Could we basically download a small subset of the data for each dataset and validate that? Seems like we'd want to run it through the whole ETL flow, so we could try something like this:

* create all datasets from scratch on sandbox
* update them with as little data as reasonable
* run ETL against them
* validate ETL outputs

I don't actually know the details of what the nightly build does, but is this similar to that?

src/pudl_archiver/archivers/ferc/ferceqr.py
e-belfer: Makes sense to me, implemented.

e-belfer:
So far I've very coarsely tested this by just printing the URLs to terminal and running a spot check to make sure all is expected as I'm developing the code, but in terms of long-term testing the url inputs I could see a few options working here. The download_hyperlinks() method by default only generates hyperlinks which exist on the page, so testing is not an issue for these methods. This leaves download_zipfile() and download_file() as the two methods which take a url as input.

- Adding a [`resp.ok`](https://stackoverflow.com/questions/55126926/get-status-code-with-max-retries-setting-for-requests-head) HEAD check into download_zipfile()/download_file(), which would add this check into all the archivers that take urls as input.
- Adding a separate url_check() method (perhaps, that can be turned off/on), which would precede the download method. Since the functions are built mostly to take singular URLs, it would be best to write this to take one URL as input.

zaneselvans:
Even just having the raw EQR archived somewhere (AWS open data...) that doesn't have a global 3-user limitation will dramatically improve data accessibility 😄 

@zschira I think once an archiver is up and running and working well, we should probably be using it to create new production archives on a schedule, while also being conservative about ensuring that the archiver fails if something unexpected happens, so we don't end up creating polluted (irrevocable) archives.

My guess is that for many of our datasets, almost every new archive will cause the nightly builds to fail, so if we're going to attempt an auto-PR to switch to a new archive, doing it one dataset at a time will be necessary.  Snapshotting the data is and versioning it has value (and is relatively easy) and should make it easy for us to switch over to a new set of DOIs once a quarter or once a month, address any new brokenness, and cut a release.

e-belfer:
Update: this PR is either waiting on the resolution of issue #44 or on the development of an interim workaround. In the meantime, I'll hold off on merging.

zaneselvans:
I guess one potential workaround is that instead of running this archiver on at GitHub runner, we spin up a GCP instance on a schedule (kicked off by a GitHub action on a schedule?) and run it there instead, with plenty of disk and a fast, reliable net connection. Maybe just monthly or quarterly if it ends up being a bear. 🐻

src/pudl_archiver/orchestrator.py
```diff
@@ -34,6 +41,22 @@ class Config:
         arbitrary_types_allowed = True
 
 
+class FileWrapper(io.BytesIO):
```
zschira, line 44: When `aiohttp` times out while uploading a file, it will then close the file, which leads to an error when we retry the upload. I created this wrapper so `aiohttp` can't close the file, which doesn't seem like the best solution, but it was easy, and maybe it's good enough?

src/pudl_archiver/orchestrator.py
```diff
@@ -182,22 +194,39 @@ async def run(self) -> RunSummary | Unchanged:
...
         await self._initialize()
-        resources = await self.downloader.download_all_resources()
-        # TODO (daz): pass around changesets instead of persisting them on the instance, this
-        # makes the ordering/dependency more explicit
-        self._generate_changes(resources)
 
-        # Leave immediately after generating changes if dry_run
+        resources = {}
+        changed = False
+        async for name, resource in self.downloader.download_all_resources():
```
zschira, line 200: Handles one file at a time now

src/pudl_archiver/orchestrator.py
jdangerx: The discussion on [this issue](https://github.com/aio-libs/aiohttp/issues/1907) implies that there's some sort of custom Payload we could pass in, and that this is the preferred method by the aiohttp maintainer. But, it's been a whole major version since then, and the documentation doesn't mention this at all.

[This issue](https://github.com/aio-libs/aiohttp/issues/3654) uses a "monkeypatch the `.close()` function" approach which seems equally janky but a little shorter. 🤷 

This existing option is good enough for me, though we should add some more documentation about this somewhat surprising behavior.

jdangerx (changes requested):
This looks good, but I want to make sure we think through the `apply_changes` interface a bit, and I wanted to flag the `tmp_dir` lifecycle question.

The FileWrapper is fine... I tried to look at how aiohttp wants us to do this, and couldn't figure it out with a few minutes of searching, so I gave up.

Finally I have a small suggestion for the test that (I think) means you get to skip making a whole new test class.

src/pudl_archiver/archivers/classes.py
```diff
@@ -241,11 +246,27 @@ async def download_all_resources(self) -> dict[str, ResourceInfo]:
...
+        # Download resources concurrently and prepare metadata
+        for resource_chunk in resource_chunks:
+            for resource_coroutine in asyncio.as_completed(resource_chunk):
+                resource_info = await resource_coroutine
+                self.logger.info(f"Downloaded {resource_info.local_path}.")
+                yield str(resource_info.local_path.name), resource_info
+
+            # If requested, create a new temporary directory per resource chunk
+            if self.directory_per_resource_chunk:
+                tmp_dir = tempfile.TemporaryDirectory()
```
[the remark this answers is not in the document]
zschira: The goal of this is to limit disc usage for large datasets to avoid overwhelming the github runner disc limits, so it's actually intentionally blowing away the directory after every chunk is finished being processed (`temp_dir` gets overwritten with every iteration and the GC will clean up the old directory). This is all configurable though, so datasets can continue to use the old directory structure if they don't need this.

src/pudl_archiver/orchestrator.py
```diff
@@ -182,22 +194,39 @@ async def run(self) -> RunSummary | Unchanged:
...
+            resources[name] = resource
+            action = self._generate_changes(name, resource)
+
+            # Leave immediately after generating changes if dry_run
+            if self.dry_run:
+                continue
+
+            if action:
+                changed = True
+                await self._apply_changes(action, name, resource)
```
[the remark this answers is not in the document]
zschira: Added a class to encapsulate a change that I think makes things nicer

src/pudl_archiver/orchestrator.py
```diff
@@ -211,62 +240,79 @@ async def run(self) -> RunSummary | Unchanged:
...
+                    self.new_deposition, file_info.filename
+                )
+            case _DepositionAction.CREATE:
+                if resource is None:
+                    raise RuntimeError("Must pass a resource to be uploaded.")
+
+                await self._upload_file(
+                    _UploadSpec(source=resource.local_path, dest=name)
+                )
+            case _DepositionAction.UPDATE:
```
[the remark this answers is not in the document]
zschira: Lmk what you think of my change. I could see it being a bit confusing that the `Update` triggers both `if` statements, but it does remove redundant code.

tests/unit/archive_base_test.py
```diff
@@ -82,6 +82,56 @@ def html_docs():
...
+
+    class TmpDir:
+        def __init__(self):
+            self.call_count = 0
+
+        def __call__(self):
+            self.call_count += 1
+            return Path(f"path{self.call_count-1}")
+
+    tmpdir_mock = mocker.Mock(side_effect=TmpDir())
```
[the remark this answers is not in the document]
zschira: Thanks! Didn't realize you could just pass an iterator as a side effect

src/pudl_archiver/orchestrator.py
zschira: Yeah I found an issue where the `aiohttp` maintainers seemed quite convinced that they should take ownership of a file, and callers should reopen the file, which would requrie a lot of re-engineering on our part, so gonna have to live with a hack I guess.

src/pudl_archiver/archivers/classes.py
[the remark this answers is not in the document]
jdangerx: Sweet, thanks for the explanation!

jdangerx (approved):
New class looks great! tiny nit: now `_apply_changes` only takes one `change`, maybe we should rename 🤷

PR #49, opened 2023-02-01
Archive PHMSA gas annual reports

e-belfer:
This archiver grabs annual reports related to gas from [PHMSA](https://www.phmsa.dot.gov/data-and-statistics/pipeline/gas-distribution-gas-gathering-gas-transmission-hazardous-liquids) and archives them in Zenodo. The files are partitioned by start year, end year and data type.

These datasets are listed in our [Future Datasets documentation](https://catalystcoop-pudl.readthedocs.io/en/dev/data_sources/wip_future.html#phmsa-natural-gas-pipelines) and our [Sloan New Data Plan](https://docs.google.com/document/d/1q9J49sDbwFNsbfk31p_sU0dSKiyq5TiO/edit#heading=h.gjdgxs).

One note: .zip files which are listed as containing data until "present" (e.g.`annual_gas_transmission_gathering_2010_present.zip`) currently actually contain data through 2021. I assign them an end year equivalent to the present year, but we should add a note in future documentation that having a current end_date may not guarantee that the most recent years of data are available.


zschira (approved):
Looks good!

zaneselvans (changes requested):

src/pudl_archiver/archivers/phmsagas.py
```diff
@@ -0,0 +1,68 @@
...
+        if "distribution" in file:
+            data_type = "distribution"
+        elif "transmission" in file:
+            data_type = "transmission"
+        elif "liquefied" in file:
+            data_type = "LNG"
+        elif "hazardous" in file:
+            data_type = "hazliq"
+        elif "underground" in file:
+            data_type = "UNGS"
```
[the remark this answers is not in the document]
e-belfer: Have updated this to just use the existing file names with `snake_case`.

src/pudl_archiver/archivers/phmsagas.py
```diff
@@ -0,0 +1,68 @@
...
+
+        download_path = self.download_directory / f"phmsagas-{file}.zip"
+        await self.download_zipfile(url, download_path)
+
+        return ResourceInfo(
+            local_path=download_path,
+            partitions={
+                "start_year": start_year,
+                "end_year": end_year,
+                "data_type": data_type,
```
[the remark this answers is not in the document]
e-belfer: Now just uses `dataset` and `start_year` as partitions, following the format of the MSHA archiver.

src/pudl_archiver/archivers/phmsagas.py
```diff
@@ -0,0 +1,51 @@
...
+            yield self.get_zip_resource(link, link_pattern.search(link))
+
+    async def get_zip_resource(
+        self, link: str, match: typing.Match
+    ) -> tuple[Path, dict]:
+        """Download zip file."""
+        url = f"https://www.phmsa.dot.gov/{link}"
+        file = str(match.group(1)).replace("-", "_")  # Get file name
+
+        # Set dataset partition
```
zaneselvans, line 36: Could you put an example of the filename pattern that we're matching in the docstring here to make the parsing below more readable?
```py
"""Download zip file.

Filenames generally look like: annual_{dataset}_{start_year}_present.zip

For example: annual_underground_natural_gas_storage_2017_present.zip
"""
```

zaneselvans (commented):
Since we haven't worked with any PHMSA data before, I'm not sure how stable their filenames are going to be over time. With many agencies these files appear to be compiled manually and uploaded by a human being, which means typos and inconsistencies crop up (e.g. the dashes instead of underscores in the one filename).

So as with the MHSA mines data, I think we may want to at least verify that all of the datasets we expect to see are present (with a failure if not), and potentially fail (or maybe just warn) if there are any new ones we didn't expect to see (since we would need to do work downstream to integrate them).

Letting a new temporal block of expected datasets appear (an expected value of `dataset`, but a new `start_year`) seems fine, as they seem to start over with a new file every few years so that's expected behavior.

src/pudl_archiver/archivers/phmsagas.py
```diff
@@ -0,0 +1,51 @@
...
+from pathlib import Path
+
+from pudl_archiver.archivers.classes import (
+    AbstractDatasetArchiver,
+    ArchiveAwaitable,
+    ResourceInfo,
+)
+
+BASE_URL = "https://www.phmsa.dot.gov/data-and-statistics/pipeline/gas-distribution-gas-gathering-gas-transmission-hazardous-liquids"
+current_year = date.today().year
```
zaneselvans, line 14: Is current year used anywhere?

zaneselvans:
I'm such a pain in the ass compared to @zschira.

src/pudl_archiver/archivers/phmsagas.py
e-belfer: This was an artifact of the end_year partition, good catch. Deleted.

e-belfer:
@zaneselvans Not at all, think the added error handling makes complete sense. Added a `ValueError` if expected dataset types aren't found, and a warning if new dataset types are discovered. 

@jdangerx Would just be good to confirm that with the automation of the archivers logged warnings are still easily visible to us / an effective form of communication for something we need to pay attention to? Otherwise I can escalate this from a warning to an error.

e-belfer:
Update: this PR is essentially ready to merge, but is waiting on the resolution of issue #60 and our decision about how best to log errors/warnings as we move to automated archiving.

zschira:
@e-belfer not sure how others feel, but I'd be ok merging this with maybe just some `log.warnings` for now, and addressing #60 in a future PR

zaneselvans:
Yeah I think it's good to merge -- needing to reliably notify us of the outcome of each archiving run seems like a separate issue that's beyond the scope of any one archiver.

zaneselvans (approved):

PR #52, opened 2023-02-02
Archive EIA Thermoelectric Cooling Water data

e-belfer:
This archiver grabs all [EIA Thermoelectric Cooling Water reports](https://www.eia.gov/electricity/data/water/) and archives them in Zenodo. EIA releases both detailed and summary tables. From what I can tell, the summary table just aggregates columns across generators/boilers, and doesn't include additional information. For now, I've partitioned data by year and table_type (detailed or summary), but we could also decide to just archive the detailed files.

An example sandbox output can be found [here](https://sandbox.zenodo.org/record/1158797).

### DIFF
--- a/src/pudl_archiver/archivers/eiawater.py
+++ b/src/pudl_archiver/archivers/eiawater.py
@@ -1,0 +1,43 @@
+"""Download EIA Thermal Cooling Water data."""
+import logging
+import re
+import typing
+from pathlib import Path
+
+from pudl_archiver.archivers.classes import (
+    AbstractDatasetArchiver,
+    ArchiveAwaitable,
+    ResourceInfo,
+)
+
+logger = logging.getLogger(f"catalystcoop.{__name__}")
+BASE_URL = "https://www.eia.gov/electricity/data/water"
+
+
+class Eia923Archiver(AbstractDatasetArchiver):
+    """EIA Thermal Cooling Water archiver."""
+
+    name = "eiawater"
+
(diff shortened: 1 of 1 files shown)


src/pudl_archiver/archivers/eiawater.py
```diff
@@ -0,0 +1,43 @@
...
+    AbstractDatasetArchiver,
+    ArchiveAwaitable,
+    ResourceInfo,
+)
+
+logger = logging.getLogger(f"catalystcoop.{__name__}")
+BASE_URL = "https://www.eia.gov/electricity/data/water"
+
+
+class Eia923Archiver(AbstractDatasetArchiver):
```
zschira, line 17: I think you might've forgot to change the class name

zschira (commented):
I think you just need to change the class name and you'll be good to go! Also, I think archiving both detailed and summary data is good. I don't really see any downside, so might as well

zaneselvans:
Even though I think it should be duplicated, I agree that archiving both isn't a bad idea -- it would let us catch cases where the detailed data has changed but the summaries hasn't, which might indicate errors on EIA's part.

src/pudl_archiver/archivers/eiawater.py
e-belfer: Indeed, typo fixed!

zschira (approved):
Looks good!

PR #54, opened 2023-02-02
Parallelize all the datasets and run them on a schedule

jdangerx:
Only in sandbox, though. If the run works I'll get to work on the production credentials.

### DIFF
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -1,15 +1,32 @@
 ---
 name: run-archiver
 
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "21 4 * * 0" # Sundays 4:21 AM UTC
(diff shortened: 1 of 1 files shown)


jdangerx:
@zschira whee! Look at all the failures! I guess we need to add some more robust retrying logic + maybe fix some other bugs that I haven't dug into yet 😅 

I wonder if we could run some sort of "subsetted" version of this on sandbox as integration test...?

zschira:
Oh no! It looks like most of these are timeouts or rate limiting problems. We're definitely going to need to investigate to make this more robust. I did notice one error I had meant to address in my last round of fixes, but I guess I forgot to get it in there: #55

zschira:
Glad you're getting this retry logic worked out, I can't wait to see this PR with all green checkmarks!

jdangerx:
Closing in favor of #59 .

PR #57, opened 2023-02-06
Retry on asyncio.TimeoutError also

jdangerx:
`aiohttp.ClientSession` throws subclasses of `aiohttp.ClientError` on connection issues, except for timeouts, where it throws `asyncio.TimeoutError`. We would like to retry in all cases.

We ran into this here: https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/4108306582/jobs/7088850155

And there's an open issue for this in `aiohttp` here: https://github.com/aio-libs/aiohttp/issues/7122


zschira (approved):
Retry everything!!

PR #59, opened 2023-02-08
Debug retry logic

jdangerx:
The runs in #54 were still breaking on various retry failures, and I got most of them working.

One last one keeps failing - the FERC1 run. [There seems to be some issue with renaming the downloaded file](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/4127426424/jobs/7130622475#step:5:184), probably worth re-running with debug logging or maybe pulling that "let you SSH in" action in... https://github.com/marketplace/actions/debugging-with-ssh


zschira:
I've seen a similar issue before as this one we're seeing for `ferc1`. This is stemming from `arelle`, which is the library we use for parsing XBRL taxonomies, and it seems to be an inconsistent webcache issue. From experience, trying to debug arelle issues is a huge pain. That being said, `arelle` has recently been acquired by a larger company that seems to be increasing the speed of development/responsiveness quite a bit, so perhaps a good approach could be to open an issue over there, and try and find a temporary solution for ourselves.

I just reran the ferc1 archiver, and it succeeded this time, so I'm also not sure how frequently this will crop up. Perhaps if it's infrequent enough, we could just rerun ferc archivers if they fail?

jdangerx (commented):
Gonna take a few minutes and see if "pass in args/kwargs" feels better than "wrap in lambda". Don't hold your breath.

src/pudl_archiver/utils.py
```diff
@@ -0,0 +1,51 @@
...
+import logging
+from collections.abc import Awaitable, Callable
+from typing import Any
+
+import aiohttp
+
+logger = logging.getLogger(f"catalystcoop.{__name__}")
+
+
+async def retry_async(
```
zschira, line 12: Smart, retry everything!

src/pudl_archiver/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        args: a list of args to pass in to the retried function.
+        kwargs: a dictionary of kwargs to pass into the retried function.
+        retry_count: how many total tries to make.
+        retry_base_s: how many seconds to wait the first time we retry.
+        retry_on: the Exception subclasses to retry on. Defaults to ClientError
+            and TimeoutError because that's what aiohttp.ClientSession.request
+            and the various .read()/.text() methods on aiohttp.ClientResponse
+            can throw.
+    """
+    if args is None:
```
zschira, line 33: Is there a reason to not just do:

```py
args: list = [],
kwargs: dict = {}
```
so these checks aren't needed?

src/pudl_archiver/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        args = []
+    if kwargs is None:
+        kwargs = {}
+    for try_count in range(1, retry_count + 1):
+        # try count is 1 indexed for logging clarity
+        coro = async_func(*args, **kwargs)
+        try:
+            return await coro
+        # aiohttp client can either throw ClientError or TimeoutError
```
zschira, line 42: This comment is maybe unnecessary now that the exceptions to catch are configurable and this is explained in the docstring

zschira (approved):
Smart to add the ability to retry all async funcs!

jdangerx:
OK, so `censusdp1tract` succeeded on [Refactor retry_async interface to be a little more pythonic.](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/4147022397/jobs/7173398298) but not on [Retry on ZenodoClientException](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/4147345364/jobs/7174120120) - I think `censusdp1tract` is just flaky.

src/pudl_archiver/utils.py
jdangerx: We're not mutating the args coming in, so we don't necessarily run into the "[python mutable objects in default args are wonky](https://docs.python-guide.org/writing/gotchas/)" problem, but I'd rather do this and not get bit when sometime in the future we decide "oh, it'd be a great idea to append to `args`".

zschira:
> I think it feels better - @zschira what do you think
Yeah agreed, I feel like the passing args/kwargs pattern is a little more common in python.

> OK, so censusdp1tract succeeded on [Refactor retry_async interface to be a little more pythonic.](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/4147022397/jobs/7173398298) but not on [Retry on ZenodoClientException](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/4147345364/jobs/7174120120) - I think censusdp1tract is just flaky.
I'd guess this is the case. There's only so much retries can do when you're dealing with flaky servers. Maybe we just need _ultimate retry_ of rerunning the entire archiver run on failure, or we could try setting the number of retries higher

src/pudl_archiver/utils.py
zschira: Wow I actually didn't know python did that. That's so annoying!

zschira:
Although it seems to be failing in the same place. I wonder if something is wrong with zenodo deposition?

src/pudl_archiver/utils.py
zaneselvans: This behavior has made for some wildly confusing bugs in the past!

PR #63, opened 2023-02-17
Fix `--dry-run` option

### LINKED ISSUE #62: `--dry-run` CLI option is causing failures
The `--dry-run` CLI option is currently causing failures because it is creating a new draft deposition, not changing that draft deposition, then trying to publish it.

zschira:
This PR fixes the `--dry-run` CLI option, which was previously skipping applying changes to a draft deposition, but then trying to publish that unchanged deposition, which was causing failures.

Closes #62

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,12 @@ conda env create -f environment.yml
 conda activate pudl-cataloger
 ```
 
+## Setting up environment
+API tokens are required to interact with Zenodo. There is one set of tokens for accessing
+the sandbox server, and one for the production server. The archiver tool expects these tokens
+to be set in the following environment variables: `ZENODO_TOKEN_PUBLISH` and `ZENODO_TOKEN_UPLOAD`
+or `ZENODO_SANDBOX_TOKEN_PUBLISH` and `ZENODO_SANDBOX_TOKEN_UPLOAD` for the sandbox server.
+
 ## Usage
 
 A CLI is provided for creating and updating archives. The basic usage looks like:
@@ -123,6 +129,8 @@ pudl_archiver --datasets {new_dataset_name} --initialize
 ## Development
 
 We only have one development specific tool, which is the Zenodo Postman collection in `/devtools`.
+This tool is used for testing and prototyping Zenodo API calls, it is not needed to use the archiver
+tool itself.
 
 To use it:
 
--- a/src/pudl_archiver/orchestrator.py
+++ b/src/pudl_archiver/orchestrator.py
@@ -233,6 +233,11 @@ class DepositionOrchestrator:
                 self.deletes.append(file_info)
 
         self.changed = len(self.uploads or self.deletes) > 0
+        if self.changed:
+            logger.info(f"To delete: {self.deletes}")
+            logger.info(f"To upload: {self.uploads}")
+        else:
+            logger.info("No changes detected.")
 
     async def _apply_changes(self):
         """Actually upload and delete what we listed in self.uploads/deletes."""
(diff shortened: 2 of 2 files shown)


src/pudl_archiver/orchestrator.py
```diff
@@ -232,9 +238,6 @@ async def _apply_changes(self):
         """Actually upload and delete what we listed in self.uploads/deletes."""
         logger.info(f"To delete: {self.deletes}")
         logger.info(f"To upload: {self.uploads}")
```
jdangerx, line 240: It'd be nice to pull this logging out of `_apply_changes` so we can still see "what would have happened if we weren't doing `--dry-run`"
```suggestion
```

src/pudl_archiver/orchestrator.py
```diff
@@ -185,6 +185,12 @@ async def run(self):
         # TODO (daz): pass around changesets instead of persisting them on the instance, this
         # makes the ordering/dependency more explicit
         self._generate_changes(resources)
+
```
jdangerx, line 188: It'd be nice to pull this logging out of _apply_changes so we can still see "what would have happened if we weren't doing --dry-run":

```suggestion

        logger.info(f"To delete: {self.deletes}")
        logger.info(f"To upload: {self.uploads}")
```

jdangerx (approved):
Looks good, tiny change suggested for logging - feel free to do what you want with that suggestion!

README.md
```diff
@@ -45,6 +45,12 @@ conda env create -f environment.yml
 conda activate pudl-cataloger
 ```
 
+## Setting up environment
+API tokens are required to interact with Zenodo. There is one set of tokens for accessing
+the sandbox server, and one for the production server. The archiver tool expects these tokens
+to be set in the following environment variables: `ZENODO_TOKEN_PUBLISH` and `ZENODO_TOKEN_UPLOAD`
+or `ZENODO_SANDBOX_TOKEN_PUBLISH` and `ZENODO_SANDBOX_TOKEN_UPLOAD` for the sandbox server.
+
```
aesharpe, line 53: What's the difference between `publish` and `upload`?

src/pudl_archiver/orchestrator.py
aesharpe: Agreed, also it would be nice to have some more clear logging outputs like `No changes detected; no new uploads` or something like that.

Alternatively, we could add documentation here explaining what some of the expected outputs are. I.e.: 

```
To delete: []
To upload: []
```
means that the archive hasn't changed and nothing new will be uploaded.

README.md
aesharpe: When I ran my update I only needed the `PUBLISH` keys (and I just replaced my upload keys with the word publish...)

zaneselvans: Upload allows you to upload new files to a deposition, but doesn't allow you to publish the deposition -- it could create a draft that a human has to publish.

Publish includes all the permissions of upload, but also allows publication.

src/pudl_archiver/orchestrator.py
zschira: I moved the logs and added a message if no changes are detected

README.md
aesharpe: Are they supposed to be different keys though? I'm confused why my old upload keys worked as publish keys, also I couldn't run the `pudl_archiver` without the PUBLISH key but I could run it without the UPDATE key. Is that because you can run --sandbox without publish or --dry-run without publish?

PR #69, opened 2023-02-22
Delete draft if there were no changes to make

### LINKED ISSUE #67: Archiver workflow sometimes creates unpublished draft depositions
After adding several new archivers to the `run-archivers` workflow in the course of #65 and running all the workflows multiple times, I'm seeing some behavior that confuses me.  The newly added archivers are:

* `eia176`
* `eia923` (didn't make it into the list before)
* `eiawater`
* `mshamines`
* `phmsagas`

My expectation was that if any changes were detected for any of the archived datasets (old or new) a new deposition would be created on Zenodo with the new data in it, and regardless of whether a new archive was published, no draft deposition would remain after the archiving process had completed.

I'm now able to run the `run-archivers` workflow manually via the workflow-dispatch, and all of the actions "succeed" ✅. What happens with the various archivers seemed to differ:

### `censusdp1tract`
* Initially the archiver was failing, for reasons I didn't understand.
* I searched for Census DP1 archives on Zenodo Sandbox and found 2 different lineages of archives, one of which was started in January, and wondered if the new archiver might somehow be incompatible with the old lineage (started in 2020), so I switched the sandbox DOI over to the newer lineage, but this didn't fix it.  It did seem to try and create a v2.0.0 archive though, which it hadn't been doing before.
* Then I looked for draft archives, and saw that there were unpublished drafts for both the old and new lineages. I deleted them, and re-ran the `run-archivers` workflow, and this time it succeeded ✅.
* However, it also created a new unpublished draft, which contains files with identical MD5 checksums as the current published version, but a new version (v2.0.0) and a new DOI.

### `eia176` (first time w/ workflow)
* The archiver runs and succeeds ✅, and (correctly) does not create a new published deposition.
* However, like the `censusdp1tract` archiver, it seems to create an unpublished draft, even though the MD5 checksums for all the files are identical. The [new unpublished draft](https://sandbox.zenodo.org/deposit/1166406) has a different DOI and a new version (v3.0.0) compared to [the previously published deposition](https://sandbox.zenodo.org/record/1166385).

### `eia860`
* Running the archiver did not result in either a newly published archive, or a **new** unpublished draft.
* However, an [old unpublished draft](https://sandbox.zenodo.org/deposit/1118880) appears to be hanging around from October, 2022, with the same MD5 sums, but a new reserved DOI and a new version.

### `eia860m`
* Running the archiver did not result in either a newly published archive, or a **new** unpublished draft.
* However, an [old unpublished draft](https://sandbox.zenodo.org/deposit/1161366) appears to be hanging around from February 6th, 2023, with the same MD5 sums, but a new reserved DOI and a new version.

### `eia861`
* Running the archiver did not result in either a newly published archive, or a **new** unpublished draft.
* However, an [old unpublished draft](https://sandbox.zenodo.org/deposit/1158841) appears to be hanging around from February 2nd, 2023, with the same MD5 sums, but a new reserved DOI and a new version.

### `eia923` (first time w/ workflow)
* Running the archiver did not result in a newly published archive, but **did** create a new unpublished draft.
* The [new unpublished draft](https://sandbox.zenodo.org/deposit/1166382) looks like it has the same MD5 sums, but a new reserved DOI and a new version.

### `eia_bulk_elec`
* Running the archiver did not result in either a newly published archive, or a **new** unpublished draft.
* However, an [old unpublished draft](https://sandbox.zenodo.org/deposit/1159788) appears to be hanging around from February 3rd, 2023, with the same MD5 sums, but a new reserved DOI and a new version.

### `eiawater` (first time w/ workflow)
* Running the archiver did not result in a newly published archive, but **did** create a new unpublished draft.
* The [new unpublished draft](https://sandbox.zenodo.org/deposit/1166364) looks like it has the same MD5 sums, but a new reserved DOI and a new version.

### `epacamd_eia`
* Running the archiver did **not** result in a new published archive **or** a new draft.
* However, [an old draft](https://sandbox.zenodo.org/deposit/1128617) from September 8th, 2022 is still hanging around.
* Also, the unpublished draft only contains the `epacamd_eia.zip` file, and not the `datapackage.json`, and the MD5 sum for the zipfile is different from that of [the most recently published archive](https://sandbox.zenodo.org/record/1103224).

### `epacems`
* Running the archiver did **not** result in a new published archive, even though it definitely should have. The current archive is more than 2 years out of date.
* It also didn't create a new draft, but there was [an old draft](https://sandbox.zenodo.org/deposit/768451) from April, 2021 which contains no uploaded files.
* I deleted the old draft and re-ran the archiving workflow, to see what would happen with the substantial downloads required to make a new CEMS archive.
* With no blocking draft deposition, the CEMS archiver is running much longer. This seems to suggest that the existence of the draft deposition prevented the download of new raw data from EPA entirely, which means it wouldn't have been able to compare the checksums of the raw files with those stored on Zenodo to even figure out if it needed to make a new archive.
* It looks like EPA has taken down the bulk EPA CEMS files entirely. When they archiver ran it got a 404 on the old index page. It looks like the CEMS data may now only be available via an API. EPA has [a repo with API examples](https://github.com/USEPA/cam-api-examples) and the [API documentation](https://www.epa.gov/airmarkets/cam-api-portal)

### `ferc1`
* A new deposition is created every time the archiver is run, because the contents of the data downloaded from RSS is different every time (probably there are some timestamps or something).
* In addition to the new deposition being created with a new version and DOI, a **draft** deposition is created and left unpublished. This draft deposition appears identical to the most recently published version -- the MD5 checksums for the files are the same, and the reserved DOI in the draft deposition is identical to the actual DOI for the new deposition that got published.

### `ferc2`
* As with FERC Form 1, a new deposition is created every time the archiver is run, again with all the XBRL (RSS) derived files having different checksums.
* However, **unlike** the FERC Form 1, no draft version of the deposition is left hanging around. The drafts seem to get published and disappear, as expected.

### `ferc6`
* As with FERC Form 1, [a new deposition](https://sandbox.zenodo.org/record/1166471) is created every time the archiver is run, again with all the XBRL (RSS) derived files having different checksums.
* However, **unlike** the FERC Form 1, no draft version of the deposition is left hanging around. The drafts seem to get published and disappear, as expected.

### `ferc60`
* As with FERC Form 1, [a new deposition](https://sandbox.zenodo.org/record/1166462) is created every time the archiver is run, again with all the XBRL (RSS) derived files having different checksums.
* However, **unlike** the FERC Form 1, no draft version of the deposition is left hanging around. The drafts seem to get published and disappear, as expected.

### `ferc714`
* As with FERC Form 1, [a new deposition](https://sandbox.zenodo.org/record/1166465) is created every time the archiver is run, again with all the XBRL (RSS) derived files having different checksums.
* However, **unlike** the FERC Form 1, no draft version of the deposition is left hanging around. The drafts seem to get published and disappear, as expected.

### `mshamines` (first time w/ workflow)
* No new published version is created.
* An [unpublished draft deposition](https://sandbox.zenodo.org/deposit/1166407) is created, with a new version and DOI, but apparently the same MD5 sums.

### `phmsagas` (first time w/ workflow)
* No new published version is created.
* An [unpublished draft deposition](https://sandbox.zenodo.org/deposit/1166366) is created, with a new version and DOI, but apparently the same MD5 sums.

## Patterns?
* It seems like any time there's a pre-existing unpublished draft, it blocks a new archive from being created.
* It also seems like maybe when no update is required because there's been no change to any of the files, the files are still getting uploaded to Zenodo unnecessarily, creating an unpublished draft (I think if none of the file checksums change, Zenodo may not allow a new version to be published?)
* It seems like maybe the RSS/XBRL sources are working as expected because every single archive requires an update.
* However the FERC Form 1 doesn't seem to follow this pattern.

jdangerx:
I think I got the draft deletion working, which should fix #67 , or at least the part of that issue that is specific to "archiver/zenodo behavior" and not the parts that are about "EPA CEMS is offline" or "FERC RSS feeds are acting up 🤪 "


src/pudl_archiver/depositors/zenodo.py
```diff
@@ -191,18 +199,21 @@ async def get_new_version(self, deposition: Deposition) -> Deposition:
             A new Deposition that is a snapshot of the old one you passed in,
             with a new major version number.
         """
+        # If we try to get a new version of a draft, delete the draft and request a fresh snapshot
         if not deposition.submitted:
             return deposition
```
zschira, line 204: Should this call `delete_deposition` instead of returning the old draft?

src/pudl_archiver/orchestrator.py
```diff
@@ -202,7 +201,8 @@ async def run(self):
                 self._update_dataset_settings(published)
             return published
         else:
-            return self.new_deposition
+            await self.depositor.delete_deposition(self.new_deposition)
+            return self.deposition
```
zaneselvans, line 205: To me it reads like this deletion is happening at the end of the archiver run -- it tries to make a new deposition, and if it finds that nothing has changed, it deletes the draft deposition, meaning that we wouldn't have zombie drafts floating around between updates. That seems like the desired behavior to me.

I might not be understanding how this works, but is it always uploading the new data to a draft, regardless of whether there's been a change? It seems like it should be possible to figure out whether a new version is necessary just based on the list of files and their MD5 checksums -- if it's the same set of files, and they all have the same checksums (the downloaded versions locally compared to the previously uploaded versions on Zenodo) then it seems like creating a new draft deposition and uploading the files could be avoided entirely. The process of uploading the EPA CEMS data to Zenodo was a major source of errors previously (thousands of files, and several GB of data).

src/pudl_archiver/depositors/zenodo.py
zaneselvans: Is the idea here that even though we expect to either delete drafts (if no changes) or publish them (if changes), and thus never expect to see an unpublished draft, if something has gone wrong such that there is a draft hanging around when the archiver runs, it'll delete it and start over with a fresh snapshot of the most recently published deposition?

jdangerx: If we initialize a new deposition, the first version that comes back is unsubmitted. I initially had "delete draft on sight, then ask for a new one" behavior, but we couldn't ask for a new one from a deposition with *no* versions.

src/pudl_archiver/orchestrator.py
zschira: It's only uploading data if there's something new or something that has changed. When it creates a draft it will automatically contains the contents of the previous version, so we will never re-upload something that hasn't changed. Basically we create a new draft to work off of, and we compare downloaded files to this mirror of the previous version. If nothing needs changing we previously would just leave this draft and come back to it next time. This new change will just clean up after it's self and delete the draft so it doesn't hang around.

jdangerx: > To me it reads like this deletion is happening at the end of the archiver run -- it tries to make a new deposition, and if it finds that nothing has changed, it deletes the draft deposition, meaning that we wouldn't have zombie drafts floating around between updates. That seems like the desired behavior to me.

Yeah, that is what is happening.

> I might not be understanding how this works, but is it always uploading the new data to a draft, regardless of whether there's been a change? 

No. It always creates a draft, which is automatically populated with a snapshot of the data. Then the archiver downloads all the data from the source, and creates a changeset based on MD5 comparisons with the draft deposition (see `DepositionOrchestrator._generate_changes`). We then apply that changeset, which may be empty (see `DepositionOrchestrator._apply_changes`).

src/pudl_archiver/depositors/zenodo.py
jdangerx: This comment is left over from when we had that behavior - sorry for the confusion!

zschira: If we're not deleting draft depositions at the start, there's still a chance that we encounter a situation where we start working on a new version, but when we try to publish it something goes wrong and it doesn't publish so we're left with a floating draft that could cause unexpected behavior (like assuming nothing has changed). Could we delete drafts if `deposition.submitted` is False and we're running with `--initialize`?

src/pudl_archiver/orchestrator.py
zaneselvans: Ah okay, I didn't realize it would just roll the old files over into the new draft deposition. That makes sense since it does file-level deduplication on the backend.

tests/integration/zenodo_test.py
```diff
@@ -239,3 +242,20 @@ async def identity(x):
...
+        orchestrator.depositor.get_deposition,
+        args=[str(v2_not_updated.conceptdoi)],
+        kwargs={"published_only": True},
+        retry_on=(
+            ZenodoClientException,
+            aiohttp.ClientError,
+            asyncio.TimeoutError,
+            IndexError,
+        ),
+        retry_base_s=0.5,
```
zaneselvans, line 259: On the retry logic, in some previous logging it looked like the exponential backoff was going up to about a 1 minute wait time. I've definitely experienced network connectivity issues with some of the federal sites that last for longer than that, and wonder if it might make sense to make it possible to back off to as much as 15 or 30 minutes in the case of the scheduled periodic archiving runs that will presumably run overnight and can take a couple of hours without causing any problems. Though maybe keep it fast in the case of the CI, which we want to complete quickly.

zaneselvans (approved):

zaneselvans:
The `censusdp1tract` dataset failed, leaving behind a draft that's identical to the only published deposition, with a new version and a publication date of today: https://sandbox.zenodo.org/deposit/1167830

Based on the logs, it looks like it found a pre-existing draft (1166464, which would also have been identical to the existing deposition) and attempted to delete it, but got a `ClientConnectionError` and didn't retry (though that deposition was deleted at some point). Then it went on to try and create a new draft (1167830, which was successfully created), but got a bunch of additional "Connection closed" errors.

So maybe this was really just network connectivity issues with Zenodo, and a legit failure beyond the scope of the archiver (short of making it try harder / longer).

None of the other archiving runs generated lingering drafts.

```
2023-02-24 22:19:51 [    INFO] catalystcoop.pudl_archiver.archivers.classes:71 Archiving censusdp1tract
2023-02-24 22:19:51 [    INFO] catalystcoop.pudl_archiver.depositors.zenodo:92 GET https://sandbox.zenodo.org/api/deposit/depositions - Query depositions for 10.5072/zenodo.1151526
2023-02-24 22:19:54 [    INFO] catalystcoop.pudl_archiver.depositors.zenodo:92 GET https://sandbox.zenodo.org/api/deposit/depositions/1166464 - Get deposition for 1166464
2023-02-24 22:20:10 [    INFO] catalystcoop.pudl_archiver.depositors.zenodo:92 DELETE https://sandbox.zenodo.org/api/deposit/depositions/1166464 - Deleting deposition
Encountered exceptions, showing traceback for last one: ["('censusdp1tract', ClientConnectionError('Connection closed'))"]
2023-02-24 22:20:15 [    INFO] catalystcoop.pudl_archiver.depositors.zenodo:92 GET https://sandbox.zenodo.org/api/deposit/depositions - Query depositions for 10.5072/zenodo.1151526

2023-02-24 22:20:59 [    INFO] catalystcoop.pudl_archiver.depositors.zenodo:92 GET https://sandbox.zenodo.org/api/deposit/depositions/1151527 - Get deposition for 1151527
2023-02-24 22:22:06 [    INFO] catalystcoop.pudl_archiver.depositors.zenodo:92 POST https://sandbox.zenodo.org/api/deposit/depositions/1151527/actions/newversion - Creating new version
2023-02-24 22:22:08 [    INFO] catalystcoop.pudl_archiver.depositors.zenodo:92 PUT https://sandbox.zenodo.org/api/deposit/depositions/1167830 - Updating version number from 1.0.0 (1151527) to 2.0.0
2023-02-24 22:27:13 [    INFO] catalystcoop.pudl_archiver.utils:46 Error while executing <coroutine object ClientResponse.read at 0x7f0889c315b0> (try #1, retry in 2s): 
2023-02-24 22:27:15 [    INFO] catalystcoop.pudl_archiver.utils:46 Error while executing <coroutine object ClientResponse.read at 0x7f0889c31850> (try #2, retry in 4s): Connection closed
2023-02-24 22:27:19 [    INFO] catalystcoop.pudl_archiver.utils:46 Error while executing <coroutine object ClientResponse.read at 0x7f0889c315b0> (try #3, retry in 8s): Connection closed
2023-02-24 22:27:27 [    INFO] catalystcoop.pudl_archiver.utils:46 Error while executing <coroutine object ClientResponse.read at 0x7f0889c31850> (try #4, retry in 16s): Connection closed
2023-02-24 22:27:43 [    INFO] catalystcoop.pudl_archiver.utils:46 Error while executing <coroutine object ClientResponse.read at 0x7f0889c315b0> (try #5, retry in 32s): Connection closed
2023-02-24 22:28:15 [    INFO] catalystcoop.pudl_archiver.utils:46 Error while executing <coroutine object ClientResponse.read at 0x7f0889c31850> (try #6, retry in 64s): Connection closed
```

Issue #73, opened 2023-02-24
Ensure each XBRL filing in FERC RSS feed has a stable ID


zschira:
The FERC RSS feed is changing `guid`'s every time the feed is downloaded. According to the RSS standard this field is supposed to be a unique identifier of entries, which is why I chose to use it for naming the XBRL filings, but they don't seem to be conforming to this standard. I emailed FERC to check if they know this is happening, and if they plan to fix it, but I haven't heard back. While this is still the case, every time we run one of the FERC archivers, all of the XBRL file names will change and we will end up creating a new version even if none of the underlying data has changed.

We can fix this one of two ways:

1. Do nothing and wait for FERC to fix this.
2. Change how we name XBRL filings

Option 1 I think would be nice as it doesn't require any work and using UUID's is robust and won't lead to any accidental collisions. However, option 2 is probably the better option as we don't know if/when FERC will actually get to this.

zaneselvans:
When a respondent submits a revision to and old filing they submit an entirely new filing, with both the old and the revised filing continuing to exist in the feed, right?

If there's a filing submission timestamp available in the RSS feed (or the XBRL) and a respondent ID, could we use those to uniquely identify a filing? E.g. something like... `20230201123456-C000123.xbrl` This wouldn't preclude FERC from fixing the GUID. Have they been responsive in the past? Do they even have control over the system, or is it managed by XBRL-US I wonder?

zschira:
Yeah there's an entirely new entry in the feed when a filer resubmits a filing, and they do contain a timestamp, so I think your suggestion seems good. They've been somewhat responsive in the past, but I still haven't heard back from them and I don't know who manages the feed. Also, the naming convention you've provided here will be unique and is more descriptive than the `guid` even if it was working correctly, so I'm in favor of just going forward with changing it.

zaneselvans:
Each form has its own RSS feed right? So we don't need to worry about ID collisions between different forms?

jdangerx:
We changed how we name things, so I think we can close this, right @zschira ?

PR #82, opened 2023-02-28
Standardize filenames and add production DOIs

### LINKED ISSUE #79: Refresh Zenodo Archives Manually
The new archivers are working, but we don't have the infrastructure set up to run them on an automated schedule.  Instead up them manually, adding new archive lineages for the new datasets

- 🆕 New Dataset
- ✅ Updated Existing Archive
- ⛔ No Update
- 💀 Archiver is broken

```[tasklist]
- [x] ⛔ [Census DP1](https://zenodo.org/record/4127049)
- [x] 🆕 [EIA-176](https://zenodo.org/record/7682358) (See #80)
- [x] ⛔ [EIA-860](https://zenodo.org/record/7113854)
- [x] ✅ [EIA-860M](https://zenodo.org/record/7682381)
- [x] ✅ [EIA-923](https://zenodo.org/record/7682417)
- [x] ✅ [EIA Bulk Electricity](https://zenodo.org/record/7682482)
- [x] 🆕 [EIA Water](https://zenodo.org/record/7683136) (See #77, #80) 
- [x] ⛔ [EPA CAMD EIA Crosswalk](https://zenodo.org/record/7650939)
- [x] 💀 [EPA CEMS](https://zenodo.org/record/6910058) (See #68)
- [x] ⛔ FERC-1 (Skipped due to #73)
- [x] ⛔ FERC-2 (Skipped due to #73)
- [x] ⛔ FERC-6 (Skipped due to #73)
- [x] ⛔ FERC-60 (Skipped due to #73)
- [x] ⛔ FERC-714 (Skipped due to #73)
- [x] 💀 FERC EQR (See #31, #43, #44)
- [x] 🆕 [MSHA Mines](https://doi.org/10.5281/zenodo.7683517) (Lots of up & download errors. See #80, #81)
- [x] 🆕 [PHMSA Natural Gas](https://zenodo.org/record/7683352) (See #80)
```

zaneselvans:
Tweak the archive filenames slightly to match naming patterns from other archivers.

Add new production DOIs for:
* `eia176`
* `eiawater`
* `mshamines`
* `phmsagas`

I used these archivers to initialize archives on the production Zenodo site, and also updated a bunch of other archivers (See #79). However, I didn't update the FERC archives because of the issue with unique IDs in the RSS feeds described in #73.

Closes #79

### DIFF
--- a/src/pudl_archiver/archivers/phmsagas.py
+++ b/src/pudl_archiver/archivers/phmsagas.py
@@ -66,18 +66,18 @@ class PhmsaGasArchiver(AbstractDatasetArchiver):
         For example: annual_underground_natural_gas_storage_2017_present.zip
         """
         url = f"https://www.phmsa.dot.gov/{link}"
-        file = str(match.group(1)).replace("-", "_")  # Get file name
+        filename = str(match.group(1)).replace("-", "_")  # Get file name
 
         # Set dataset partition
-        dataset = "_".join(file.lower().split("_")[0:-2])
+        dataset = "_".join(filename.lower().split("_")[0:-2])
 
         if dataset not in PHMSA_DATASETS:
             logger.warning(f"New dataset type found: {dataset}.")
 
         # Set start year
-        start_year = int(file.split("_")[-2])
+        start_year = int(filename.split("_")[-2])
 
-        download_path = self.download_directory / f"phmsagas_{file}.zip"
+        download_path = self.download_directory / f"{self.name}-{filename}.zip"
         await self.download_zipfile(url, download_path)
 
         return ResourceInfo(
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -14,23 +14,23 @@ jobs:
     strategy:
       matrix:
         dataset:
-          - censusdp1tract
-          - eia176
-          - eia860
-          - eia860m
-          - eia861
-          - eia923
-          - eia_bulk_elec
-          - eiawater
-          - epacamd_eia
-          - epacems
-          - ferc1
-          - ferc2
-          - ferc6
-          - ferc60
-          - ferc714
+          # - censusdp1tract
+          # - eia176
+          # - eia860
+          # - eia860m
+          # - eia861
+          # - eia923
+          # - eia_bulk_elec
+          # - eiawater
+          # - epacamd_eia
+          # - epacems
+          # - ferc1
+          # - ferc2
+          # - ferc6
+          # - ferc60
+          # - ferc714
           - mshamines
-          - phmsagas
+          # - phmsagas
 
       fail-fast: false
     runs-on: ubuntu-latest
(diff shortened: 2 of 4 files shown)


src/pudl_archiver/archivers/phmsagas.py
```diff
@@ -66,18 +66,18 @@ async def get_zip_resource(
...
 
         if dataset not in PHMSA_DATASETS:
             logger.warning(f"New dataset type found: {dataset}.")
 
         # Set start year
-        start_year = int(file.split("_")[-2])
+        start_year = int(filename.split("_")[-2])
 
-        download_path = self.download_directory / f"phmsagas_{file}.zip"
+        download_path = self.download_directory / f"{self.name}-{filename}.zip"
```
zaneselvans, line 80: All of the other archivers use a file naming pattern like:

`dataset-part1-part2.ext`

where:
- dataset is the name of the dataset (e.g. ferc1, eia923, mshamines)
- part1 is the value of the first partition dimension contained in
  the file (e.g. ca, ny, wy, for states, 2019 if it's a year)
- part2 is the value of the second partition dimension contained in
  the file, if applicable.
- ext is the file name extension indicating file type.

I've changed the naming for phmsagas to follow the `dataset-*` part of the naming convention, but have left the filename in place rather than using only the dataset and start year, since if we just use the start year it looks like years are missing in the archive, as each file contains several years of data.

The `dataset` and `start_year` partitions are still used in datapackage.json to refer to the individual resources for programmatic purposes.

e-belfer: I agree with the decision not to use start date for the filenames here. But in this case the expected output here is a file like: `phmsagas-annual_underground_natural_gas_storage_2017_present.zip` ?

Would it be worth reversing the substitution from "-" to "_" in line 69 to result in filenames like ``phmsagas-annual-underground-natural-gas-storage-2017-present.zip``? Having both "-" and "_" is a bit visually confusing here and at least this would bring us towards having a consistent snake case filename, even if the structure is slightly different than those of other archivers.

e-belfer (approved):
One minor non-blocking suggestion on the naming of the phmsagas files. Otherwise all the DOI's and edits look good!

src/pudl_archiver/archivers/phmsagas.py
zaneselvans: I wasn't totally thrilled about this combo either. It also occurs in the MSHA archives (though there we do use the actual partition values, rather than the filename).  Really we don't want anybody parsing these filenames directly, and they should be using the metadata in `datapackage.json` to address individual partitions. I guess even what we have right now isn't literally the original filenames since we're changing the one with dashes to underscores. We haven't had to deal with this data-table based partitioning before. Previously it was either one file per timeslice (with individual tables inside it) or in the case of EPA CEMS, files that were all part of the same table, but partitioned the data based on values (year and state) within that one table.

I just wanted to somehow differentiate the `phmsagas` part of the name from the other parts, since it's a different kind of thing, and it was possible to follow the same convention as all the FERC and EIA data.

Another thing that we're seeing come up is overlap in partitions where what differentiates them is the file format, e.g. years where both XBRL and DBF data are available in FERC, which seems like another type of partitioning we might want to work with and make obvious in the filenames and partition metadata.

PR #111, opened 2023-05-03
Make partitions (data_format) lowercase when building archives

rousik:
What it says on the tin. Make sure to use lowercase values for `data_format` partition.


zaneselvans (approved):

zaneselvans:
@rousik is there any reason to hold off on merging this into `main`?

PR #130, opened 2023-07-19
Add hook for running dataset specific archive validation tests

zschira:
This PR creates the initial infrastructure for automatically generating and summarizing archive runs.

### DIFF
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -6,7 +6,6 @@ import tempfile
 import typing
 import zipfile
 from abc import ABC, abstractmethod
-from functools import reduce
 from html.parser import HTMLParser
 from pathlib import Path
 
@@ -52,6 +51,8 @@ class AbstractDatasetArchiver(ABC):
     """An abstract base archiver class."""
 
     name: str
+    # Configurable option to run _check_missing_files during archive validation
+    check_missing_files: bool = True
 
     def __init__(self, session: aiohttp.ClientSession):
         """Initialize Archiver object.
@@ -164,6 +165,34 @@ class AbstractDatasetArchiver(ABC):
 
         return hyperlinks
 
+    def _check_missing_files(
+        self,
+        baseline_datapackage: DataPackage | None,
+        new_datapackage: DataPackage,
+    ) -> ValidationTestResult:
+        """Check for any files from previous archive version missing in new version."""
+        baseline_resources = set()
+        if baseline_datapackage is not None:
+            baseline_resources = {
+                resource.name for resource in baseline_datapackage.resources
+            }
+
+        new_resources = {resource.name for resource in new_datapackage.resources}
+
+        # Check for any files only in baseline_datapackage
+        missing_files = baseline_resources - new_resources
+
+        note = None
+        if len(missing_files) > 0:
+            note = f"The following files would be deleted by new archive version: {missing_files}"
+
+        return ValidationTestResult(
+            name="Missing file test",
+            description="Check for files from previous version of archive that would be deleted by the new version",
+            success=len(missing_files) == 0,
+            note=note,
+        )
+
     def validate_archive(
         self,
         baseline_datapackage: DataPackage | None,
@@ -180,16 +209,18 @@ class AbstractDatasetArchiver(ABC):
         Returns:
             Bool indicating whether or not all tests passed.
         """
-        validation_tests = self.dataset_validate_archive(
+        validations = []
+        if self.check_missing_files:
+            validations.append(
+                self._check_missing_files(baseline_datapackage, new_datapackage)
+            )
+
+        dataset_specific_validations = self.dataset_validate_archive(
             baseline_datapackage, new_datapackage, resources
         )
-        test_results = reduce(
-            lambda result, test: result and (test.result or test.ignore_failure),
-            validation_tests,
-            True,
-        )
-
-        return test_results
+        validations += dataset_specific_validations
+        test_results = [(test.success or test.ignore_failure) for test in validations]
+        return all(test_results)
 
     def dataset_validate_archive(
         self,
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -6,16 +6,15 @@ import tempfile
 import typing
 import zipfile
 from abc import ABC, abstractmethod
-from collections import namedtuple
+from functools import reduce
 from html.parser import HTMLParser
 from pathlib import Path
 
 import aiohttp
 
+from pudl_archiver.archivers.validate import ValidationTestResult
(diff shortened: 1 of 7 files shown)


src/pudl_archiver/archivers/classes.py
```diff
@@ -165,6 +164,42 @@ async def get_hyperlinks(
 
         return hyperlinks
 
+    def validate_archive(
```
zschira, line 167: Eventually we can aggregate generalize-able tests in this method.

src/pudl_archiver/archivers/validate.py
```diff
@@ -0,0 +1,35 @@
...
+
+class FileDiff(BaseModel):
+    """Model summarizing changes to a single file in a deposition."""
+
+    name: str
+    diff_type: Literal["CREATE", "UPDATE", "DELETE"]
+    size_diff: int
+
+
+class RunSummary(BaseModel):
```
zschira, line 26: I started outlining a report that we could generate for every run that would include a summary of the changes, and results of any tests.

src/pudl_archiver/archivers/classes.py
```diff
@@ -165,6 +164,42 @@ async def get_hyperlinks(
...
+        return test_results
+
+    def dataset_validate_archive(
+        self,
+        baseline_datapackage: DataPackage | None,
+        new_datapackage: DataPackage,
+        resources: dict[str, ResourceInfo],
+    ) -> list[ValidationTestResult]:
+        """Hook to add archive validation tests specific to each dataset."""
+        return []
```
zaneselvans, line 201: I think we need at least one example validation test that we can use as a template to understand what these look like. Maybe a test that checks none of the files which were present in the previous archive have disappeared from the new archive?  What do we put here? The actual tests inline in this function? Or do we define other methods that are specific to the child class which pertains to a particular dataset, and then compile a list of their results here and return it?

src/pudl_archiver/archivers/classes.py
```diff
@@ -165,6 +164,42 @@ async def get_hyperlinks(
...
+
+        Args:
+            baseline_datapackage: DataPackage descriptor from previous version of archive.
+            new_datapackage: DataPackage descriptor from newly generated archive.
+            resources: Dictionary mapping resource name to ResourceInfo.
+
+        Returns:
+            Bool indicating whether or not all tests passed.
+        """
+        validation_tests = self.dataset_validate_archive(
```
zaneselvans, line 183: It seems like this is the results of the tests, rather than the tests themselves. Maybe call it `validation_results`?

Why are there both `validate_archive` and `dataset_validate_archive` methods?

src/pudl_archiver/archivers/validate.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class ValidationTestResult(BaseModel):
+    """Class containing results of a validation test, and metdata about the test."""
+
+    name: str
+    description: str
+    ignore_failure: bool = False
+    resource_name: str | None = None
+    result: bool
```
zaneselvans, line 14: I think `success` would give a clearer expectation of this being a bool, and what True vs. False would mean than `result`

src/pudl_archiver/archivers/validate.py
```diff
@@ -0,0 +1,35 @@
...
+    note: str | None = None
```
zaneselvans, line 15: What's the difference between `description` and `note`?

src/pudl_archiver/orchestrator.py
```diff
@@ -194,10 +195,17 @@ async def run(self):
 
         await self._apply_changes()
         self.new_deposition = await self.depositor.get_record(self.new_deposition.id_)
-        await self._update_datapackage(resources)
+        new_datapackage, old_datapackage = await self._update_datapackage(resources)
         await self._apply_changes()
 
         if self.changed:
+            validation_result = self.downloader.validate_archive(
```
zaneselvans, line 202: I think `validation_success` would be a bit more readily interpretable here.

src/pudl_archiver/archivers/classes.py
zaneselvans: Meaning we would have some validations that are run for every dataset, and specified here, and concatenated with the list of validation results that comes from the dataset-specific validation method?

zaneselvans (changes requested):
Mostly made some naming / idiomatic suggestions.

I think it would also be very helpful to have at least 1 example of an implemented test to go off of.

src/pudl_archiver/archivers/classes.py
```diff
@@ -165,6 +164,42 @@ async def get_hyperlinks(
...
+            baseline_datapackage, new_datapackage, resources
+        )
+        test_results = reduce(
+            lambda result, test: result and (test.result or test.ignore_failure),
+            validation_tests,
+            True,
+        )
```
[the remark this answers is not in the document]
zschira: Yeah the `reduce` pattern is more common in some other languages, but I think your way is more pythonic, so I've changed it.

And the flow is a little different now that I've added an example, but no difference between validations and tests, I've used the terms somewhat interchangeably.

src/pudl_archiver/archivers/classes.py
zschira: I added an example that checks for missing files from the previous archive. I think the cleanest way to add new tests would be to define a method for each test, and then call each method and compile results in a list. This is more or less what I've done in the example I added.

src/pudl_archiver/archivers/classes.py
zschira: Changed to `validation_results`.

As for the two methods, the intention is that `dataset_validate_archive` will be overridden by subclasses to add dataset specific validations. The naming of these two methods is definitely a bit confusing/wonky, but I felt like it would likely change as this evolves, so I didn't worry about it too much. For example, in a separate branch I'm working on, I've replaced `validate_archive` with a method `generate_summary` that returns a `RunSummary` object instead of just a bool.

src/pudl_archiver/archivers/validate.py
zschira: Agreed.

src/pudl_archiver/archivers/validate.py
zschira: The idea was that `note` was for adding details like why the test failed. I could change to something like `error_message` to be more clear, but I wasn't sure if I should constrain it to only be for error messages, or leave the option open to include other types of notes.

src/pudl_archiver/orchestrator.py
zschira: Agreed.

src/pudl_archiver/archivers/classes.py
zschira: Yeah that's the idea

zaneselvans (approved):

Issue #68, opened 2023-02-22
Rewrite EPA CEMS archiver to use API


zaneselvans:
At some point in the last few months, EPA seems to have removed the bulk CSV files containing hourly emissions data from their site, and are now exclusively making the data available through an API.

* The [old URL](https://gaftp.epa.gov/DMDnLoad/emissions/hourly/monthly) now gives a 404, though a lot of other data remains on the server. [Top level index](https://gaftp.epa.gov/)
* [Emissions data API Documentation](https://www.epa.gov/airmarkets/cam-api-portal#/swagger/emissions-mgmt)
* [GitHub repo with example usage](https://github.com/USEPA/cam-api-examples) 

Looking at their [example bulk Python download script](https://github.com/USEPA/cam-api-examples/blob/main/Python/bulk_data_api_demo.py) it seems like it's possible to download CSVs from an s3 bucket. Hopefully they're not too different from the CSVs we were previously downloading directly from EPA!

Note that this API replaces .zip files with one .csv file for each state-year, so we will also need to adjust the extraction step in `pudl` to match the new data format.

zaneselvans:
@bendnorman as you work on the gridstatus data archiving maybe keep this issue in mind too for application of similar patterns.

ggurjar333:
@zaneselvans Can I take this?

PR #131, opened 2023-07-25
Rewrite CEMS archiver to use API

e-belfer:
See #68. This PR modifies the API guide provided for bulk data access here: https://github.com/USEPA/cam-api-examples/blob/main/Python/bulk_data_api_demo.py#L9

The sandbox deposition created by this PR is here: https://sandbox.zenodo.org/record/1224605

To do:
- [ ] Fix errors produced in rebase
- [ ] Fix 'connection closed' errors
- [ ] Do we want to add our own date range here, or just download all available data - present?
- [ ] Add in-line documentation
- [ ] Very archiver working as expected
- [ ] Run production for 2022


e-belfer:
Still to resolve:

```
ImportError while loading conftest .../PUDL/pudl-archiver/tests/conftest.py'.
tests/conftest.py:6: in <module>
    from pudl_archiver.frictionless import DataPackage
.env_tox/lib/python3.10/site-packages/pudl_archiver/__init__.py:28: in <module>
    ARCHIVERS = {archiver.name: archiver for archiver in all_archivers()}
.env_tox/lib/python3.10/site-packages/pudl_archiver/__init__.py:24: in all_archivers
    __import__(module)
.env_tox/lib/python3.10/site-packages/pudl_archiver/archivers/epacems.py:69: in <module>
    parameters = {"api_key": os.environ["EPACEMS_API_KEY"]}  # Set to API key
../../../../mambaforge/envs/pudl-cataloger/lib/python3.10/os.py:680: in __getitem__
    raise KeyError(key) from None
E   KeyError: 'EPACEMS_API_KEY'
```

src/pudl_archiver/archivers/epacems.py
```diff
@@ -66,7 +64,9 @@
     "wi",
     "wy",
 ]
-BASE_URL = "https://gaftp.epa.gov/DMDnLoad/emissions/hourly/monthly"
+
+BASE_URL = "https://api.epa.gov/easey/bulk-files/emissions/hourly/state/"
+parameters = {"api_key": os.environ["EPACEMS_API_KEY"]}  # Set to API key
```
zaneselvans, line 69: How do the parameters / API key get passed through to the API request?

zaneselvans: Also this would be a great kind of thing to check for in the archive validations, and it's probably not too hard.

Or maybe this should be a more generic check that's done in the requests for all file downloads if we aren't already -- if we get a 404 or a 403 or some other error code, then probably something is wrong.

zaneselvans: Oh I guess it's an API error not an HTTP error response so it's dataset specific.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Error>
  <Code>AccessDenied</Code>
  <Message>Access Denied</Message>
  <RequestId>FWYCXG03J3B1X2F4</RequestId>
  <HostId>xP3qPNcx25yKpg1tNpVPQ+39RPCTG1VSpnmU1UraUfm91CXfACUaJlUgl6nm04zMh+wdgBOBJONGmx5Tcx0Ayw==</HostId>
</Error>
```

e-belfer: Great question. This works as is because turns out you can query the file downloads directly without needing an API - the API is only used if you're asking for details about files, which we aren't because of the way I've constructed this here. So perhaps I will deprecate the use of the API altogether in this archiver and see if it continues to run as expected.

zaneselvans: If you just try and download one of these CSVs directly in a browser or using `wget` or `curl` it works? That seems surprising to me.  In their example they're using the `requests` library, and are setting the parameters dictionary such that it includes the API key, which I think means it should be appending `?api_key=siodfsjdhkajsdskskldjdlkjf` to the URL that's actually accessed when they use `request.get()`

zaneselvans: Well, downloading with `wget` worked fine locally!

```bash
wget "https://api.epa.gov/easey/bulk-files/emissions/hourly/state/emissions-hourly-2020-vt.csv"
```

But I don't know what about the API there is to remove from this archiver. It's currently just downloading from a URL directly, and not attempting to use the API key.

e-belfer: Exactly what I mean, I'm just going to clean it up from the various places I've added it to (tox, etc.). Even more straightforward than it seems.

e-belfer:
Running into some weird hiccups with Zenodo on this archiver. Running the full archiver successfully downloads/uploads all datasets, but produces the following error when trying to upload the datapackage.json:

image

Then, when I try to run the archiver again on that archive, I get a "list index out of range" error and the request just returns an empty file list. The archive uploads to Zenodo but the datapackage file is corrupted.

@zschira Is this an error you've encountered elsewhere in dealing with the archivers, or do you have any thoughts on what piece of the puzzle to try to debug here?

zschira:
Did some digging, and it looks like the we're getting bad responses from the api, and so what's actually in the CSV files in just the JSON from the error response. We've set the `aiohttp` session to not raise exceptions on bad responses, so this kind of quietly fails right now, which is definitely not ideal, and I'll rethink how we should handle bad responses. I'm still not 100% sure why/if this is causing the weird zenodo errors, but I think this is a good place to start.

I assumed the bad responses were a result of not passing the api key to the requests, but I changed it to pass the key, and it's still giving bad responses. I'll continue digging on this and see what I can come up with.

src/pudl_archiver/archivers/epacems.py
```diff
@@ -76,46 +78,35 @@ class EpaCemsArchiver(AbstractDatasetArchiver):
...
-                states[match.group(2)].append(int(match.group(3)))
-
-            for state, months in states.items():
-                if len(months) > 0:
-                    yield self.get_state_year_resource(year, state, months)
+        file_list = requests.get(
+            "https://api.epa.gov/easey/camd-services/bulk-files",
+            params=parameters,
+            timeout=5,
+        )  # FIX ME to use existing classes
```
e-belfer, line 85: Running this produces "Connection closed" errors and I'm wondering if this line is the source @jdangerx

e-belfer:
Still getting `Connection closed` errors even with a concurrency of 1. Something must still be wrong here.

PR #132, opened 2023-07-25
Update dates on FERC60 archiver

e-belfer:
Update the dates on the FERC60 archiver to include 2020 data.

### DIFF
--- a/src/pudl_archiver/archivers/ferc/ferc60.py
+++ b/src/pudl_archiver/archivers/ferc/ferc60.py
@@ -16,7 +16,7 @@ class Ferc60Archiver(AbstractDatasetArchiver):
 
     async def get_resources(self) -> ArchiveAwaitable:
         """Download FERC 60 resources."""
-        for year in range(2006, 2020):
+        for year in range(2006, 2022):
             yield self.get_year_dbf(year)
 
         filings = xbrl.index_available_entries()[xbrl.FercForm.FORM_60]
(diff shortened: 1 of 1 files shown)


src/pudl_archiver/archivers/ferc/ferc60.py
```diff
@@ -16,7 +16,7 @@ class Ferc60Archiver(AbstractDatasetArchiver):
 
     async def get_resources(self) -> ArchiveAwaitable:
         """Download FERC 60 resources."""
-        for year in range(2006, 2020):
+        for year in range(2006, 2021):
```
zaneselvans, line 19: Whoops, that looks like it might have been a misunderstanding about how `range()` works.

All the Forms 1, 2, and 6 all seem to include some data (I think it's just Q1-Q2) from 2021 as well. I wonder why this one is different. Maybe it's only ever annual data?

e-belfer: On the website the page says it was last updated June 2022. Looks like this form is only required annually.

e-belfer: I'm going to go ahead and merge this today unless you have any qualms @zaneselvans ?

zaneselvans: Nope go for it!

PR #133, opened 2023-07-26
Archive summary

zschira:
This PR generates a `RunSummary` after every archiver run, which contains key details such as the number of files that have changed, the results of all validation tests, and metadata about the new version and previous version. The summary can be output as a JSON file, which will be attached to any runs of the `run-archiver` github action.

### DIFF
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -68,7 +68,13 @@ jobs:
           # ZENODO_TOKEN_UPLOAD: ${{ secrets.ZENODO_TOKEN_UPLOAD }}
           # ZENODO_TOKEN_PUBLISH: ${{ secrets.ZENODO_TOKEN_PUBLISH }}
         run: |
-          conda run -n pudl-cataloger pudl_archiver --sandbox --datasets ${{ matrix.dataset }}
+          conda run -n pudl-cataloger pudl_archiver --sandbox --datasets ${{ matrix.dataset }} --sumary-file ${{ matrix.dataset }}_run_summary.json
+
+      - name: Archive code coverage results
+        uses: actions/upload-artifact@v3
+        with:
+          name: $${{ matrix.dataset }}-run-summary
+          path: ${{ matrix.dataset }}_run_summary.json
 
   archive-notify:
     runs-on: ubuntu-latest
--- a/src/pudl_archiver/__init__.py
+++ b/src/pudl_archiver/__init__.py
@@ -1,5 +1,6 @@
 """Tool to download data resources and create archives on Zenodo for use in PUDL."""
 import asyncio
+import json
 import os
 import pathlib
 
@@ -33,6 +34,7 @@ async def archive_datasets(
     sandbox: bool = True,
     initialize: bool = False,
     dry_run: bool = True,
+    summary_file: str | None = None,
 ):
     """A CLI for the PUDL Zenodo Storage system."""
     if sandbox:
@@ -79,3 +81,13 @@ async def archive_datasets(
                 f"Encountered exceptions, showing traceback for last one: {[repr(e) for e in exceptions]}"
             )
             raise exceptions[-1][1]
+
+    if summary_file is not None:
+        run_summaries = [result.dict() for _, result in results]
+
+        with open(summary_file, "w") as f:
+            json.dump(run_summaries, f, indent=2)
+
+    validation_results = [result.success for _, result in results]
+    if not all(validation_results):
+        raise RuntimeError("Error: archive validation tests failed.")
--- a/src/pudl_archiver/orchestrator.py
+++ b/src/pudl_archiver/orchestrator.py
@@ -199,10 +199,10 @@ class DepositionOrchestrator:
         await self._apply_changes()
 
         if self.changed:
-            validation_success = self.downloader.validate_archive(
+            run_summary = self.downloader.generate_summary(
                 old_datapackage, new_datapackage, resources
             )
-            if not validation_success:
+            if not run_summary.success:
                 logger.error("Archive validation failed. Not publishing new archive.")
                 return self.deposition
 
--- a/src/pudl_archiver/orchestrator.py
+++ b/src/pudl_archiver/orchestrator.py
@@ -9,6 +9,7 @@ import yaml
 from pydantic import BaseModel
 
 from pudl_archiver.archivers.classes import AbstractDatasetArchiver
+from pudl_archiver.archivers.validate import RunSummary, Unchanged
 from pudl_archiver.depositors import ZenodoDepositor
 from pudl_archiver.frictionless import DataPackage, ResourceInfo
 from pudl_archiver.utils import retry_async
@@ -168,7 +169,7 @@ class DepositionOrchestrator:
         metadata = DepositionMetadata.from_data_source(data_source_id)
         return await self.depositor.create_deposition(metadata)
 
-    async def run(self):
+    async def run(self) -> RunSummary | Unchanged:
         """Run the entire deposition update process.
 
         1. Create pending deposition version to stage changes.
@@ -178,9 +179,7 @@ class DepositionOrchestrator:
         5. Update the dataset settings if this was a new deposition.
 
         Returns:
-            Updated Deposition object - if published, then return that; else
-            return the pending deposition version. Distinguishable by `state`
-            field being 'done' and 'unsubmitted', respectively.
+            RunSummary object or Unchanged if no changes are detected or run is a dry run.
         """
         await self._initialize()
         resources = await self.downloader.download_all_resources()
@@ -191,7 +190,7 @@ class DepositionOrchestrator:
         # Leave immediately after generating changes if dry_run
         if self.dry_run:
             logger.info("Dry run, aborting")
-            return self.new_deposition
+            return Unchanged(dataset_name=self.data_source_id, reason="Dry run.")
 
         await self._apply_changes()
         self.new_deposition = await self.depositor.get_record(self.new_deposition.id_)
@@ -204,15 +203,16 @@ class DepositionOrchestrator:
             )
             if not run_summary.success:
                 logger.error("Archive validation failed. Not publishing new archive.")
-                return self.deposition
+                await self.depositor.delete_deposition(self.new_deposition)
+                return run_summary
 
             published = await self.depositor.publish_deposition(self.new_deposition)
             if self.create_new:
                 self._update_dataset_settings(published)
-            return published
+            return run_summary
         else:
             await self.depositor.delete_deposition(self.new_deposition)
-            return self.deposition
+            return Unchanged(dataset_name=self.data_source_id)
 
     def _generate_changes(self, resources):
         for name, resource in resources.items():
@@ -332,7 +332,10 @@ class DepositionOrchestrator:
             files.pop("datapackage.json")
 
         datapackage = DataPackage.from_filelist(
-            self.data_source_id, files.values(), resources
+            self.data_source_id,
+            files.values(),
+            resources,
+            self.new_deposition.metadata.version,
         )
 
         datapackage_json = io.BytesIO(
--- a/tests/integration/zenodo_test.py
+++ b/tests/integration/zenodo_test.py
@@ -2,6 +2,7 @@
 import asyncio
 import os
 import tempfile
+import time
 from pathlib import Path
 
 import aiohttp
@@ -10,9 +11,11 @@ import pytest_asyncio
 import requests
 from dotenv import load_dotenv
 
+from pudl.metadata.classes import DataSource
+from pudl.metadata.constants import LICENSES
 from pudl_archiver.archivers.classes import AbstractDatasetArchiver, ResourceInfo
+from pudl_archiver.archivers.validate import Unchanged
 from pudl_archiver.depositors.zenodo import ZenodoClientException
-from pudl_archiver.frictionless import DataPackage
 from pudl_archiver.orchestrator import DepositionOrchestrator
 from pudl_archiver.utils import retry_async
 from pudl_archiver.zenodo.entities import (
@@ -111,6 +114,19 @@ def test_files():
 
 
 @pytest.fixture()
+def datasource():
+    """Create fake datasource for testing."""
+    return DataSource(
+        name="pudl_test",
+        title="Pudl Test",
+        description="Test dataset for the sandbox, thanks!",
+        path="https://fake.link",
+        license_raw=LICENSES["cc-by-4.0"],
+        license_pudl=LICENSES["cc-by-4.0"],
+    )
+
+
+@pytest.fixture()
 def test_settings():
     """Create temporary DOI settings file."""
     with tempfile.TemporaryDirectory() as tmp_dir:
@@ -129,7 +145,7 @@ async def test_zenodo_workflow(
     test_settings: Path,
     test_files: dict[str, list[dict[str, str]]],
     deposition_metadata: DepositionMetadata,
-    datapackage: DataPackage,
+    datasource: DataSource,
     mocker,
 ):
     """Test the entire zenodo client workflow."""
@@ -183,10 +199,10 @@ async def test_zenodo_workflow(
     )
 
     # Mock out creating datapackage with fake data source
-    datapackage_mock = mocker.MagicMock(return_value=datapackage)
+    datasource_mock = mocker.MagicMock(return_value=datasource)
     mocker.patch(
-        "pudl_archiver.orchestrator.DataPackage.from_filelist",
-        new=datapackage_mock,
+        "pudl_archiver.frictionless.DataSource.from_id",
+        new=datasource_mock,
     )
 
     # Create new deposition and add files
@@ -206,8 +222,12 @@ async def test_zenodo_workflow(
             }
         )
     )
-    v1 = await orchestrator.run()
-    v1_refreshed = await orchestrator.depositor.get_record(v1.id_)
+    v1_summary = await orchestrator.run()
+    assert v1_summary.success
+
+    v1_refreshed = await orchestrator.depositor.get_record(
+        orchestrator.new_deposition.id_
+    )
     verify_files(test_files["original"], v1_refreshed)
 
     # Update files
@@ -226,16 +246,31 @@ async def test_zenodo_workflow(
         )
     )
 
-    v2 = await orchestrator.run()
-    v2_refreshed = await orchestrator.depositor.get_record(v2.id_)
+    # Should fail due to deleted file
+    v2_summary = await orchestrator.run()
+    assert not v2_summary.success
+
+    # Wait for deleted deposition to propogate through
+    time.sleep(1)
+
+    # Disable test and re-run
+    orchestrator.downloader.check_missing_files = False
+    v2_summary = await orchestrator.run()
+    assert v2_summary.success
+
+    v2_refreshed = await orchestrator.depositor.get_record(
+        orchestrator.new_deposition.id_
+    )
     verify_files(test_files["updated"], v2_refreshed)
 
     # no updates to make, should not leave the conceptdoi pointing at a draft
-    v2_not_updated = await orchestrator.run()
+    v3_summary = await orchestrator.run()
+    assert isinstance(v3_summary, Unchanged)
+
     # unfortunately, it looks like Zenodo doesn't propagate deletion instantly - retry this a few times.
     latest_for_conceptdoi = await retry_async(
         orchestrator.depositor.get_deposition,
-        args=[str(v2_not_updated.conceptdoi)],
+        args=[str(orchestrator.deposition.conceptdoi)],
         kwargs={"published_only": True},
         retry_on=(
             ZenodoClientException,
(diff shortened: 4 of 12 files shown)


.github/workflows/run-archiver.yml
```diff
@@ -68,7 +68,13 @@ jobs:
           # ZENODO_TOKEN_UPLOAD: ${{ secrets.ZENODO_TOKEN_UPLOAD }}
           # ZENODO_TOKEN_PUBLISH: ${{ secrets.ZENODO_TOKEN_PUBLISH }}
         run: |
-          conda run -n pudl-cataloger pudl_archiver --sandbox --datasets ${{ matrix.dataset }}
+          conda run -n pudl-cataloger pudl_archiver --sandbox --datasets ${{ matrix.dataset }} --sumary-file ${{ matrix.dataset }}_run_summary.json
```
jdangerx, line 71: nit: this flag should probably be called `--summary-file`

.github/workflows/run-archiver.yml
```diff
@@ -68,7 +68,13 @@ jobs:
...
+
+      - name: Archive code coverage results
+        uses: actions/upload-artifact@v3
+        with:
+          name: $${{ matrix.dataset }}-run-summary
+          path: ${{ matrix.dataset }}_run_summary.json
```
jdangerx, line 77: For now, we'd only access these through the Actions UI, right?

src/pudl_archiver/__init__.py
```diff
@@ -79,3 +81,13 @@ async def archive_datasets(
...
+
+    if summary_file is not None:
+        run_summaries = [result.dict() for _, result in results]
+
+        with open(summary_file, "w") as f:
+            json.dump(run_summaries, f, indent=2)
+
+    validation_results = [result.success for _, result in results]
+    if not all(validation_results):
+        raise RuntimeError("Error: archive validation tests failed.")
```
jdangerx, line 93: Is this your imagined workflow?:

1. see this
2. shoot, need to go look at the run summary file

tests/unit/archive_validate_test.py
```diff
@@ -0,0 +1,266 @@
...
+                    hash="hash1",
+                ),
+            },
+            [],
+        ),
+    ],
+)
+def test_process_resource_diffs(baseline_resources, new_resources, diffs):
+    """Test resource diffs."""
+    test_diffs = validate._process_resource_diffs(baseline_resources, new_resources)
```
jdangerx, line 265: Typically I'd avoid testing the private function and test the public interface instead - here, that'd be `RunSummary.create_summary`. 

It's 6PM on a Friday so no code sample, but I'll crank on one on Monday morning so you have something to actually compare things to.

tests/unit/archive_validate_test.py
```diff
@@ -0,0 +1,266 @@
...
+        ),
+    ],
+)
+def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
+    """Test partition diffs."""
+    test_diffs = validate._process_partition_diffs(baseline_partitions, new_partitions)
+    assert test_diffs == diffs
+
+
+@pytest.mark.parametrize(
```
jdangerx, line 58: Good instinct to parametrize, though I personally still find this data kind of hard to follow just because there's so many lines of it. Seems like there's a lot of repetition with small tweaks in the data - maybe consider a test data factory that churns out similar-looking data? I'll mess around a bit on Monday and see what I come up with.

tests/unit/archive_base_test.py
```diff
@@ -235,7 +235,10 @@ async def test_get_hyperlinks(docname, pattern, links, request, html_docs):
 def test_validate_archive(datapackage, test_results, success):
```
jdangerx, line 235: nit: probably we should change the name of this test if the method under test is changing.

src/pudl_archiver/orchestrator.py
```diff
@@ -191,28 +190,29 @@ async def run(self):
...
-            return self.new_deposition
+            return Unchanged(dataset_name=self.data_source_id, reason="Dry run.")
 
         await self._apply_changes()
         self.new_deposition = await self.depositor.get_record(self.new_deposition.id_)
         new_datapackage, old_datapackage = await self._update_datapackage(resources)
         await self._apply_changes()
 
         if self.changed:
-            validation_success = self.downloader.validate_archive(
```
jdangerx, line 202: The first structure that jumps to mind is

```
orchestrator
    |--> downloader
    |--> validator/summarizer
    |--> uploader
```

Where the three "thing-doers" are siblings and the orchestrator tells them what to do.

What were you considering when thinking about where to put the validation/summary stuff?

tests/integration/zenodo_test.py
```diff
@@ -226,16 +246,31 @@ async def identity(x):
         )
     )
 
-    v2 = await orchestrator.run()
-    v2_refreshed = await orchestrator.depositor.get_record(v2.id_)
+    # Should fail due to deleted file
+    v2_summary = await orchestrator.run()
```
jdangerx, line 250: Is the idea that this will remove a file, and `check_missing_files` says "if there was a file before, and now that file is gone, SCREAM"?

jdangerx (changes requested):
This looks like it will work, and be super useful eventually! I've some readability / structure suggestions to discuss before we merge, and will play around with the actual code on Monday morning so I'm not just idly speculating on what "could" be better.

tests/unit/archive_validate_test.py
jdangerx:
```diff
diff --git a/tests/unit/archive_validate_test.py b/tests/unit/archive_validate_test.py
index 46eb31a..7b332b9 100644
--- a/tests/unit/archive_validate_test.py
+++ b/tests/unit/archive_validate_test.py
@@ -59,8 +59,8 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
     "baseline_resources,new_resources,diffs",
     [
         (
-            {
-                "resource0": Resource(
+            [
+                Resource(
                     name="resource0",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -71,7 +71,7 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=10,
                     hash="hash0",
                 ),
-                "resource1": Resource(
+                Resource(
                     name="resource1",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -82,9 +82,9 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=20,
                     hash="hash1",
                 ),
-            },
-            {
-                "resource0": Resource(
+            ],
+            [
+                Resource(
                     name="resource0",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -95,7 +95,7 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=10,
                     hash="hash0",
                 ),
-                "resource1": Resource(
+                Resource(
                     name="resource1",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -106,7 +106,7 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=40,
                     hash="hash1_changed",
                 ),
-            },
+            ],
             [
                 validate.FileDiff(
                     name="resource1",
@@ -116,8 +116,8 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
             ],
         ),
         (
-            {
-                "resource0": Resource(
+            [
+                Resource(
                     name="resource0",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -128,7 +128,7 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=10,
                     hash="hash0",
                 ),
-                "resource1": Resource(
+                Resource(
                     name="resource1",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -139,9 +139,9 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=20,
                     hash="hash1",
                 ),
-            },
-            {
-                "resource0": Resource(
+            ],
+            [
+                Resource(
                     name="resource0",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -152,7 +152,7 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=10,
                     hash="hash0",
                 ),
-            },
+            ],
             [
                 validate.FileDiff(
                     name="resource1",
@@ -162,8 +162,8 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
             ],
         ),
         (
-            {
-                "resource0": Resource(
+            [
+                Resource(
                     name="resource0",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -174,9 +174,9 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=10,
                     hash="hash0",
                 ),
-            },
-            {
-                "resource0": Resource(
+            ],
+            [
+                Resource(
                     name="resource0",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -187,7 +187,7 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=10,
                     hash="hash0",
                 ),
-                "resource1": Resource(
+                Resource(
                     name="resource1",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -198,7 +198,7 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=20,
                     hash="hash1_created",
                 ),
-            },
+            ],
             [
                 validate.FileDiff(
                     name="resource1",
@@ -208,8 +208,8 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
             ],
         ),
         (
-            {
-                "resource0": Resource(
+            [
+                Resource(
                     name="resource0",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -220,7 +220,7 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=10,
                     hash="hash0",
                 ),
-                "resource1": Resource(
+                Resource(
                     name="resource1",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -231,9 +231,9 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=20,
                     hash="hash1",
                 ),
-            },
-            {
-                "resource0": Resource(
+            ],
+            [
+                Resource(
                     name="resource0",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -244,7 +244,7 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=10,
                     hash="hash0",
                 ),
-                "resource1": Resource(
+                Resource(
                     name="resource1",
                     path="https://www.fake.link",
                     remote_url="https://www.fake.link",
@@ -255,12 +255,24 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
                     bytes=20,
                     hash="hash1",
                 ),
-            },
+            ],
             [],
         ),
     ],
 )
-def test_process_resource_diffs(baseline_resources, new_resources, diffs):
+def test_process_resource_diffs(baseline_resources, new_resources, diffs, mocker):
     """Test resource diffs."""
-    test_diffs = validate._process_resource_diffs(baseline_resources, new_resources)
+    baseline = mocker.MagicMock(
+        resources=baseline_resources, version="0.0.1", created="2023-01-01"
+    )
+    new = mocker.MagicMock(
+        resources=new_resources, version="0.0.2", created="2023-01-02"
+    )
+    summary = validate.RunSummary.create_summary(
+        name="test package",
+        baseline_datapackage=baseline,
+        new_datapackage=new,
+        validation_tests=[],
+    )
+    test_diffs = summary.file_changes
     assert test_diffs == diffs
```

jdangerx: This leaves us freer to mess with the implementation details later down the line.

tests/unit/archive_validate_test.py
jdangerx: Here's an example of what such a factory might look like, and how it makes it easier to add more test cases:

```diff
diff --git a/tests/unit/archive_validate_test.py b/tests/unit/archive_validate_test.py
index 46eb31a..1b7d384 100644
--- a/tests/unit/archive_validate_test.py
+++ b/tests/unit/archive_validate_test.py
@@ -55,58 +55,30 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
     assert test_diffs == diffs
 
 
+def fake_resource(num=0, **kwargs):
+    params = {
+        "name": f"resource{num}",
+        "path": "https://www.fake.link",
+        "remote_url": "https://www.fake.link",
+        "title": f"Resource {num}",
+        "parts": {},
+        "mediatype": "zip",
+        "format": "format",
+        "bytes": 10,
+        "hash": f"hash{num}",
+    } | kwargs
+    return Resource(**params)
+
+
 @pytest.mark.parametrize(
     "baseline_resources,new_resources,diffs",
     [
         (
-            {
-                "resource0": Resource(
-                    name="resource0",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 0",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=10,
-                    hash="hash0",
-                ),
-                "resource1": Resource(
-                    name="resource1",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 1",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=20,
-                    hash="hash1",
-                ),
-            },
-            {
-                "resource0": Resource(
-                    name="resource0",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 0",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=10,
-                    hash="hash0",
-                ),
-                "resource1": Resource(
-                    name="resource1",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 1",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=40,
-                    hash="hash1_changed",
-                ),
-            },
+            [
+                fake_resource(0),
+                fake_resource(1, bytes=20),
+            ],
+            [fake_resource(0), fake_resource(1, bytes=40, hash="hash1_changed")],
             [
                 validate.FileDiff(
                     name="resource1",
@@ -116,43 +88,11 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
             ],
         ),
         (
-            {
-                "resource0": Resource(
-                    name="resource0",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 0",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=10,
-                    hash="hash0",
-                ),
-                "resource1": Resource(
-                    name="resource1",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 1",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=20,
-                    hash="hash1",
-                ),
-            },
-            {
-                "resource0": Resource(
-                    name="resource0",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 0",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=10,
-                    hash="hash0",
-                ),
-            },
+            [
+                fake_resource(0),
+                fake_resource(1, bytes=20),
+            ],
+            [fake_resource(0)],
             [
                 validate.FileDiff(
                     name="resource1",
@@ -162,43 +102,8 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
             ],
         ),
         (
-            {
-                "resource0": Resource(
-                    name="resource0",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 0",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=10,
-                    hash="hash0",
-                ),
-            },
-            {
-                "resource0": Resource(
-                    name="resource0",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 0",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=10,
-                    hash="hash0",
-                ),
-                "resource1": Resource(
-                    name="resource1",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 1",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=20,
-                    hash="hash1_created",
-                ),
-            },
+            [fake_resource(0)],
+            [fake_resource(0), fake_resource(1, bytes=20)],
             [
                 validate.FileDiff(
                     name="resource1",
@@ -208,59 +113,30 @@ def test_process_partition_diffs(baseline_partitions, new_partitions, diffs):
             ],
         ),
         (
-            {
-                "resource0": Resource(
-                    name="resource0",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 0",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=10,
-                    hash="hash0",
-                ),
-                "resource1": Resource(
-                    name="resource1",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 1",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=20,
-                    hash="hash1",
-                ),
-            },
-            {
-                "resource0": Resource(
-                    name="resource0",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 0",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=10,
-                    hash="hash0",
-                ),
-                "resource1": Resource(
-                    name="resource1",
-                    path="https://www.fake.link",
-                    remote_url="https://www.fake.link",
-                    title="Resource 1",
-                    parts={},
-                    mediatype="zip",
-                    format="format",
-                    bytes=20,
-                    hash="hash1",
-                ),
-            },
+            [fake_resource(0), fake_resource(1, bytes=20)],
+            [fake_resource(0), fake_resource(1, bytes=20)],
             [],
         ),
+        (
+            [fake_resource(0)],
+            [fake_resource(0, hash="hash0_changed")],
+            [validate.FileDiff(name="resource0", diff_type="UPDATE", size_diff=0)],
+        ),
     ],
 )
def test_process_resource_diffs(baseline_resources, new_resources, diffs):
     """Test resource diffs."""
     test_diffs = validate._process_resource_diffs(baseline_resources, new_resources)
     assert test_diffs == diffs
```

.github/workflows/run-archiver.yml
zschira: Yeah right now that's the case. I've thought about adding a json file to the actual depositions, so there will be a history right in the deposition. Do you think this would be worth doing?

src/pudl_archiver/__init__.py
zschira: Yeah I thought about printing out any failed validation tests here, but decided it was sufficient to just go to the file if there's a failure. I'm definitely open to changing this though

tests/integration/zenodo_test.py
zschira: Yeah the `check_missing_files` test will be enabled by default since that would generally indicate something like a missing year, so this will first check that it failed, then switch the test off and check that it succeeds the second time.

.github/workflows/run-archiver.yml
jdangerx: Sometime down the line, I guess that could be useful. But this seems fine for now.

src/pudl_archiver/orchestrator.py
zschira: The reason I put the validator/summarizer stuff in the downloader is so each dataset can override the `dataset_validate_archive` method to add one off validation tests. One other pattern I could see is adding a decorator for adding validation tests, so tests could still be defined in the same module as a datasets archiver without needing them to be in the class itself. Something like:

```py
class Eia860Archiver(AbstractDatasetArchiver):
    # Do downloading stuff


@validation_test(dataset_name="eia860")
def test_something():
    return True
```

Do you think this pattern would make more sense? It would allow us to keep the downloading/validating/uploading separate

jdangerx: Ah, so basically if we have the downloader & the validator/summarizer be different classes, we end up with two dataset-specific classes per dataset. Which is annoying for sure. I think I'd rather keep everything in the same class than have a bunch of dangling validation tests in the module, that separation doesn't seem useful enough to justify the sort of strange pattern.

.github/workflows/run-archiver.yml
zschira: Fixed

tests/unit/archive_validate_test.py
zschira: Changed, thanks

tests/unit/archive_validate_test.py
zschira: Good call, much cleaner now

tests/unit/archive_base_test.py
zschira: Fixed

src/pudl_archiver/orchestrator.py
zschira: Yeah ideally we'll get to the place where if you subclass the `AbstractDatasetArchiver` and implement a couple of basic methods, you'll end up with new archives being auto-generated  and not need to think about anything else

jdangerx (approved):
:shipit:

PR #138, opened 2023-08-01
Rewrite CEMS archiver to use API

e-belfer:
See #68. This PR modifies the API guide provided for bulk data access here: https://github.com/USEPA/cam-api-examples/blob/main/Python/bulk_data_api_demo.py#L9

Relevant restrictions for the CAM API are found [here](https://www.epa.gov/power-sector/cam-api-portal#/frequent-questions):

> For the CAM APIs, we are using the following:
> 
> Default rate limits of 1,000 requests per hour
> 
> 15-minute timeout on any API request, and
> 
> Endpoints that support paging are limited to 25,000 rows per page
> 
> The CAMPD app user interface limits downloads from the streaming services to 1,000,000 records but there is no download records limit on the streaming services APIs

To meet these expectations, I've added a more moderate version of the chunking found in #43, which enables splitting the read/write operations into sets. I've used 6 here as a number that tends not to span past the 15 minute API timeout above, even for the largest files.

The API also incorporates aiohttp tracking of requests as debugging log statements, to make it clearer in what order operations are occurring.

To do:
- [x] Fix 'connection closed' errors caused when all requests are initiated at once, but the later requests are waiting to be processed by asyncio.
- [x] Do we want to add our own end date range here, or just download all available data - present? --> Downloading all available data, consistent with other archivers.
- [x] Debug read errors
- [x] Rezip each csv into a zip file
- [x] Write a test for `download_and_zip` method
- [x] Update sandbox DOI
- [x] Verify archiver working as expected
- [ ] Run production for 2022

Out of scope:
- [ ] Speed up I/O operations, which are the major source of slowdown here


e-belfer:
Successful sandbox archive here: https://sandbox.zenodo.org/record/1228519

dataset_doi.yaml
```diff
@@ -27,7 +27,7 @@ epacamd_eia:
   sandbox_doi: 10.5072/zenodo.1072000
 epacems:
   production_doi: 10.5281/zenodo.4127054
-  sandbox_doi: 10.5072/zenodo.672962
+  sandbox_doi: 10.5072/zenodo.1228518
```
zaneselvans, line 30: Why does the concept DOI have to change? I would imagine that we want to keep any new CEMS archives part of the same versioned lineage.

src/pudl_archiver/archivers/epacems.py
```diff
@@ -66,56 +69,85 @@
...
+BASE_URL = "https://api.epa.gov/easey/bulk-files/"
+parameters = {"api_key": os.environ["EPACEMS_API_KEY"]}  # Set to API key
 
 
 class EpaCemsArchiver(AbstractDatasetArchiver):
     """EPA CEMS archiver."""
 
     name = "epacems"
+    concurrency_limit = 6  # Number of files to concurrently download
+    check_missing_files = False  # Temporary to allow set up of new sandbox
```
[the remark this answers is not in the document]
e-belfer: Oops, this should be reset to default settings now.

jdangerx (approved):
A bunch of small code nits, but the general structure looks good! I'm also curious about why you had to change the sandbox DOI - but I think it's non-blocking since it's just the sandbox.

dataset_doi.yaml
jdangerx: 😅 that makes a lot of sense!

e-belfer:
Archiver currently fails because of the following files:

> "note": "The following files would be deleted by new archive version: {'epacems-1996-mt.zip', 'epacems-1996-ri.zip', 'epacems-1996-ut.zip', 'epacems-1996-wa.zip', 'epacems-1996-az.zip', 'epacems-1995-wa.zip', 'epacems-1996-me.zip', 'epacems-1995-ca.zip', 'epacems-1996-nc.zip', 'epacems-1995-nd.zip', 'epacems-1996-vt.zip', 'epacems-1995-la.zip', 'epacems-1995-sc.zip', 'epacems-1996-nm.zip', 'epacems-1996-nv.zip', 'epacems-1995-ne.zip', 'epacems-1995-ri.zip', 'epacems-1996-sc.zip', 'epacems-1995-me.zip', 'epacems-1996-ok.zip', 'epacems-1996-id.zip', 'epacems-1995-sd.zip', 'epacems-1995-vt.zip', 'epacems-1996-co.zip', 'epacems-1995-mt.zip', 'epacems-1995-nc.zip', 'epacems-1995-id.zip', 'epacems-1995-ok.zip', 'epacems-1996-dc.zip', 'epacems-1995-az.zip', 'epacems-1995-de.zip', 'epacems-1996-sd.zip', 'epacems-1996-ca.zip', 'epacems-1996-la.zip', 'epacems-1995-or.zip', 'epacems-1995-ar.zip', 'epacems-1995-nm.zip', 'epacems-1996-de.zip', 'epacems-1995-nv.zip', 'epacems-1996-tx.zip', 'epacems-1995-dc.zip', 'epacems-1995-tx.zip', 'epacems-1995-va.zip', 'epacems-1996-va.zip', 'epacems-1996-ne.zip', 'epacems-1995-ct.zip', 'epacems-1996-ct.zip', 'epacems-1996-ar.zip', 'epacems-1995-co.zip', 'epacems-1996-nd.zip', 'epacems-1996-or.zip'}"

zaneselvans:
Okay so in the old (last year) CEMS archive, which came from the old EPA data source, they had files for those years but they were empty.  And now they’ve just gotten rid of the empty files, which seems like it makes sense.  I could see that creating issues in the ETL downstream where it might expect all of the files to exist.

The weird one-off Puerto Rico data and the post 2019 AK data could also be weird.

e-belfer:
> Okay so in the old (last year) CEMS archive, which came from the old EPA data source, they had files for those years but they were empty. And now they’ve just gotten rid of the empty files, which seems like it makes sense. I could see that creating issues in the ETL downstream where it might expect all of the files to exist.
> 
> The weird one-off Puerto Rico data and the post 2019 AK data could also be weird.

Exactly. But I'd like to resolve those issues in PUDL in the integration PR. To get around this I think my plan is to temporarily turn off the flag for the full run and run with `skip-publish`, check manually that all the files expected to be there are there by looking at the summary file and then turn it back on for future runs (which shouldn't drop any files). Shouldn't change anything that's in this PR in the end, but will just change what I'm actually running.

@zaneselvans Thoughts on this plan?

zaneselvans:
@e-belfer that plan seems good for now. At some point we should probably figure out a way to turn checks on and off at the command line rather than by editing the code, so we can temporarily ignore certain problems after checking on them.  But for now let's get it integrated!

zaneselvans (approved):

src/pudl_archiver/archivers/epacems.py
```diff
@@ -66,56 +69,87 @@
...
-                await self.download_zipfile(url, f_memory, ssl=False)
-
-                # Write to zipfile
-                with zipfile.ZipFile(
-                    archive_path, "a", compression=zipfile.ZIP_DEFLATED
-                ) as archive:
-                    with archive.open(filename, "w") as f_disk:
-                        f_disk.write(f_memory.read())
+        Args:
+            file: a dictionary containing file characteristics from the EPA API.
```
jdangerx, line 129: non-blocking: Is there a published schema we expect this metadata to follow? If so, might be nice to link to it here so we can find it the next time this all breaks down.

src/pudl_archiver/cli.py
```diff
@@ -55,6 +55,11 @@ def parse_main():
         type=str,
         help="Generate a JSON archive run summary",
     )
+    parser.add_argument(
+        "--skip-publish",
```
jdangerx, line 59: non-blocking: I wonder if requiring review should be the default behavior and auto-publishing should be the flagged behavior? Seems a little safer, but I'll leave that up to your judgment.

jdangerx (changes requested):
A bunch of nits, a couple of them blocking - but feel free to dismiss my review and merge once you've addressed stuff!

src/pudl_archiver/archivers/classes.py
```diff
@@ -212,6 +247,38 @@ def _check_missing_files(
             note=note,
         )
 
+    def _check_valid_files(self) -> ValidationTestResult:
```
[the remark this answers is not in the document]
e-belfer: Yes this gets actually run in `download_all_resources`, this just compiles the results into a Note.

PR #139, opened 2023-08-02
Download data from specific years to a specific location

jdangerx:
When investigating the FERC 2022 data for catalyst-cooperative/pudl#2766, I found that it was hard to "Just get the raw 2022 FERC 1 XBRL data and save it somewhere on disk where I can mess with it."

This adds:
* yearly filtering for the datasets where that was easy to do
* ability to dump the downloaded file to a non-temporary directory

### DIFF
--- a/src/pudl_archiver/archivers/eia861.py
+++ b/src/pudl_archiver/archivers/eia861.py
@@ -22,21 +22,20 @@ class Eia861Archiver(AbstractDatasetArchiver):
         link_pattern = re.compile(r"f861(\d{2,4}).zip")
 
         for link in await self.get_hyperlinks(BASE_URL, link_pattern):
-            yield self.get_year_resource(link, link_pattern.search(link))
+            year = int(link_pattern.search(link).group(1))
+            # Older file names only have last two digits of year in name
+            # Convert to 4-digit years
+            if year < 100 and year >= 90:
+                year += 1900
+            elif year < 100 and year < 90:
+                year += 2000
+            if self.valid_year(year):
+                yield self.get_year_resource(link, year)
 
-    async def get_year_resource(
-        self, link: str, match: typing.Match
-    ) -> tuple[Path, dict]:
+    async def get_year_resource(self, link: str, year: int) -> tuple[Path, dict]:
         """Download zip file."""
         # Use archive link if year is not most recent year
         url = f"{BASE_URL}/{link}"
-        year = int(match.group(1))
-        # Older file names only have last two digits of year in name
-        # Convert to 4-digit years
-        if year < 100 and year >= 90:
-            year += 1900
-        elif year < 100 and year < 90:
-            year += 2000
         download_path = self.download_directory / f"eia861-{year}.zip"
         await self.download_zipfile(url, download_path)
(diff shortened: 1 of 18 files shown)


e-belfer:
If you branch this off of the CEMS archiver adding a year check in there should follow the exact same pattern, since we're pulling the year from the metadata already. Alternately if you get this merged in first I can adapt my code pretty simply.

src/pudl_archiver/archivers/eia861.py
```diff
@@ -22,21 +21,21 @@ async def get_resources(self) -> ArchiveAwaitable:
...
-        self, link: str, match: typing.Match
-    ) -> tuple[Path, dict]:
+            year = int(link_pattern.search(link).group(1))
+            # Older file names only have last two digits of year in name
+            # Convert to 4-digit years
+            if year < 100 and year >= 90:
+                year += 1900
+            elif year < 100 and year < 90:
+                year += 2000
+                print(year)
```
zschira, line 31: Do we need this print statement? Also a nitpick, we could just do `elif year < 90:`

tests/unit/ferc_archiver_test.py
```diff
@@ -0,0 +1,76 @@
...
+        {"years": [2004, 2005, 2006, 2007, 2021], "num_dbf": 0, "num_xbrl": 2},
+    ],
+}
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    "module_name,test_spec", make_testdata(valid_years_test_spec_dict)
+)
+async def test_valid_years(module_name, test_spec, mocker):
```
zschira, line 59: This test seems useful!

src/pudl_archiver/archivers/eia923.py
```diff
@@ -22,14 +21,13 @@ async def get_resources(self) -> ArchiveAwaitable:
         link_pattern = re.compile(r"f((923)|(906920))_(\d{4})\.zip")
 
         for link in await self.get_hyperlinks(BASE_URL, link_pattern):
-            yield self.get_year_resource(link, link_pattern.search(link))
+            year = int(link_pattern.search(link).group(4))
```
zschira, line 24: I was trying to think of a way to have the base class handle checking the year without having to call `valid_year` from every subclass, but I don't see any particularly straightforward ways to do that, so this seems fine

zschira (approved):
Looks good! Just a couple very minor nitpicks, feel free to do as you will with them and merge when ready

src/pudl_archiver/archivers/eia923.py
jdangerx: Yeah, I had been hoping for this too but because different datasets handle years differently, and some have XBRL/DBF and some don't, it just seemed easier to put it in each subclass. If we figure something out later it's always easier to add abstraction than remove it.

src/pudl_archiver/archivers/eia861.py
jdangerx: Nope, this was a debug statement that fell through the `git add -p` cracks. Good catch!

PR #154, opened 2023-08-22
Partition types

zschira:
This PR allows non-string partition types. Making all partitions strings created some downstream errors when filtering by year.

### DIFF
--- a/src/pudl_archiver/archivers/validate.py
+++ b/src/pudl_archiver/archivers/validate.py
@@ -1,7 +1,7 @@
 """Defines models used for validating/summarizing an archiver run."""
 import zipfile
 from pathlib import Path
-from typing import Literal
+from typing import Any, Literal
 
 from pydantic import BaseModel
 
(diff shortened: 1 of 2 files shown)


jdangerx (approved):

rousik:
@zschira - can you elaborate on the nature of your problem with filtering by non string partitions? Because in `datastore.py` we cast both sides (parts as well as the value you search-by) to lowercase strings before matching, see https://github.com/catalyst-cooperative/pudl/blob/bfcf49530980d26856c34fd417fa0d4e9f24c946/src/pudl/workspace/datastore.py#L95

zschira:
Hey @rousik, the problem showed up in `src/pudl/extract/dbf.py` line 545:

```py
if self.is_valid_partition(p) and p.get("year", None) in self.settings.years
```

Since `settings.years` is a list of `int`s this is always false. We could do casting here, but I decided to change the archiver because the previous behavior allowed non-string types.

PR #161, opened 2023-09-05
Xbrl taxonomy

zschira:
This PR adjusts the layout of FERC XBRL archives. It moves the taxonomies to be their own resource instead of bundled with a year of filings.

### DIFF
--- a/src/pudl_archiver/archivers/ferc/ferc1.py
+++ b/src/pudl_archiver/archivers/ferc/ferc1.py
@@ -23,7 +23,20 @@ class Ferc1Archiver(AbstractDatasetArchiver):
         filings = xbrl.index_available_entries()[xbrl.FercForm.FORM_1]
         for year, year_filings in filings.items():
             if self.valid_year(year):
+                if year > 2019:
+                    yield self.get_year_taxonomy_xbrl(year)
                 yield self.get_year_xbrl(year, year_filings)
+
+    async def get_year_taxonomy_xbrl(self, year: int) -> tuple[Path, dict]:
+        """Download XBRL taxonomy from a single year."""
+        download_path = await xbrl.archive_taxonomy(
+            year, xbrl.FercForm.FORM_1, self.download_directory, self.session
+        )
+
+        return ResourceInfo(
+            local_path=download_path,
+            partitions={"year": year, "data_format": "XBRL_TAXONOMY"},
+        )
 
     async def get_year_xbrl(
         self, year: int, filings: xbrl.FormFilings
--- a/src/pudl_archiver/archivers/ferc/ferc1.py
+++ b/src/pudl_archiver/archivers/ferc/ferc1.py
@@ -24,19 +24,13 @@ class Ferc1Archiver(AbstractDatasetArchiver):
         for year, year_filings in filings.items():
             if self.valid_year(year):
                 if year > 2019:
-                    yield self.get_year_taxonomy_xbrl(year)
+                    yield xbrl.archive_taxonomy(
+                        year,
+                        xbrl.FercForm.FORM_1,
+                        self.download_directory,
+                        self.session,
+                    )
                 yield self.get_year_xbrl(year, year_filings)
-
-    async def get_year_taxonomy_xbrl(self, year: int) -> tuple[Path, dict]:
-        """Download XBRL taxonomy from a single year."""
-        download_path = await xbrl.archive_taxonomy(
-            year, xbrl.FercForm.FORM_1, self.download_directory, self.session
-        )
-
-        return ResourceInfo(
-            local_path=download_path,
-            partitions={"year": year, "data_format": "XBRL_TAXONOMY"},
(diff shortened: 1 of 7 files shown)


src/pudl_archiver/archivers/ferc/ferc1.py
```diff
@@ -23,8 +23,21 @@ async def get_resources(self) -> ArchiveAwaitable:
         filings = xbrl.index_available_entries()[xbrl.FercForm.FORM_1]
         for year, year_filings in filings.items():
             if self.valid_year(year):
+                if year > 2019:
+                    yield self.get_year_taxonomy_xbrl(year)
                 yield self.get_year_xbrl(year, year_filings)
 
+    async def get_year_taxonomy_xbrl(self, year: int) -> tuple[Path, dict]:
```
zschira, line 30: Could be nice to create a base XBRL archiver class that each of these can inherit from or have as a class member to avoid this duplicated logic.

jdangerx: Yeah, that would be nice :) Since the "getting a taxonomy" functionality isn't particularly form-specific, I would probably shunt that into a helper function that's not in any class, `ferc.xbrl.get_taxonomy(year, form, download_dir, session)` - and then include the download_dir and session at the callsite.

jdangerx (changes requested):
I'm glad you pushed to do this the right way :) I think we could probably benefit from asserting that we tried to download these taxonomies, sort of like how we assert we tried to `get_year_dbf` or `get_year_xbrl` in `ferc_archiver_test`:

```diff
diff --git a/tests/unit/ferc_archiver_test.py b/tests/unit/ferc_archiver_test.py
index 5a19151..98ee13e 100644
--- a/tests/unit/ferc_archiver_test.py
+++ b/tests/unit/ferc_archiver_test.py
@@ -72,5 +72,7 @@ async def test_valid_years(module_name, test_spec, mocker):
     # don't await these, just check to make sure they have right intention
     dbfs = [r for r in resources if r.__name__ == "get_year_dbf"]
     xbrls = [r for r in resources if r.__name__ == "get_year_xbrl"]
+    taxonomies = [r for r in resources if r.__name__ == "get_year_taxonomy_xbrl"]
     assert len(dbfs) == num_dbf
     assert len(xbrls) == num_xbrl
+    assert len(taxonomies) == num_xbrl
```

When I ran that, I found this error:

```bash
>       assert len(taxonomies) == num_xbrl                                                                                      
E       assert 1 == 2                                                                                                           
E        +  where 1 = len([<coroutine object Ferc1Archiver.get_year_taxonomy_xbrl at 0x15c13b920>])    
```

And I had set up the test data like this...
```python
    "ferc1": [
        {"years": [1848, 2349], "num_dbf": 0, "num_xbrl": 0},
        {"years": [2021, 2022, 2023], "num_dbf": 1, "num_xbrl": 1},
        {"years": [2003, 2004, 2005, 2021], "num_dbf": 4, "num_xbrl": 2},
    ],
```

Which was confusing at first but I think it's because, in the test, I set the XBRL available filings to 2004 and 2021. 

Which means that our `year > 2019` filters out the 2004 XBRL filing that the mocker is spitting out. 

By the time we're doing that `year > 2019` check, we already know that it's a year with an XBRL filing - do you think that check is still necessary? Or do we want to guard against the situation where they publish filings for a year that had no associated taxonomy?

I think we can either:
* remove that check and expect 2 taxonomies, or 
* keep the check in and just expect 1 taxonomy
* use more "realistic" data i.e. don't try to have xbrl filings for 2004. if we do this, I'd like to add some sort of error-throwing behavior if we see years < 2019 in the XBRL loop.

I think the most "correct" thing is looking for some sort of listing of all available taxonomies like whatever powers [this page](https://ecollection.ferc.gov/taxonomyHistory), but I'm not sure that's worth the time/money.

zaneselvans:
I think we **do** probably expect to have filings (though hopefully not taxonomies) for older years, since there seems to be an endless trickle of revised filings, which we eventually need to figure out a way of using to supersede the updated filing coming from the no longer updated DBF files for older years.  We have a bug for that but haven't looked into how to do it yet.

jdangerx:
That's what I thought too, which I guess pushes me towards "keep the check, expect 1 taxonomy."

src/pudl_archiver/archivers/ferc/ferc1.py
zschira: Good call, I was definitely overthinking this. I've just changed `archive_taxonomy` to return a `ResourceInfo` object.

zschira:
@jdangerx I adjusted the test, and had each test case just define the `num_taxonomy`, because it's 0 for many. We could also add a test case where `num_xbrl` is >0, but `num_taxonomy` is still 0 (cases where there are 2004 filings, but not 2021), but it didn't feel that important.

jdangerx (approved):

PR #169, opened 2023-09-14
Update to depending on PUDL development branch.

zaneselvans:
While trying to debug a dependency update CI failure, I noticed that we still had this repo pinned to depending on the `catalystcoop.pudl@pre-dagster` branch 😲 

So I did some updates:

* Point PUDL dependency at `catalsytcoop.pudl@dev`
* Use Python 3.11
* Switch to using the newer and maintained `mamba-org/setup-micromamba` action both in the `tox-pytest` and `run-archiver` workflows.
* Updated one test that was expecting an old FERC Form 1 title and not getting it.

### DIFF
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -11,6 +11,9 @@ on:
 
 jobs:
   archive-run:
+    defaults:
+      run:
+        shell: bash -l {0}
     strategy:
       matrix:
         dataset:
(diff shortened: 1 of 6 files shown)


zaneselvans:
@e-belfer I imagine that this pin to a pretty old version of PUDL was preventing your fresh new keywords from being available for use in your PHMSA archive creation?

e-belfer (approved):
Environment initialized correctly and resulted in an updated successful archive, looks good. Unit tests ran as expected.

PR #181, opened 2023-10-11
Rewrite CEMS archiver to use quarterly partitions rather than state-year.

e-belfer:
See [epic](https://github.com/catalyst-cooperative/pudl/issues/2902). Closes issue [#2929](https://github.com/catalyst-cooperative/pudl/issues/2929). In order to make quarterly updates of CEMS data possible, switch from archiving state-year data from the CEMS API to archiving quarterly data for all states and years, as it is released quarterly rather than annually.

Still to do:
- [x] Test using `conda-test` GCS.
- [x] Resolve `File size too large, try using force_zip64` zipfile error

Current run-time for the archiver is about ~1.5 hours on a GCE VM, with concurrency at 2 files to prevent memory overload. A successful sandbox archive can be seen [here](https://sandbox.zenodo.org/records/1770).

To generate a successful production archive this will need to be run with the "missing files" validation turned off in order to accommodate the change in file structure. Waiting for #177 to be merged first would provide additional validation of zipfile contents and improve the datapackage.

### DIFF
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -12,11 +12,7 @@ from pathlib import Path
 
 import aiohttp
 
-from pudl_archiver.archivers.validate import (
-    FileValidation,
-    RunSummary,
-    ValidationTestResult,
-)
+from pudl_archiver.archivers import validate
 from pudl_archiver.frictionless import DataPackage, ResourceInfo
 from pudl_archiver.utils import retry_async
 
@@ -95,7 +91,7 @@ class AbstractDatasetArchiver(ABC):
         if only_years is None:
             only_years = []
         self.only_years = only_years
-        self.file_validations: dict[str, FileValidation] = {}
+        self.file_validations: list[validate.FileSpecificValidation] = []
 
         # Create logger
         self.logger = logging.getLogger(f"catalystcoop.{__name__}")
@@ -225,7 +221,7 @@ class AbstractDatasetArchiver(ABC):
         self,
         baseline_datapackage: DataPackage | None,
         new_datapackage: DataPackage,
-    ) -> ValidationTestResult:
+    ) -> validate.DatasetSpecificValidation:
         """Check for any files from previous archive version missing in new version."""
         baseline_resources = set()
         if baseline_datapackage is not None:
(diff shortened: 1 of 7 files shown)


src/pudl_archiver/archivers/epacems.py
```diff
@@ -14,67 +13,13 @@
...
-    "wy",
-]
 
 
 class EpaCemsArchiver(AbstractDatasetArchiver):
     """EPA CEMS archiver."""
 
     name = "epacems"
-    concurrency_limit = 10  # Number of files to concurrently download
+    concurrency_limit = 2  # Number of files to concurrently download
```
zaneselvans, line 22: Did it end up just being impossible to run this archive via GitHub Actions? What's the bottleneck? Would we be able to run it on a 4CPU 16GB runner?

src/pudl_archiver/archivers/epacems.py
```diff
@@ -90,26 +35,28 @@ async def get_resources(self) -> ArchiveAwaitable:
...
+                yield self.get_quarter_year_resource(
+                    file=cems_file, request_count=i + 1
+                )  # Count files starting at 1 for human legibility.
         else:
             raise AssertionError(
                 f"EPACEMS API request did not succeed: {file_list.status_code}"
             )
 
-    async def get_state_year_resource(
+    async def get_quarter_year_resource(
```
zaneselvans, line 56: Saying "quarter_year" rather than "year_quarter" (which is the ordering in the filenames and ISO8601) seems a little confusing / unnatural.

src/pudl_archiver/archivers/epacems.py
```diff
@@ -119,26 +66,21 @@ async def get_state_year_resource(
...
+        filename = f"epacems-{year}-{quarter}.csv"
+        archive_path = self.download_directory / f"epacems-{year}-{quarter}.zip"
         # Override the default asyncio timeout to 14 minutes, just under the API limit.
         await self.download_and_zip_file(
             url=url, filename=filename, archive_path=archive_path, timeout=60 * 14
         )
-        logger.info(  # Verbose but helpful to keep track of progress.
-            f"File no. {request_count}: Downloaded {year} EPACEMS data for {state.upper()}"
+        logger.info(  # Verbose but helpful to track progress
+            f"File no. {request_count}: Downloaded Q{quarter} {year} EPA CEMS hourly emissions data."
```
zaneselvans, line 82: Would suggest year-quarter rather than quarter-year here too.

zaneselvans (approved):
I looked at the datapackage.json, and the archive metadata, and downloaded a file and opened it up, and everything looked good.

This isn't blocking, but I think it would be clearer if in the variable names, logging, and file names we always used "year-quarter" rather than mixing it with "quarter-year" sometimes, as this would match the desired sorting order, ISO8601 date formats, and the naming we're using on the files already.

Do you know if we're doomed to have to run this on a VM? Or is the resource use such that we could do it via a larger GitHub runner?

Weirdly, it looked like the files were getting sorted alphabetically in both the datapackage.json and the display of the archive on Zenodo.  Is that just chance?  It only looked sorted when viewing the record, not when editing it.

e-belfer:
> Weirdly, it looked like the files were getting sorted alphabetically in both the datapackage.json and the display of the archive on Zenodo. Is that just chance? It only looked sorted when viewing the record, not when editing it.

I manually sorted the files in the `datapackage.json` [here](https://github.com/catalyst-cooperative/pudl-archiver/pull/192/commits/ef926bc2a2d4b4f61b87202f3139f06ef162a90d), since in the present version of Zenodo the files automatically sort when you publish them (helpfully). We'll need to revisit this when the new API is released, not sure what the behavior will be like or if it will change.

src/pudl_archiver/archivers/epacems.py
e-belfer: I haven't tried yet, let me see! I think the Github action we'd want to use is getting defined in #197 but I've just tried to kick it off with the standard archiver.

Relatedly, @jdangerx our branches are in a bit of an intertwined state so let me know how you want to handle merge order in this instance?

Issue #183, opened 2023-10-16
Move to new Zenodo API


jdangerx:
See also https://github.com/catalyst-cooperative/pudl/issues/2939 -

Zenodo migrated to their new API, we need to use it if we want to keep creating new archives.

* use `f"/records/{record_number}/files/{name}"` in the datapackage.json
* hit the correct endpoints in the `ZenodoDepositor` - I think the same *methods* need to exist (create/get/publish/delete deposition/version/record/file) but we should look more into this.
* upgrade to `frictionless>=5.0`?

zaneselvans:
Automatically adding the `catalyst-cooperative` community would be great -- or maybe, add it if the previous version of the record was in the community. Right now the fact that a record is in a community doesn't seem to be inherited by subsequent versions which is very annoying, and leads to the most recent records often not showing up.

zaneselvans:
I notice that we do not list `frictionless` as one of the dependencies in this module, but it does get installed, because we depend on `catalystcoop.pudl` which depends on `catalystcoop.ferc-xbrl-extractor` which depends on `frictionless`.

We also don't *use* it in this module, even though we redefine classes which mirror the `Resource` and `DataPackage` concepts. This seems like a recipe for ending up with outputs that look like data packages, but don't actually conform to the specification.

jdangerx:
Ooooh yeah. I noticed that dependency funniness but did not connect the dots with "we don't actually use `frictionless` for data packages." We *definitely* should update that - could be a separate PR, though.

e-belfer:
According to email communications with Zenodo this API will not be launched for another year and it's behavior is expected to change. So let's pause on this.

PR #184, opened 2023-10-19
Icebox: Migrate to new Zenodo InvenioRDM API

e-belfer:
See #183 for detailed issue description. Zenodo's endpoints for their API have changed, breaking our existing archivers. Let's get them migrated.

- [x] update `create_deposition`
- [x] update `get_deposition`
- [x] update `get_record`
- [x] update `get_new_version`
- [x] update `publish_deposition`
- [x] update `delete_deposition`
- [x] update `create_file`
- [x] update `delete_file`
- [x] update `update_file`
- [x] update `datapackage.json` to use new path
- [x] Update `creators` metadata to match new format https://inveniordm.docs.cern.ch/reference/metadata/#creators-1-n
- [x] fix license https://inveniordm.docs.cern.ch/reference/metadata/#rights-licenses-0-n
- [x] deal with key names that mask python methods
- [x] debug `delete_deposition` 
- [x] update Zenodo sandbox APIs to `.5281` prefix and new values
- [x] check if old API still works - it does not
- [x] check to make sure new datapackage format for newly created archives is the same as the old version
- [x] fix tests to expect different response than what is posted for metadata
- [x] Drop `/content` extension from datapackage
- [x] Figure out how to deal with metadata updates from existing production archives - add `refresh_metadata` flag
- [x] fix datapackage.json to remove content headers
- [ ] once reviewed: make new sandbox archives for: `census_dp1tract`, `eia176`, `eia860`, `eiawater`, `epacems`, `ferc60` and update sandbox dois for all data sources

Out of scope:
- add community to post https://inveniordm.docs.cern.ch/reference/metadata/#future-directions (maybe not yet possible?)

Check that the code works for:
- [x] non-flagged update run
- [x] dataset with partitions
- [x] `--sandbox`
- [x] `--dry-run`
- [x] `--initialize`

Updates about remaining kinks in the transition:
- When testing `--auto-publish` on the sandbox server - I get a "Attempt to decode JSON with unexpected mimetype: text/html" 404 error. Underlying this is a 504 Gateway Timeout for the sandbox. This also means the tests fail.
- The [API documentation](https://inveniordm.docs.cern.ch/reference/rest_api_drafts_records/#download-a-draft-file) says to use the /content file link to GET a file. There are two options here: "https://zenodo.org/api/records/10064652/files/eia923-2023.zip/content" and "https://zenodo.org/records/10064652/files/eia923-2023.zip" - presuming in this PR that we prefer the latter?
- Sandbox DOIs are unstable and likely to change again. It's still unclear whether the transition from 10.5072 to 10.5281 for sandbox DOIs is permanent. Sandbox DOIs are not actually being really registered right now, based on email communication with Zenodo technical support.
- The `creators` field response != the `creators` field expected format. This means you can't `GET` a deposition and `PUT` its metadata back. Updating a draft record with one field (e.g. the `order` field for the `files` dictionary) also seems to wipe all the rest of the inputted values for the `metadata`. Very annoying behavior that means we should avoid updating records by any means other than linking once they're created.

### DIFF
--- a/src/pudl_archiver/cli.py
+++ b/src/pudl_archiver/cli.py
@@ -65,6 +65,11 @@ def parse_main():
         help="Directory to download files to. Use tmpdir if not specified.",
         default=None,
     )
+    parser.add_argument(
+        "--refresh-metadata",
+        action="store_true",
+        help="Regenerate metadata from PUDL data source rather than existing archived metadata.",
+    )
     return parser.parse_args()
 
 
--- a/src/pudl_archiver/depositors/zenodo.py
+++ b/src/pudl_archiver/depositors/zenodo.py
@@ -145,8 +145,8 @@ class ZenodoDepositor:
                 "access": {
                     "record": "public",
                     "files": "public",
-                },  # TO DO: don't hardcode?
-                "files": {"enabled": True},
+                },
+                "files": {"enabled": True},  # Must always be True for Zenodo
                 "metadata": metadata,
             }
         )
@@ -202,7 +202,7 @@ class ZenodoDepositor:
             params=params,
             headers=self.auth_write,
         )
-        logger.info(response)
+
         if response["hits"]["total"] > 1:
             logger.info(
                 f"{conceptid} points at multiple records: {[r['id'] for r in response['hits']['hits']]}"
@@ -244,7 +244,11 @@ class ZenodoDepositor:
         return Deposition(**response)
 
     async def get_new_version(
-        self, deposition: Deposition, clobber: bool = False
+        self,
+        deposition: Deposition,
+        data_source_id: str,
+        clobber: bool = False,
+        refresh_metadata: bool = False,
     ) -> Deposition:
         """Get a new version of a deposition.
 
@@ -254,6 +258,9 @@ class ZenodoDepositor:
         Args:
             deposition: the deposition you want to get the new version of.
             clobber: if there is an existing draft, delete it and get a new one.
+            data_source_id: the deposition dataset name (to clobber existing metadata.)
+            refresh_metadata: regenerate metadata from data source rather than existing
+                archives' metadata.
 
         Returns:
             A new Deposition that is a snapshot of the old one you passed in,
@@ -281,7 +288,14 @@ class ZenodoDepositor:
 
         # Update metadata from previous deposition.
         old_metadata = deposition.metadata.dict(by_alias=True)
-        new_metadata = new_deposition.metadata.dict(by_alias=True)
+
+        if refresh_metadata:
+            logger.info("Re-creating deposition metadata from PUDL source data.")
+            new_metadata = DepositionMetadata.from_data_source(data_source_id).dict(
+                by_alias=True
+            )
+        else:
+            new_metadata = new_deposition.metadata.dict(by_alias=True)
 
         # Update version number from previous metadata.
         previous = semantic_version.Version(old_metadata["version"])
@@ -328,6 +342,7 @@ class ZenodoDepositor:
             log_label="Creating new version.",
             headers=headers,
         )
+        logger.info(response)
         return DepositionVersion(**response)
 
     async def link_previous_files(
@@ -353,11 +368,13 @@ class ZenodoDepositor:
         """
         url = deposition.links.publish
         headers = {
+            "Accept": "application/vnd.inveniordm.v1+json",
             "Content-Type": "application/json",
         } | self.auth_actions
         response = await self.request(
             "POST", url, log_label="Publishing deposition", headers=headers
         )
+        logger.info(response)
         return Deposition(**response)
 
     async def delete_deposition(self, deposition) -> None:
--- a/src/pudl_archiver/depositors/zenodo.py
+++ b/src/pudl_archiver/depositors/zenodo.py
@@ -357,7 +357,7 @@ class ZenodoDepositor:
             headers=self.auth_write,
         )
 
-    async def publish_deposition(self, deposition: Deposition) -> Deposition:
+    async def publish_deposition(self, deposition: Deposition) -> DepositionVersion:
         """Publish a deposition.
 
         Needs to have at least one file, and needs to not already be published.
@@ -373,7 +373,8 @@ class ZenodoDepositor:
         response = await self.request(
             "POST", url, log_label="Publishing deposition", headers=headers
         )
-        return Deposition(**response)
+        logger.info(response)
+        return DepositionVersion(**response)
 
     async def delete_deposition(self, deposition) -> None:
         """Delete an un-submitted deposition.
--- a/src/pudl_archiver/depositors/zenodo.py
+++ b/src/pudl_archiver/depositors/zenodo.py
@@ -172,7 +172,7 @@ class ZenodoDepositor:
         return response
 
     async def get_deposition(
-        self, conceptrecid: str, published_only: bool = False
+        self, conceptid: str, published_only: bool = False
     ) -> Deposition:
         """Get the latest deposition associated with a concept DOI.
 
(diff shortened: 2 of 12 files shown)


src/pudl_archiver/depositors/zenodo.py
```diff
@@ -122,7 +125,7 @@ async def create_deposition(self, metadata: DepositionMetadata) -> Deposition:
         Returns:
             The brand new deposition - unpublished and with no files in it.
         """
-        url = f"{self.api_root}/deposit/depositions"
+        url = f"{self.api_root}/records"
```
e-belfer, line 128: To do: add `records` to the `api_root` rather than specifying it, since every endpoint seems to refer to it.

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -324,44 +321,97 @@ async def create_file(
...
+    async def start_file_upload(
+        self,
+        deposition: Deposition,
+        target: str,
+    ) -> None:
+        """Start a file deposition.
 
         Args:
             deposition: the deposition you are applying this change to.
             target: the filename of the file you want to delete.
```
e-belfer, line 348: should be "create" not delete

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -219,46 +222,40 @@ async def get_new_version(
...
-                metadata[key] = val
-
-        previous = semantic_version.Version(source_metadata["version"])
-        version_info = previous.next_major()
-
-        metadata["version"] = str(version_info)
-
-        # Update metadata of new deposition with new version info
-        data = json.dumps({"metadata": metadata})
-
```
e-belfer, line 247: Add this whole section back as these are removed by the `create a new version` endpoint.

src/pudl_archiver/zenodo/entities.py
```diff
@@ -38,19 +35,169 @@ class SandboxDoi(ConstrainedStr):
...
+        else:
+            person_or_org = Organization.from_contributor(contributor)  # Debug
+        return cls(
+            person_or_org=person_or_org,
+            affiliations=[Affiliations.from_contributor(contributor)],
+            role=Role.from_contributor(contributor),
+        )
+
+    @classmethod
+    def from_response(cls, contributor: Contributor) -> "DepositionCreator":
```
e-belfer, line 167: This is outdated and duplicated, remove.

src/pudl_archiver/orchestrator.py
```diff
@@ -239,7 +244,9 @@ async def run(self) -> RunSummary | Unchanged:
             for name in files_to_delete
         ]
 
-        self.new_deposition = await self.depositor.get_record(self.new_deposition.id_)
+        self.new_deposition = await self.depositor.get_draft_record(
```
jdangerx, line 247: I'm pleased that the changes to `orchestrator.py` are so light / the interface to the depositor didn't change much!

Once Zenodo stabilizes and the docs exist, we can maybe think about what the depositor interface should look like going forward... something along the lines of:

* create new deposition
* update the deposition file contents (all the draft state management might want to be hidden from the outside)
* publish the deposition

But - since we're most likely going to be tied to Zenodo for the forseeable future, and I don't expect them to be changing the API willy-nilly anytime soon, maybe that's not a super useful refactor.

jdangerx (commented):
So you've successfully made archives in prod. Which is good, and means we have probably one of the few functioning Zenodo API clients in the world.

However, tests are failing, and there are some weird data modelling issues e.g. `DepositionVersion` not having a `submitted` field, which don't feel ready for merge into `main`.

There's also just lots of undocumented behavior that we're using, but I'm not sure is necessary, but we can't really tell without the documentation or lots of trial and error.

I think we have a few options here:
* merge this into `main`, and make a big cleanup pass later when Zenodo is stabilized
* keep this un-merged, wait for Zenodo to stabilize, and then make a cleanup pass before merging; in the meantime, do all manual archive creation on this branch
* keep this un-merged, but investigate whether the legacy API integration that's currently on `main` works again - they had mentioned trying to get it back online and it appears to be at least somewhat back.
* refactor this to make a "new zenodo API depositor" that shares the same public API as the legacy zenodo API depositor and lives alongside it (instead of replacing the existing zenodo depositor), so we can switch between the two as Zenodo figures out what's happening

I think that if we can't make prod archives using the legacy API integration on `main` and we can make prod archives with this version, we should just merge into `main` and fix things once zenodo calms down a bit.

It would maybe make sense to wait a few more days and see if we can get sandbox tests passing. It looks like the records are now being successfully published on sandbox, even though the publish request returns a 500. So maybe in a few days the sandbox will be less flaky and we can actually try to run these tests.

src/pudl_archiver/cli.py
```diff
@@ -65,6 +65,11 @@ def parse_main():
         help="Directory to download files to. Use tmpdir if not specified.",
         default=None,
     )
+    parser.add_argument(
+        "--refresh-metadata",
+        action="store_true",
+        help="Regenerate metadata from PUDL data source rather than existing archived metadata.",
```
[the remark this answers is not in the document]
e-belfer: Yes, creators, title, keywords, basically all metadata. The outcome is the same as if we'd run `initialize` except for version and link to the pre-existing depositions.

src/pudl_archiver/zenodo/entities.py
```diff
@@ -38,19 +35,156 @@ class SandboxDoi(ConstrainedStr):
...
+                "project member",
+                "registration agency",
+                "registration authority",
+                "related person",
+                "researcher",
+                "rights holder",
+                "sponsor",
+                "supervisor",
+                "work package leader",
+            ]:  # Unclear to me what roles are allowed here.
```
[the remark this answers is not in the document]
e-belfer: By manually copying over the list of options on the Zenodo site, essentially.

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -122,26 +127,52 @@ async def create_deposition(self, metadata: DepositionMetadata) -> Deposition:
...
+        )
+        metadata["publication_date"] = str(date.today())
+
+        payload = json.dumps(
+            {
+                "access": {
+                    "record": "public",
+                    "files": "public",
+                },
+                "files": {"enabled": True},  # Must always be True for Zenodo
```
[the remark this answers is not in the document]
e-belfer: It throws an error if True, not documented anywhere. As far as I can tell it's technically a setting that the network configurator can change to allow metadata only records but this doesn't seem to be enabled in Zenodo. This is ok since we only ever want to upload records with files. It also matches the behavior of old Zenodo API (see the expected failure for the `empty_deposition` test, e.g.

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -122,26 +127,52 @@ async def create_deposition(self, metadata: DepositionMetadata) -> Deposition:
...
+                "metadata": metadata,
+            }
+        )
         response = await self.request(
-            "POST", url, "Create new deposition", json=payload, headers=headers
+            "POST", url, "Create new deposition", data=payload, headers=headers
         )
+        # Reserve DOI for deposition
+        doi_response = await self.reserve_doi(response)
```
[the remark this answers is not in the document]
e-belfer: I followed the `"reserve_doi"` link returned in the list of links for a file. It's not listed anywhere other than being described as an option for developers using InvenioRDM to enable. Made some informed guesses about what it would take based on [this publish method](https://inveniordm.docs.cern.ch/reference/rest_api_drafts_records/#publish-a-draft-record) that wound up being right, if I remember right.

jdangerx:
We'll close this for now since we're not actively working on it for a minute and will probably have to redo a bunch of stuff anyways. But, this will be super useful documentation for when they finally do release the new API.

PR #195, opened 2023-11-17
Set up Ruff and fix dependency conflicts with PUDL repo.

zaneselvans:
I went in to try and investigate the inclusion of draft URLs in the `datapackage.json` metadata, and discovered that there were conflicts between the dependencies in `pudl-archiver` and the main PUDL repo, so it could never solve the environment.

While resolving that I went ahead and switched over to using the Ruff linter in this repo and fixed some issues it highlighted.

There's still one failing unit test though, which is maybe related to some mocking? Which I don't really understand. I tried to switch it from mocking `open` to mocking `Path.open` which seems to have worked, but this seems to have messed up another call to open. :-/ 

```
pytest tests/unit/pudl_archiver_test.py::test_archive_datasets
```

I tried to do a dry-run of archiving the `eia861` to see if the draft paths were still there, but it failed with a bunch of other problems.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,5 +1,5 @@
 ---
-name: pudl-cataloger
+name: pudl-archiver
 channels:
   - conda-forge
   - defaults
--- a/environment.yml
+++ b/environment.yml
@@ -1,5 +1,5 @@
 ---
-name: pudl-archiver
+name: pudl-cataloger
 channels:
   - conda-forge
   - defaults
--- a/src/pudl_archiver/orchestrator.py
+++ b/src/pudl_archiver/orchestrator.py
@@ -74,8 +74,8 @@ class DatasetSettings(BaseModel):
 
 def _compute_md5(file_path: Path) -> str:
     """Compute an md5 checksum to compare to files in zenodo deposition."""
-    hash_md5 = md5()  # nosec: B324
-    with open(file_path, "rb") as f:
+    hash_md5 = md5()  # noqa: S324
+    with Path.open(file_path, "rb") as f:
         for chunk in iter(lambda: f.read(4096), b""):
             hash_md5.update(chunk)
 
@@ -137,7 +137,7 @@ class DepositionOrchestrator:
 
         self.create_new = create_new
         self.deposition_settings = deposition_settings
-        with open(deposition_settings) as f:
+        with Path.open(deposition_settings) as f:
             self.dataset_settings = {
                 name: DatasetSettings(**dois)
                 for name, dois in yaml.safe_load(f).items()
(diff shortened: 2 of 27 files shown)


tests/unit/pudl_archiver_test.py
```diff
@@ -50,7 +49,7 @@ def failed_run():
 @pytest.mark.asyncio
 async def test_archive_datasets(successful_run, failed_run, unchanged_run, mocker):
     """Test that archive datasets creates run_summary file."""
-    open_mock = mocker.patch("pudl_archiver.open")
+    open_mock = mocker.patch("pudl_archiver.Path.open")
```
zaneselvans, line 52: This is the mocking that I think is related to the one unit test failure.

src/pudl_archiver/orchestrator.py
```diff
@@ -137,7 +137,7 @@ def __init__(
 
         self.create_new = create_new
         self.deposition_settings = deposition_settings
-        with open(deposition_settings) as f:
+        with Path.open(deposition_settings) as f:
             self.dataset_settings = {
                 name: DatasetSettings(**dois)
                 for name, dois in yaml.safe_load(f).items()
```
zaneselvans, line 143: I don't understand why, but switching to using `Path.open()` and then mocking that `Path.open()` in the tests seems to cause this call to `yaml.safe_load()` to get a string rather than a dictionary.

tests/unit/pudl_archiver_test.py
jdangerx: The problem here is that `open` is a function that returns file contents, and `Path.open` is an instance method - so you need to create a whole `Path` object and then mock _its_ `.open()` method.

jdangerx: I was wrong about this - the mocking is actually working fine as is, but since we pass in a string (`"file"`) to `yaml.safe_load()` it returns a string.

for example:

```python
>>> import yaml
>>> yaml.safe_load("file")
'file'
>>> yaml.safe_load("- file")
['file']
>>> yaml.safe_load("file: a")
{'file': 'a'}
```

So we need to give it better fake data. I'm not sure why it worked before...

zaneselvans: Is it really supposed to be affecting the `yaml.safe_load()` at all?  It seemed like a far-flung side effect of the mocking -- which maybe makes sense if I was accidentally overwriting the classmethod `Path.open()` rather than a particular instance method.

zaneselvans: In the old setup, it was just overriding a the `open()` function's behavior in a single module (`pudl_archiver.open`) and not having any effects elsewhere... which seemed like the desired behavior.

jdangerx (approved):
This all looks good modulo the weird patching stuff. Thanks for wrangling the dependencies!

I think it's better to do `path_instance.open()` rather than `Path.open()` since that gives you more control over the individual paths when testing, as seen in my diff.

environment.yml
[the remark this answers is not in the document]
zaneselvans: Can we check whether this is still happening? I was hoping that the bug had been fixed and we could go back to having the environment names match the repository name.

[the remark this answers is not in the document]
jdangerx: PR is still unmerged & bug is still open: https://github.com/python/cpython/pull/32178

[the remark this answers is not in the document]
zaneselvans: Oof, okay. Wild.

PR #197, opened 2023-11-21
Update schedule to be quarterly

jdangerx:
Also, only archive the specific datasets requested by RMI... should be easy to add the rest back in later, but for now I want to keep the system small.

Closes catalyst-cooperative/pudl#3066

### DIFF
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -4,10 +4,7 @@ name: run-archiver
 on:
   workflow_dispatch:
   schedule:
-    - cron: "21 4 * * 0" # Sundays 4:21 AM UTC
-  push:
-    branches:
-      - "daz/**"
+    - cron: "21 4 1 */3 *" # 4:21 AM UTC, 1st of Jan/Apr/Jul/Oct
 
 jobs:
   archive-run:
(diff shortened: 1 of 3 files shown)


tests/unit/pudl_archiver_test.py
```diff
@@ -55,6 +55,14 @@ async def test_archive_datasets(
     tmp_path,
 ):
     """Test that archive datasets creates run_summary file."""
+    mocker.patch.dict(
```
jdangerx, line 58: I accidentally ran this test without the corresponding environment variables set, and figured it best to mock them here since it's a unit test (vs. an integration test which should probably expect those env vars).

.github/workflows/run-archiver.yml
```diff
@@ -4,10 +4,7 @@ name: run-archiver
 on:
   workflow_dispatch:
   schedule:
-    - cron: "21 4 * * 0" # Sundays 4:21 AM UTC
-  push:
-    branches:
-      - "daz/**"
+    - cron: "21 4 1 */3 *" # 4:21 AM UTC, 1st of Jan/Apr/Jul/Oct
```
e-belfer, line 7: Will confirm with RMI what the schedule should be, records are typically updated towards the end of the month.

.github/workflows/run-archiver.yml
```diff
@@ -17,24 +14,10 @@ jobs:
     strategy:
       matrix:
         dataset:
-          - censusdp1tract
-          - eia176
           - eia860
```
e-belfer, line 17: 860 will almost always report no change except for when annual data is updated, but I think it's ok to run once a quarter anyways!

.github/workflows/run-archiver.yml
zaneselvans: I think we might want to consider oversampling on the archives -- e.g. if we want to actually have workable quarterly archives, trying to generate new archives each month instead.

This would:

* reduce the chances of making an archive right before the data gets updated, and then having to wait a whole quarter or do a manual archiving run to get the fresh data.
* give us an early heads-up when something is broken, so we don't end up shocked that there's a lot of work to do when we get around to trying to do the quarterly update.
* make the system more robust to transient failures due to network hiccups, etc.
* provide a more granular versioned public record of the evolution of these data sources, which are often updated with little or no public explanation.

PR #200, opened 2023-11-22
PHMSA archiver: add multi-year years partition, add new dataset

e-belfer:
Adds a multi-year `years` partition to the datapackage which includes a list of all years included in a multi-year ZIP file, and adds a new dataset while we're at it.

See sample sandbox archive [here](https://sandbox.zenodo.org/records/1981).

- [x] PUDL `datastore` can ingest listed partitions without breakdown
- [x] Change `dataset` partition to `form` to avoid `datastore` naming conflicts

After #177 merged, we should run a production archive and check that `datapackage.json` is valid using `frictionless`.

### DIFF
--- a/dataset_doi.yaml
+++ b/dataset_doi.yaml
@@ -48,4 +48,4 @@ mshamines:
   sandbox_doi: 10.5072/zenodo.1158828
 phmsagas:
   production_doi: 10.5281/zenodo.7683351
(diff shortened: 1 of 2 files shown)


cmgosnell (changes requested):
everything looks good except one hm idk question and one small suggestion.

also also, is `start_year` as a partition still necessary with the addition of `years`?

Issue #76, opened 2023-02-27
Add all new archives to the Catalyst Cooperative community


zaneselvans:
We have a "community" on Zenodo that collects all of our published archives together at: https://zenodo.org/communities/catalyst-cooperative/

Currently new archives have to be added to the community manually, but we should automate the process so all of our data can be found more easily by users.

Unfortunately Zenodo currently only allows a single community curator, and the account associated with it belongs to @zaneselvans rather than the PUDL Bot which owns all of the datastore archives, so Zane will have to approve requests to join the community. This issue will be fixed in a future version of the Zenodo backend.

e-belfer:
Should we incorporate this into our 2022 new data integration blitz? What needs to be done here to finish out this issue @zaneselvans ?

zaneselvans:
I think that there should be some way to use the Zenodo API to add a new record to the `catalyst-cooperative` community when it's created or right after it's been created.  Then I'll have to go approve it since I own the community and there's still no way to share or transfer curation duties.  I think @zschira is most familiar with the Zenodo API at this point.

e-belfer:
The Zenodo migration to InvenioRDM has changed the way that communities work, making this addition simpler from the API. I'm picking this issue back up, should be relatively trivial to add communities to our default `DepositionMetadata`

zaneselvans:
Oh great, this is such a hassle right now. Thank you for picking it up!

e-belfer:
In the current iteration of Zenodo, it's not possible to add a record to a community using the API and then manually approve it using the GUI - records are only adding through the GUI or when the `--auto-publish` flag forces direct publication through the API. Since this doesn't address our needs right now I am closing this issue but will merge #202 with just the `--refresh-metadata` functionality.

PR #202, opened 2023-11-23
Add flag to refresh metadata for existing depositions

e-belfer:
Sometimes we change a data source's metadata but when updating a dataset, and wish to update the corresponding Zenodo metadata for an existing archive. This PR adds a `refresh_metadata` flag to allow updates to record metadata rather than inheriting it from previous archives.

This initially began as an effort to add communities automatically to archives, but this is impossible in our current workflow due to Zenodo constraints (see below and #76).

See for example this [861 draft archive](https://zenodo.org/uploads/10425463) updated using `--refresh-metadata` and compare it to previous versions.


e-belfer:
Sticking points:
- When looking at the requests and responses on this branch for `create_version`, the communities identifier is constructed as expected and present in the Zenodo response, suggesting the formatting is correct. However this doesn't result in an addition to the community. This was working only with the test records before, but I've now removed test records from the community since it's a lot of additional noise.

e-belfer:
As of today, this code works when using the `auto-publish` flag but does not add a record to the community if manually reviewing and publishing, despite noting in the record metadata that it is part of the community. This is a bug on Zenodo's end that I'm emailing to learn more about how to resolve.

zschira (approved):
Looks good!

PR #203, opened 2023-11-24
Prod Zenodo tweaks

jdangerx:
* handling deposition deletion flakiness a little more gracefully
* wrongly assumed that zenodo URL would always have a subdomain e.g. `https://www.zenodo.org` - but instead sometimes it's `https://zenodo.org`. Fixed regex.
* update test to allow for slow updates on zenodo side

### DIFF
--- a/src/pudl_archiver/depositors/zenodo.py
+++ b/src/pudl_archiver/depositors/zenodo.py
@@ -74,6 +74,7 @@ class ZenodoDepositor:
             url: str,
             log_label: str,
             parse_json: bool = True,
+            retry_count: int = 7,
             **kwargs,
         ) -> dict | aiohttp.ClientResponse:
             """Make requests to Zenodo.
@@ -117,6 +118,7 @@ class ZenodoDepositor:
                     asyncio.TimeoutError,
                     ZenodoClientError,
                 ),
+                retry_count=retry_count,
             )
             return response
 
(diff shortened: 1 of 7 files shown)


src/pudl_archiver/depositors/zenodo.py
```diff
@@ -117,6 +118,7 @@ async def run_request():
                     asyncio.TimeoutError,
                     ZenodoClientError,
                 ),
+                retry_count=retry_count,
```
e-belfer, line 121: Just confirming, you're basically passing the option here to lower the default retry count from 7 to 5 for deletions? What's the motivation for doing this?

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -316,16 +318,29 @@ async def publish_deposition(self, deposition: Deposition) -> Deposition:
...
+                log_label="Deleting deposition",
+                headers=self.auth_write,
+                parse_json=False,
+                retry_count=5,
+            )
+        except ZenodoClientError as e:
+            if e.status != 404:
+                raise e
+            logger.info(
+                f"Got 404 when deleting {deposition.links.self}, assuming "
```
e-belfer, line 341: We could also look for 'persistent identifier does not exist' in the error message which I think would be a clearer determination that the deletion had occurred?

e-belfer (approved):
I have a few non-blocking suggestions and a clarifying question. Otherwise the tests all pass locally on my end and I'm able to make a sandbox archive without trouble.

src/pudl_archiver/depositors/zenodo.py
jdangerx: The motivation is impatience - it seems to typically resolve within 5 retries, so I don't want to wait an extra 90 seconds while waiting for the deletion to register.

src/pudl_archiver/depositors/zenodo.py
jdangerx: Yeah, I think that's easier to read but also the HTTP status 404 is less likely to change in the future.

src/pudl_archiver/depositors/zenodo.py
e-belfer: Fair enough!

src/pudl_archiver/zenodo/entities.py
```diff
@@ -122,11 +122,13 @@ def canonical(self):
         *.zenodo.org/records/<record_id>/files/<filename>
```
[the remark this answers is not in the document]
jdangerx: Sure thing!

src/pudl_archiver/depositors/zenodo.py
e-belfer: Maybe just printing the error message would be helpful then, in case we see a very different 404 error message that suggests non-deletion and we want to revise our assumption.

PR #205, opened 2023-11-24
Update datapackage.json, but only sometimes

jdangerx:
Sometimes, e.g. when you've updated how you generate the `path` and `remote_url` fields of `datapackage.json`, you want to update `datapackage.json` EVEN THOUGH there were no changes to the files themselves.

Other times, you don't want to update `datapackage.json` because the difference between the old and new `datapackage.json` is just creation time / version number / deposition ID incrementing between the two versions.

Made some refactoring / cleanup changes while I was in here banging my head against how and when to do things:

* in-lined some very straightforward methods that were only being used in one place
* broke out helper functions from `run()` body to better communicate the flow of the orchestrator
* stopped bailing out immediately after changeset generation on dry run, in an attempt to allow dry-run to go through more of the logic. Unfortunately, datapackage creation is contingent on deposition state change on Zenodo (we want to get the path, remote_url, hash, and file-size to fully flesh out a `Resource` in a datapackage, and none of that is really captured in `ResourceInfo` so we get it from `DepositionFile`.) - so `--dry-run` now dumps a bunch of silliness into the summary file. I wonder if we even still need it now that `--auto-publish` defaults to `False`.

Also, did some funny stuff in test:
* sometimes Zenodo is fine with trying to download the "canonical" URL... that they told us about... other times it throws a tantrum (i.e., 500 error), and refuses to accept anything but the `file.links.download` link. So I changed the test "download a file to verify its contents" method to handle both 🙄 
* added another scenario in the mega-integration test that patches out the canonical URL to something silly to check that datapackage.json does, in fact, get updated

Finally, I did the following to test manually:
1.`pudl_archiver --datasets eia861 --summary-file eia861-summary.json --sandbox` - the resources in the old `datapackage.json` were empty, so this updated the sandbox version to a fully well-formed archive.
2. publish manually,
3. run `pudl_archiver --datasets eia861 --summary-file eia861-summary.json --sandbox` again - got "No changes detected".
4. update the `FileLinks.canonical()` method to append `"bogus"` to each canonical URL.
5. Re-run `pudl_archiver --datasets eia861 --summary-file eia861-summary.json --sandbox` - now I see that the new version has the new datapackage.json. The run summary shows no file changes, as expected
6. delete the bogus draft to avoid confusion.

There's still some stuff I'd like to clean up, but I figured that is going out of the scope of "make it so that we can publish a `datapackage.json`-only change to Zenodo."
* The `Unchanged`/`RunSummary` split seems like it introduces more complexity than it removes - A RunSummary summarizes the changes in a run, and it's easy to check if `file_changes` is empty
* I'd love to be able to generate the `datapackage.json` from just `resources` dictionary so that the dry-run outputs a better summary

### DIFF
--- a/src/pudl_archiver/orchestrator.py
+++ b/src/pudl_archiver/orchestrator.py
@@ -183,7 +183,7 @@ class DepositionOrchestrator:
         """
         await self._initialize()
 
-        resources = await self._upload_resources(self.downloader)
+        resources = await self._download_then_upload_resources(self.downloader)
         for deletion in self._get_deletions(resources):
             await self._apply_change(deletion)
 
@@ -207,7 +207,7 @@ class DepositionOrchestrator:
 
         return run_summary
 
-    async def _upload_resources(
+    async def _download_then_upload_resources(
         self, downloader: AbstractDatasetArchiver
     ) -> dict[str, ResourceInfo]:
         resources = {}
--- a/src/pudl_archiver/orchestrator.py
+++ b/src/pudl_archiver/orchestrator.py
@@ -1,6 +1,7 @@
 """Core routines for archiving raw data packages."""
 import io
 import logging
+import re
 from dataclasses import dataclass
 from enum import Enum, auto
 from hashlib import md5
@@ -146,7 +147,12 @@ class DepositionOrchestrator:
     async def _initialize(self):
         if self.create_new:
             doi = None
-            self.deposition = await self._create_deposition(self.data_source_id)
+            metadata = DepositionMetadata.from_data_source(self.data_source_id)
+            if not metadata.keywords:
+                raise AssertionError(
+                    "New dataset is missing keywords and cannot be archived."
+                )
+            self.deposition = await self.depositor.create_deposition(metadata)
         else:
             settings = self.dataset_settings[self.data_source_id]
             doi = settings.sandbox_doi if self.sandbox else settings.production_doi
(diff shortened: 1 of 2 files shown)


src/pudl_archiver/orchestrator.py
```diff
@@ -160,41 +166,8 @@ async def _initialize(self):
...
-            Deposition object, per
-            https://developers.zenodo.org/?python#depositions
-        """
-        metadata = DepositionMetadata.from_data_source(data_source_id)
-        if not metadata.keywords:
-            raise AssertionError(
-                "New dataset is missing keywords and cannot be archived."
-            )
-        return await self.depositor.create_deposition(metadata)
+        self.deposition_files = {f.filename: f for f in self.new_deposition.files}
```
jdangerx, line 169: We could probably start passing the actual deposition objects around since we want them to be *run*-scoped, not *orchestrator*-scoped. But that seems out of scope for this PR.

src/pudl_archiver/orchestrator.py
```diff
@@ -292,18 +248,24 @@ def _generate_changes(
...
+                files_to_delete.append(
+                    _DepositionChange(_DepositionAction.DELETE, name=filename)
+                )
 
         return files_to_delete
 
     async def _apply_change(self, change: _DepositionChange):
         """Actually upload and delete what we listed in self.uploads/deletes."""
+        self.changes.append(change)
+        if self.dry_run:
```
jdangerx, line 266: Thought it would be good to move the dry-run checks closer to the actual write operations, but I didn't feel like making a whole bunch of changes to the depositor (or write a dry-run specific depositor). So here we are.

src/pudl_archiver/orchestrator.py
```diff
@@ -402,12 +357,50 @@ async def _update_datapackage(self, resources: dict[str, ResourceInfo]):
...
+        self, old_datapackage: DataPackage, new_datapackage: DataPackage
+    ) -> bool:
+        # ignore differences in created/version
+        # ignore differences resource paths if it's just some ID number changing...
+        for field in new_datapackage.dict():
+            if field in {"created", "version"}:
+                continue
+            if field == "resources":
+                for r in old_datapackage.resources + new_datapackage.resources:
+                    r.path = re.sub(r"/\d+/", "/ID_NUMBER/", str(r.path))
```
jdangerx, line 390: I'm not happy about this heuristic for "some ID number has changed, please ignore," but I'm also not sure what a better version of this would look like.

src/pudl_archiver/orchestrator.py
zschira: Yeah this has always felt like probably an unnecessary/error prone bit of state to maintain

src/pudl_archiver/orchestrator.py
```diff
@@ -210,62 +183,45 @@ async def run(self) -> RunSummary | Unchanged:
...
+        if not run_summary.success:
+            logger.error("Archive validation failed. Not publishing new archive.")
+            await self.depositor.delete_deposition(self.new_deposition)
+            return run_summary
+
+        await self._publish()
+
+        return run_summary
+
+    async def _upload_resources(
```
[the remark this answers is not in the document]
jdangerx: Ha! Yes, I thought about `_download_then_upload_resources` which seemed ridiculous at the time but now seems very reasonable.

src/pudl_archiver/orchestrator.py
jdangerx: Well - the URLs are sort of important. If we update `datapackage.json` to use the "canonical" URL of `/records/<record_id>/files/<file_name>` vs. whatever the hell `self.links.download` gives us (`/api/records/<record_id>(/draft)?/files/<file_name>(/content)?`), then we *do* want that to register as a change in the datapackage. But if we see the format stay the same & the record ID increment, then we don't want that to register as a change.

jdangerx: Overriding `__eq__` seemed like a too-general place to do this weird and specific logic to me - there are lots of ways to define equality of `Resource`s and I wouldn't presume to promote this one to default.

Issue #204, opened 2023-11-24
Recreate sandbox archives for all Zenodo datasets


e-belfer:
The recent Zenodo migration wiped all of our datasets in the sandbox. We should update each one to be able to test updates of our datasets. Some of these have already been updated through other PRs in the last few weeks, so the task here is to either confirm the creation of a new sandbox archive and update the DOI, or to create one. In the process we will also learn about whether our archivers are all working as expected.

- Fixes `mshamines.py` to account for migration of 107A dataset into main dataset page, removing special additional download page from archiver.
- Fix handling of `year` in EIA 860 archiver to address issue with `match` object being passed as `year`

Datasets to archive:
- [x] epacems
- [x] eia923
- [x] eia860
- [x] epacamd_eia
- [x] censusdp1tract
- [x] eiawater
- [x] eia861
- [x] mshamines
- [x] eia860m
- [x] eia176
- [x] eia_bulk_elec
- [x] ferc1
- [x] ferc2
- [x] ferc6
- [x] ferc60
- [x] ferc714

Out of scope:
- `phmsagas` will be addressed in #199 
- `ferceqr`

PR #206, opened 2023-11-27
Update sandbox DOIs

e-belfer:
See #204 for detailed discussion of PR scope. Make new sandbox archives for all datasets except `phmsagas` and `ferceqr`, fix issues with archivers.


jdangerx (approved):
Hooray! Thanks for fixing these!

PR #209, opened 2023-11-29
Allow integers and lists for partition pydantic validation, fix PHMSA bug

e-belfer:
With PHMSA, we use both integers (former) and lists of integers (current) as our partitions. This breaks the existing checks on `PartitionDiff`. This tiny PR allows more types for partitions.

Also fixes incorrect year range on the PHMSA archiver - `present` files should point to last year's data, while `_2003` should contain data up until 2003, e.g.


cmgosnell (approved):
i think this is actually much more straightforward to see what's going on now

PR #221, opened 2023-12-08
Push some state to `orchestrator.run()`

jdangerx:
We talked a big game about wanting to stop using `self.deposition`, `self.new_deposition` everywhere.

It wasn't such a bad change in the end!

In the future I think we can try to split up the concept of a "Zenodo `Deposition`" (the API response object) from a "PUDL Dataset Version" (some sort of concept that *we* have). That could look something like:

```python
class DatasetVersion:
    rec_id: int
    ...

class PublishedDataset:
    def from_doi(concept_doi) -> PublishedDataset:
       ...

    def get_draft_version(self) -> DraftDataset:
        # call Zenodo API here, maybe cry a little bit?

class DraftDataset(Dataset):
    def sync_with_resources(self, resources) -> DraftDataset:
        # call Zenodo API here, or we call the New Zenodo API here, whatever

# maybe some other states too - could see "completed" vs. "in-progress" drafts being a useful distinction

# in orchestrator.run...
original = PublishedDataset.from_doi(doi)
draft = original.get_draft_version()
draft.sync_with_resources()
```

That would make the `ZenodoDepositor` much more of a "thin shell around literal API calls" and the "convert operations that *we* care about into operations that *Zenodo* cares about" messiness would live in the various subclasses of `Dataset`.

I think we almost got to this level of abstraction with the last big refactor we did, but stopped short - so we never separated "the set of operations we care about doing to a dataset version" from "the set of operations that Zenodo knows how to do" cleanly.

I *might* try to get that restructuring in ahead of time, before the next Zenodo API change happens. Maybe we can even experiment with archiving to multiple RDMs...

### DIFF
--- a/src/pudl_archiver/orchestrator.py
+++ b/src/pudl_archiver/orchestrator.py
@@ -91,7 +91,7 @@ class DepositionOrchestrator:
         session: aiohttp.ClientSession,
         upload_key: str,
         publish_key: str,
-        deposition_settings: Path,
+        dataset_settings_path: Path,
         create_new: bool = False,
         dry_run: bool = True,
         sandbox: bool = True,
@@ -106,7 +106,7 @@ class DepositionOrchestrator:
             session: Async http client session manager.
             upload_key: Zenodo API upload key.
             publish_key: Zenodo API publish key.
-            deposition_settings: where the various production/sandbox concept
+            dataset_settings_path: where the various production/sandbox concept
                 DOIs are stored.
             create_new: whether or not we are initializing a new Zenodo Concept DOI.
             dry_run: Whether or not we upload files to Zenodo in this run.
@@ -134,8 +134,8 @@ class DepositionOrchestrator:
         self.dry_run = dry_run
 
         self.create_new = create_new
-        self.dataset_settings_path = deposition_settings
-        with Path.open(deposition_settings) as f:
+        self.dataset_settings_path = dataset_settings_path
+        with Path.open(dataset_settings_path) as f:
             self.dataset_settings = {
                 name: DatasetSettings(**dois)
                 for name, dois in yaml.safe_load(f).items()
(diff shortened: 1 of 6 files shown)


src/pudl_archiver/orchestrator.py
```diff
@@ -127,46 +131,35 @@ def __init__(
...
-        # TODO (daz): don't hold references to the depositions at the instance level.
-        self.deposition: Deposition | None = None
-        self.new_deposition: Deposition | None = None
-        self.deposition_files: dict[str, DepositionFile] = {}
-
         self.dry_run = dry_run
 
         self.create_new = create_new
-        self.deposition_settings = deposition_settings
+        self.dataset_settings_path = deposition_settings
```
zschira, line 137: Should we just change the `deposition_settings` parameter to `dataset_settings_path`?

src/pudl_archiver/orchestrator.py
```diff
@@ -179,38 +172,52 @@ async def run(self) -> RunSummary:
         Returns:
             RunSummary object.
         """
-        await self._initialize()
+        self.changes = []
+        if self.create_new:
+            original = await self._create_new_deposition()
```
zschira, line 177: `original` and `draft` are much better names

zschira (approved):
One small name suggestion, but looks great!

I also really love the idea of separating the zenodo deposition from a PUDL dataset version. Being able to easily accommodate different backends definitely seems more and more appealing by the day, but I also could imagine using those structures elsewhere. Having one coherent set of objects that wrap an archive no matter where it came for or lives, and can be used by the datastore, and fed into PUDL seems super useful.

src/pudl_archiver/orchestrator.py
jdangerx: Yes! Good thought!

PR #222, opened 2023-12-08
cems partition tweaks

cmgosnell:
this is attached to https://github.com/catalyst-cooperative/pudl/pull/3139

I *triiiiiedd* to make a new archive version which both updated the partition from `year` & `quarter` -> `year_quarter` as well as update the file names to be more in line with the partition update.

But renaming the file names means a much more difficult archive update because it breaks the connection between the old files and the new ones. So I reverted that and just updated the datapackage.json with @jdangerx 's postman help. I left the desired filenames as commented out lines in the pr.

I also added small i think updates to the readme to add some additional pointers for cats who haven't interacted w/ the archiver in some time.


dataset_doi.yaml
```diff
@@ -26,7 +26,7 @@ epacamd_eia:
   production_doi: 10.5281/zenodo.6633769
   sandbox_doi: 10.5072/zenodo.3136
 epacems:
-  production_doi: 10.5281/zenodo.4127054
+  production_doi: 10.5281/zenodo.10233185
```
cmgosnell, line 29: this was an old-ish requried change. ella had to make a whole new cems archive on the transition from the year/state annual files to the quarterly files but probably forgot to update this guy.

src/pudl_archiver/archivers/epacems.py
```diff
@@ -73,7 +73,9 @@ async def get_year_quarter_resource(
 
         # Create zipfile to store year/quarter combinations of files
         filename = f"epacems-{year}-{quarter}.csv"
+        # filename = f"epacems-{year}q{quarter}.csv"
         archive_path = self.download_directory / f"epacems-{year}-{quarter}.zip"
+        # archive_path = self.download_directory / f"epacems-{year}q{quarter}.zip"
```
cmgosnell, line 78: I left these commented out file name and path updates because in an ideal world we would convert all of the file names to be more consistent with the partition. this is not a required change because the datastore knows how to grab a file based on the partition regardless of if the partition and the filename match because of the info stored in the datapackage.json. Nonetheless this would still be a good change to make if we could easily fully update the file names.

README.md
```diff
@@ -46,10 +46,16 @@ conda activate pudl-cataloger
 ```
 
 ## Setting up environment
+
 API tokens are required to interact with Zenodo. There is one set of tokens for accessing
 the sandbox server, and one for the production server. The archiver tool expects these tokens
 to be set in the following environment variables: `ZENODO_TOKEN_PUBLISH` and `ZENODO_TOKEN_UPLOAD`
 or `ZENODO_SANDBOX_TOKEN_PUBLISH` and `ZENODO_SANDBOX_TOKEN_UPLOAD` for the sandbox server.
+Catalyst uses a set of institutional tokens - you can contact a mainter for tokens.
```
jdangerx, line 54:
```suggestion
Catalyst uses a set of institutional tokens - you can contact a maintainer for tokens.
```

src/pudl_archiver/archivers/epacems.py
jdangerx: I think these comments as-is would be kind of confusing in the future, since they don't have context as to why they're commented out.

When we make the next quarterly updates to the archive, we'll need to re-download and re-upload all the files anyways, right? Could we rename everything then? If we can, then I think we should probably go ahead and change the code now.

jdangerx (changes requested):
Nits to pick about your comments but otherwise good! And my condolences for having to deal with long CEMS downloads / Zenodo uploads.

README.md
cmgosnell: ha! ty

src/pudl_archiver/archivers/epacems.py
cmgosnell: yeah this is a very good point! I guess I am partial to adding a comment in there with a `# TODO:` and creating an issue once we talk about how to make re-making this archive a less flaky thing to do.

jdangerx (approved):

PR #228, opened 2023-12-13
Point automated archiver at production Zenodo.

jdangerx:
Closes catalyst-cooperative/pudl#3068.

Now that we have automatically running archivers that send stuff to Slack when they succeed/fail and point us towards places we can actually approve them, we should point these at production!


cmgosnell (approved):
🍏

PR #231, opened 2023-12-15
Don't do a weird song-and-dance to guarantee a fresh draft.

### LINKED ISSUE #230: CEMS archiver fails frequently
Running the CEMS archiver locally seems to fail frequently. It usually runs for several hours, then fails at some point on the Zenodo end.

Things that we've identified as issues:
- [ ] Zenodo upload + deletion flakiness #231 
- [ ] Zenodo 100 file limit seems to be enforced #237
- [ ] CEMS download is slow/flaky locally

We're working on the first two bullets - if we can get CEMS running on GHA, we can see if the download speeds are still an issue.

jdangerx:
As part of #230 I found that we were running into flakiness when deleting drafts and re-creating them to guarantee a fresh starting state for our archiver - i.e., one where the new draft matches the existing deposition exactly.

But, I realized that we don't need to guarantee a fresh starting state and can ditch a lot of that logic - and the less logic we have that tries to deal with Zenodo state, the less that instability in that behavior will affect us 🥲 

This doesn't fix #230 completely, because we seem to also sometimes fail to download from CEMS, but should improve our Zenodo interaction somewhat.

### DIFF
--- a/dataset_doi.yaml
+++ b/dataset_doi.yaml
@@ -6,7 +6,7 @@ eia176:
   sandbox_doi: 10.5072/zenodo.3158
 eia860:
   production_doi: 10.5281/zenodo.4127026
-  sandbox_doi: 10.5072/zenodo.3299
+  sandbox_doi: 10.5072/zenodo.11168
 eia860m:
   production_doi: 10.5281/zenodo.4281336
   sandbox_doi: 10.5072/zenodo.3172
@@ -27,7 +27,7 @@ epacamd_eia:
   sandbox_doi: 10.5072/zenodo.3136
 epacems:
   production_doi: 10.5281/zenodo.10233185
-  sandbox_doi: 10.5072/zenodo.2236
+  sandbox_doi: 10.5072/zenodo.9811
 ferc1:
   production_doi: 10.5281/zenodo.4127043
   sandbox_doi: 10.5072/zenodo.3267
--- a/src/pudl_archiver/archivers/epacems.py
+++ b/src/pudl_archiver/archivers/epacems.py
@@ -44,6 +44,7 @@ class EpaCemsArchiver(AbstractDatasetArchiver):
                 and (self.valid_year(file["metadata"]["year"]))
             ]
             logger.info(f"Downloading {len(quarterly_emissions_files)} total files.")
+            logger.debug(f"File info: {quarterly_emissions_files}")
             for i, cems_file in enumerate(quarterly_emissions_files):
                 yield self.get_year_quarter_resource(
                     file=cems_file, request_count=i + 1
@@ -69,7 +70,7 @@ class EpaCemsArchiver(AbstractDatasetArchiver):
         quarter = file["metadata"]["quarter"]
 
         # Useful to debug at download time-outs.
-        logger.debug(f"Downloading {year} Q{quarter} EPACEMS data.")
+        logger.info(f"Downloading {year} Q{quarter} EPACEMS data from {url}.")
 
         # Create zipfile to store year/quarter combinations of files
         filename = f"epacems-{year}-{quarter}.csv"
--- a/src/pudl_archiver/depositors/zenodo.py
+++ b/src/pudl_archiver/depositors/zenodo.py
@@ -224,91 +224,42 @@ class ZenodoDepositor:
             A new Deposition that is a snapshot of the old one you passed in,
             with a new major version number.
         """
-        # if this is a draft, either this is a potentially dirty draft (in
-        # which case, delete it) - or, this is the first deposition of a
-        # concept and we should just return it unharmed.
-        if not deposition.submitted:
-            if clobber:
-                await self.delete_deposition(deposition)
-                deposition = await self.get_deposition(str(deposition.conceptdoi))
-            else:
-                return deposition
-
-        url = f"{self.api_root}/deposit/depositions/{deposition.id_}/actions/newversion"
-
-        # create a new unpublished deposition version
-        try:
-            response = await self.request(
+        # just get the new draft from new API.
+        new_draft_record = Record(
+            **await self.request(
                 "POST",
-                url,
-                log_label="Creating new version",
-                headers=self.auth_write,
-                retry_count=2,
-            )
-        # Except if abandoned in progress draft
-        except ZenodoClientError as excinfo:
-            if (
-                clobber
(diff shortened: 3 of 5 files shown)


dataset_doi.yaml
```diff
@@ -6,7 +6,7 @@ eia176:
   sandbox_doi: 10.5072/zenodo.3158
 eia860:
   production_doi: 10.5281/zenodo.4127026
-  sandbox_doi: 10.5072/zenodo.3299
```
jdangerx, line 9: These changed because I needed to re-run test datasets with `--initialize` - before the draft discard change in this PR, updating an existing sandbox dataset would get Zenodo into a weird state. Anytime after you updated an existing dataset, a request for `/records/<concept_record_id>` would return `410: Resource deleted`.

src/pudl_archiver/archivers/epacems.py
```diff
@@ -69,7 +70,7 @@ async def get_year_quarter_resource(
         quarter = file["metadata"]["quarter"]
 
         # Useful to debug at download time-outs.
-        logger.debug(f"Downloading {year} Q{quarter} EPACEMS data.")
+        logger.info(f"Downloading {year} Q{quarter} EPACEMS data from {url}.")
```
jdangerx, line 73: This was handy for manually downloading files later to play with.

src/pudl_archiver/orchestrator.py
```diff
@@ -188,14 +188,15 @@ async def run(self) -> RunSummary:
...
-            draft, resources
+            original=original, draft=draft, resources=resources
         )
         run_summary = self._summarize_run(
             old_datapackage, new_datapackage, resources, draft.links.html
         )
 
-        if not self.changes:
-            # must run *after* potentially updating datapackage.json
+        if len(run_summary.file_changes) == 0 and not self._datapackage_worth_changing(
```
jdangerx, line 197: Still don't love this `_datapackage_worth_changing` thing.

src/pudl_archiver/depositors/zenodo.py
```diff
@@ -224,91 +224,42 @@ async def get_new_version(
...
-            if key not in ["doi", "prereserve_doi", "publication_date"]:
-                metadata[key] = val
-
-        previous = semantic_version.Version(source_metadata["version"])
-        version_info = previous.next_major()
-
-        metadata["version"] = str(version_info)
+        )
+        # then, get the deposition based on that ID.
+        new_draft_deposition = await self.get_deposition_by_id(new_draft_record.id_)
```
cmgosnell, line 237: clarifying question.... i get confused about all the different kinds of ids flying around. If the `new_draft_record.id` going to try to make a draft archive based on the same underlying ... idk what it is called ... "core" id (or deposition id?)... the one that links CEMS version 1 to CEMS version 2?

cmgosnell (commented):
This overall looks like a great simplification! My one high level question is.... if we make bad drafts will there just be a bunch of stale drafts lying around? And if so should we add some clean-up logic that is independent from actually creating a fresh archive so we can keep this simplification but also delete old guys?

jdangerx:
So Zenodo guarantees that there is only ever one draft lying around for any Concept ID (which is what ties multiple versions of the same dataset together). If we "get new version" and there's already a draft, it will just give us the existing draft back.

If we make a bunch of bad drafts we'll be stuck with the latest bad draft lying around. But then when we re-run the archiver, it will try to upload new good files and delete the files that are in the draft but not in the source material.

src/pudl_archiver/depositors/zenodo.py
jdangerx: Ah, so what's happening here is:

1. We're using the New API for creating a new draft, since it's less flaky - which returns a slightly different data shape than the Old API.
2. But, everywhere else, we're using the Old API. So we need to use the ID that we just got back from the New API endpoint, and feed it through the Old API to get the right shape back.

cmgosnell: wait the new and the old api are both running and interacting with the same draft records?!? do we expect the new api is actually stable enough to rely on? (good news that it is less flaky though!)

cmgosnell (approved):
i'm honestly not sure if i fully understand the implications of this change but i definitely like it in theory. I'm a little queasy about interacting with both zenodo APIs but if that's what we gotta do then that's what we gotta do.

Structurally if we are relying on the end step of the new draft creation to check for bad files and then delete them i think we will probably end up with failures at the end of the process bc this seems to be a flaky step. But then the whole archive won't be deleted presumably so we could hand finish it if need be...

i think it would be good to get @zschira 's review in here as well bc i am not nearly as familiar w/ the quirks of the archiving

src/pudl_archiver/depositors/zenodo.py
jdangerx: So, there are two ways to create a new draft: the old API and the new API.

The main relevant difference here is that the old API will *fail* if there's already a draft lying around, and the new API will just return that draft.

We *could* use the old API and catch the exception, except... then we have to go find the existing draft through some other means. We used to do this by GETting the "list depositions with an optional query string" endpoint in the legacy API. We would ask it for "give us the latest deposition that belongs to this concept DOI" and it would give us the draft back.

But, now that endpoint doesn't work. So the ONLY WAY we can get the latest draft in that case is actually by using the new API.

So our options are:

1. Use the new API and just handle things breaking when the behavior changes
2. Use the old API, but then rely on the new API anyways when the old API fails...

So we're sort of stuck using the new API either way.

src/pudl_archiver/orchestrator.py
zschira: We call `_summarize_run` before this, and it takes both datapackages, so maybe we could `run_summary` contain a field that tells us if files changed, or the datapackage changed?

src/pudl_archiver/depositors/zenodo.py
zschira: Yeah I don't love mixing API's, but overall I think it seems simpler than doing the error handling ourselves, so overall I'm ok with this if it has been working consistently.

zschira (approved):
Looks good to me (as good as zenodo will allow us to make it anyway)

src/pudl_archiver/orchestrator.py
jdangerx: Yeah, that could work! I think this is probably OK for now and then we can fix it later if we revisit the code 😬

PR #232, opened 2023-12-15
Validate cems archive

aesharpe:
This validation test validates that `year_month` and `year_quarter` partitions are 1) not duplicated and 2) continuous from first to last partition. This will run on partitions that are formatted as lists or single items. It will work on CEMS and EIA 860M for now, but could be expanded to other datasets or to include `year` in the future.

Out of scope:
- test works on datasets with multiple partitions to be tested (e.g. `year_month` and `year_quarter`)
- test works on datasets with multiple partitions where only one should be tested (e.g. `form` and `year_quarter`) but values of partitions may be duplicated (e.g. each `form` has a resource with Q1 1995 data).


src/pudl_archiver/archivers/epacems.py
```diff
@@ -86,5 +89,56 @@ async def get_year_quarter_resource(
             f"File no. {request_count}: Downloaded {year} Q{quarter} EPA CEMS hourly emissions data."
         )
         return ResourceInfo(
-            local_path=archive_path, partitions={"year_quarter": f"{year}q{quarter}"}
+            local_path=archive_path,
+            partitions={"year": year, "quarter": quarter},
+            layout=ZipLayout(filepaths=[filename]),
```
e-belfer, line 94: Blocking: This needs to be `file_paths`, otherwise it produces the following error when running an archive:

```
file_paths
  Field required [type=missing, input_value={'filepaths': ['epacems-1996-2.csv']}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.5/v/missing
```

When amended to `file_paths` we get: `No validations defined for files of type: .csv - epacems-1996-4.csv`
Perhaps this is the wrong filename that is being passed, and the method wants the zipfile name rather than the CSV? Check with @zschira that this is being implemented as intended.

e-belfer (dismissed):
Major blocking change is that the addition of `layout` here in the `ResourceInfo`, which I think actually comes from @zschira 's zipfile validation PR. However, it is currently configured incorrectly and fails all new archive creations. Left one comment about how to fix the failure, but would be worth checking with him about how to get it to actually work as expected.

Once fixed, this is still failing on the following `KeyError`:
```
year_quarters = [x.parts["year_quarter"] for x in new_datapackage.resources]
KeyError: 'year_quarter'
```
Most likely, this is because we have `year` and `quarter` as separate partitions in our CEMS archive, but `year_quarter` in the datapackage. I think this is a weird side effect of the workaround in #222 that @cmgosnell set up.

A few minor changes could cut a lot of lines from your unit test setup. Otherwise, the unit test looks as expected to me. 

I left a few other non-blocking questions. I think having testing for consecutive partitions and no out of bounds dates are probably pretty fairly generalizable functions we could imagine wanting to do with other archives (e.g., no month 13), so writing each of these as a function in `pudl_archivers.archivers.validate` with parameters would be preferable. Then we could just import `validate` and call `validate.partitions_in_bounds()` or `validate.consecutive_partitions()`, for example.

e-belfer:
I'm curious about @cmgosnell's thoughts, I'd prefer to just set the `check_missing_files` flag false for one CEMS update and clean up the partition to be `year_quarter` if that's what we want it to be, rather than adding a workaround in the datapackage as we do in #222. That's slightly out of scope of this PR but also clearly seems to be breaking your workflow here.

aesharpe:
Thanks @e-belfer! Wondering if it's worth continuing down the path of fixing these specific CEMS tests or whether I should start to generalize some of these functions for other datasets. @cmgosnell happy to have your input as well. 

I asked about this early on in the process and the verdict was to stick with specific-to-CEMS tests, but worth checking in again before I move forward.

e-belfer:
Ah sorry I might have missed this convo last week, and I'm not quickly finding it on Github so I might've missed the pros/cons others raised. I think it's a nice to have but non-blocking for me, as I can imagine us wanting to run basically the same check on many other datasets (e.g. EIA 860M, any annual data).

aesharpe:
Agree! It was mostly off-the-record convos with @zschira @jdangerx and @cmgosnell.

I think making the consecutive paritions generalizable will be a little trickier given the different types of partitions available (year, month, quarter)

src/pudl_archiver/archivers/epacems.py
```diff
@@ -86,5 +89,56 @@ async def get_year_quarter_resource(
...
+                f"The following years do not have four quarters: \
+                    {q_groups[q_groups['quarter']!=4].index.tolist()}. """
+            )
+        # Test that the newest year has consecutive quarters
+        new_quarters = yq_df[yq_df["year"] == newest_year].quarter.unique().tolist()
+        if len(set(range(1, len(new_quarters) + 1)) - set(new_quarters)) != 0:
+            success = False
+            description = (
+                description + "There are missing quarters in the new year of data. "
+            )
```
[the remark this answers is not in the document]
aesharpe: Do you mean like a logging output or a code comment?

src/pudl_archiver/archivers/epacems.py
aesharpe: @zschira I think this line was your addition, any thoughts here?

src/pudl_archiver/archivers/epacems.py
```diff
@@ -86,5 +89,56 @@ async def get_year_quarter_resource(
...
+        new_datapackage: DataPackage,
+        resources: dict[str, ResourceInfo],
+    ) -> list[validate.DatasetSpecificValidation]:
+        """Look for consecutive year quarter partitions in CEMS archive."""
+        year_quarters = [x.parts["year_quarter"] for x in new_datapackage.resources]
+        yq_split = [x.split("q") for x in year_quarters]
+        yq_df = pd.DataFrame(yq_split, columns=["year", "quarter"]).apply(
+            pd.to_numeric, errors="coerce"
+        )
+        newest_year = yq_df.year.max()
```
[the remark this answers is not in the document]
aesharpe: The idea here is to allow for incomplete (but non non-consecutive) data in the newest year. This is because we want to be able to ingest the data one quarter at a time, as it comes out. For all years prior to the newest year, we expect all four quarters to be available. 

This would only cause an issue if we pulled a subset of the data and there happened to be quarters missing from the last pulled year when there should be complete data. This function would not flag those missing quarters because it enables incomplete "last" years.

Given that we haven't really used the `only_years` flag, this doesn't seem like a huge issue. However, I see how it could potentially cause trouble. I agree that puling the current date and performing some math to get the appropriate lag time is probably more likely to cause problems than leaving it as is.

src/pudl_archiver/archivers/epacems.py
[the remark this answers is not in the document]
e-belfer: For other tests `description` is a static description of what the test does (e.g. "Check to ensure that CEMS data is consecutive") and the `notes` field offers descriptions of the specific test info (e.g., these quarters are missing, this file fails the conditions). I think keeping to this format is more consistent and will give more info to the person reviewing the summary .json about what the test is, since currently the description is blank if it passes.

src/pudl_archiver/archivers/epacems.py
zschira: I just checked, and right now we don't actually have a specific validation for CSV's defined, so it's still checking that the file exists, and that it's not empty, but that's all it does. We could add a function that opens up the CSV and reads it in to see that it's truly a valid CSV, but the default checks might be sufficient.

tests/unit/archive_validate_test.py
```diff
@@ -299,3 +306,160 @@ def test_run_summary_success(specs, expected_success):
...
+                    "hash": "e9ece6bb8190e14d44834bf0a068ac5d",
+                },
+            ],
+            "created": "2023-11-30 20:51:43.255388",
+            "version": "1",
+        }
+    )
+)
+fake_new_datapackage_month_success = copy.deepcopy(fake_new_datapackage_quarter_success)
+# Make a successful monthly datapackage based on the quarterly one
```
e-belfer, line 406: We can add an `id` field to the `pytest.mark.parameterize` with a list of test IDs here rather than comments, so we know what each test is. I just did this in https://github.com/catalyst-cooperative/pudl-archiver/pull/236 if you need an example.

src/pudl_archiver/archivers/validate.py
```diff
@@ -285,3 +286,58 @@ def _validate_xml(buffer: BytesIO) -> bool:
...
 
     return True
+
+
+def validate_data_continuity(new_datapackage: DataPackage) -> DatasetSpecificValidation:
+    """Check that the archived data are continuous and complete."""
+    description = "Test that data are continuous and complete"
+    success = True
+    note = []
+    part_range_dict = {"year_quarter": [1, 4, 7, 10], "year_month": list(range(1, 13))}
```
e-belfer, line 296: Ideally we'd code this as a parameter outside of this function.

src/pudl_archiver/archivers/validate.py
```diff
@@ -285,3 +286,58 @@ def _validate_xml(buffer: BytesIO) -> bool:
...
+    # Make a dictionary of part type and parts (ex: {"quarter_year": [1995q1, 1995q2]}) from the new
+    # archive datapackage to make it easier to loop through.
+    parts_dict_list = [resource.parts for resource in new_datapackage.resources]
+    # Identify the newest year for the whole archive
+    newest_year_part = max(
```
e-belfer, line 301: This assumes that all data sources have partitions with a date in them, which isn't always the case. I think we need to either config each dataset to specify what partition is getting checked, or set a list of partitions that are configured to work with this test (e.g. `year_month`, `year_quarter`) and only test those.

tests/unit/archive_validate_test.py
```diff
@@ -299,3 +306,160 @@ def test_run_summary_success(specs, expected_success):
...
+        (fake_new_datapackage_quarter_fail2, False),
+        (fake_new_datapackage_month_fail2, False),
+        (fake_new_datapackage_month_fail3, True),
+    ],
+)
+def test_validate_data_continuity(new_datapackage, success):
+    """Test the dataset archiving valiation for epacems."""
+    validation = validate.validate_data_continuity(new_datapackage)
+    logger.info(validation)
+    if validation["success"] != success:
```
e-belfer, line 462: can also just `assert validation['success'] == success`

aesharpe: I feel like I always shy away from `assert` statements because I was once told they were unstable compared to `raise AssertionError`. But happy either way.

src/pudl_archiver/archivers/classes.py
```diff
@@ -364,6 +367,73 @@ def _check_dataset_size(
...
+
+        # Only perform this test if the part label is year quarter or year month
+        # Note that this currently only works if there is one set of partitions,
+        # and will fail if year_month and form are used to partition a dataset, e.g.
+        if any(partition in VALID_PARTITION_RANGES for partition in partition_names):
+            partition_to_test = [
+                partition
+                for partition in partition_names
+                if partition in VALID_PARTITION_RANGES
+            ][0]
```
cmgosnell, line 396: this isn't the case rn but if an archive has multiple partitions in `VALID_PARTITION_RANGES` we'll only test the first one we find

src/pudl_archiver/archivers/classes.py
```diff
@@ -364,6 +367,73 @@ def _check_dataset_size(
...
+            interval = (
+                "QS"
+                if partition_to_test == "year_quarter"
+                else "MS"
+                if partition_to_test == "year_month"
+                else None
+            )
+
+            if interval:
```
cmgosnell, line 405: doesn't this if have the same functionality as the `if any(partition in VALID_PARTITION_RANGES for partition in partition_names):`? could we remove one of them? could `VALID_PARTITION_RANGES` be a dictionary of `part: freq`? so you don't have to to this nested if-ing?

src/pudl_archiver/archivers/classes.py
```diff
@@ -364,6 +367,73 @@ def _check_dataset_size(
...
+                        f"Partition contains duplicate time periods of data: f{observed_date_range.duplicated()}"
+                    ]
+
+                else:
+                    diff = expected_date_range.difference(observed_date_range)
+                    if not diff.empty:
+                        success = False
+                        note = [
+                            f"Downloaded partitions are not consecutive. Missing the following {partition_to_test} partitions: {diff.to_numpy()}"
+                        ]
```
cmgosnell, line 423: i think this would be more readable if this was a combined `elif`

```suggestion
                elif not diff.empty:
                    success = False
                    note = [
                            f"Downloaded partitions are not consecutive. Missing the following {partition_to_test} partitions: {diff.to_numpy()}"
                        ]
                else:
                    success = True
                    note = None
```

src/pudl_archiver/archivers/classes.py
```diff
@@ -364,6 +367,73 @@ def _check_dataset_size(
...
+        else:
+            note = [
+                "The dataset partitions are not configured for this test, and the test was not run."
+            ]
+
+        return validate.DatasetSpecificValidation(
```
cmgosnell, line 429: is this dataset specific now? it seems general but it just doesn't run unless it has the right partition

tests/unit/archive_base_test.py
```diff
@@ -540,3 +543,158 @@ def test_year_filter():
...
+fake_new_datapackage_month_success = copy.deepcopy(fake_new_datapackage_quarter_success)
+for resource in fake_new_datapackage_month_success.resources:
+    resource_months_min = pd.to_datetime(min(resource.parts["year_quarter"]))
+    resource_months_max = pd.to_datetime(
+        max(resource.parts["year_quarter"])
+    ) + relativedelta(months=3)
+    dd_idx = pd.date_range(start=resource_months_min, end=resource_months_max, freq="M")
+    resource.parts = {
+        "year_month": [pd.to_datetime(x).strftime("%Y-%m") for x in dd_idx]
+    }
```
cmgosnell, line 649: why is this chunk necessary? because its converting the quarterly datapackage to a monthly datapackage?

cmgosnell (changes requested):
some readability suggestions and one question but this looks _relatively_ straightforward!

the tests were a little inscrutable. I think it might be good to simplify them a little - like only add more examples when that example is testing a new edge case instead of just more examples.

lots of these are nits and none of these concerns are really blocking but there are enough lil guys in here that i am going to request changes.

src/pudl_archiver/archivers/classes.py
aesharpe: I originally had it designed with this in mind but I talked to @e-belfer and she said that this would probably never happen.

tests/unit/archive_base_test.py
aesharpe: Yes, this code converts the quarterly test datapackage (the only one fully written out) to a monthly datapackage.

src/pudl_archiver/archivers/classes.py
e-belfer: Agreed and done

src/pudl_archiver/archivers/classes.py
e-belfer: I can't imagine us wanting to partition both by `year_quarter` and `year_month`, and this raises a whole other host of questions - are we enforcing consecutiveness within all the partition values of one key for one value of the other key (e.g., all the `year_months` by `form`, if `form` were capable of being consecutive?). I agree the silent failure is bad and have explicitly added a test failure to flag this.

src/pudl_archiver/archivers/classes.py
e-belfer: Agreed and added!

src/pudl_archiver/archivers/classes.py
e-belfer: The names are slightly confusing. We have `DatasetSpecificValidation` which is the "ValidationTestResult specific to an entire dataset." Then we also have `dataset_validate_archive` which is the method for adding dataset-specific validations - validations for any one particular dataset. I think we could probably rename `DatasetSpecificValidation` and it's twin `FileSpecificValidation` to `GenericDatasetValidation` and `GenericFileValidation`?

cmgosnell (approved):
this looks great! these are great simplifications and help readability imo

oh im pressing approve but you should probs remove the `fake_new_datapackage_quarter_success`, `fake_new_datapackage_month_success ` etc etc.

Issue #217, opened 2023-12-04
Identify and implement 1-2 additional generic validations for all datasets to help us catch remaining gaps (e.g. size of all files hasn’t changed by more than 10%).


jdangerx:
Current validations run on all datasets include:
- No files missing from prior deposition
- Files valid based on type (.xlsx, .zip, .xml, .xbrl, .xsd all checked)
- Files not empty
- Zipfile layout valid

To improve this process, we could add a few additional validation tests based on the kinds of manual checks that we typically do when approving draft archives:
- **File size**: If a file has changed in size by >|X|% from its last archive, this is probably a problem. Or, if a file is substantially different in size than the average file size in the deposition, this might be worth noting. (Essentially, tiny files are often empty or corrupted). This check would be tricky for depositions where there are different types of forms (e.g. PHMSA, MSHA), but we may just want to turn it off for these archives or tune it differently.
- **Deposition size**: If a deposition has reduced in size overall from the last deposition, this is likely an issue worth investigating.
- **Unexpected partitions**: We expect a new year or month of data, but if other partitions (new form, earlier year of missing data) appear we should flag this for review.

This issue is also discussed in https://github.com/catalyst-cooperative/pudl-archiver/issues/70, with some more extensive suggestions there.

Issue #70, opened 2023-02-23
Specify archiver success & failure conditions


zaneselvans:
Our goal is to have the archivers running on an automated schedule in the background, taking snapshots of the original data sources which can be accessed programmatically. This will minimize the overhead associated with keeping our raw inputs up to date, but we still need the system to alert us when something goes wrong so we can fix it.

* Success or failure should be defined at the level of each individual dataset so as to effectively direct our attention to where the problem is.
* Each dataset's archiver should produce a report describing the outcome of the archiving run. See #60 
* What constitutes a successful archiving run will vary by dataset. We need a way to specify our expectations, and if anything outside of those expectations happens we should get a failure. If we start with stringent criteria and find we're getting too many false positives ("failures" that are actually okay), we can loosen the criteria based on the actual outcomes we see.
* In general we expect the set of data partitions to either remain constant or grow over time, but we have fairly specific expectations about what new data partitions should look like.
* Typically new partitions are just additional timesteps, e.g.:
  * a new year of EIA-923 data
  * a new month of EIA-860m data
  * a new month of EPA CEMS data
* If any expected data partition is not found, that should result in failure.
* If a new data partition of an unexpected form is found, that should result in failure, since it means our expectations about what the data should look like are no longer correct. We should be required to explicitly update our expectations about the data.
* We should at least report some measure of the scale of changes detected between versions, and beyond a certain threshold we might want to cause a failure. E.g.:
  * If all the expected data partitions are found, but some of them have decreased in size by 90% something is probably wrong and needs to be investigated.
  * If the file type has changed even though the file name has not changed (as happened when the EIA-176 started publishing Excel spreadsheets but calling them CSVs) that should also probably result in a failure that demands investigation.
  * If an unusually high proportion of the data partitions has changed, but they still have the right names, file types, and sizes, maybe it's not a failure, but should be investigated as it could indicate big revisions to the data (as often happens without any kind of notice or documentation).

zschira:
I think having success/failure conditions like this is a really good idea for making automated archives useful and hopefully catch errors early.

> In general we expect the set of data partitions to either remain constant or grow over time

I think we could probably fail or at least require human review any time we would delete a partition outright as that's almost always unexpected.

Another thing we should probably start considering is some procedures for handling failures. For example, we should plan some sort of human intervention mechanism if we deem an archive to actually be acceptable even if it does generate a failure.

PR #236, opened 2023-12-18
Add file and dataset size validation checks

e-belfer:
To respond to #70 and complete #217, this PR adds two generic dataset validations for all archives:

- dataset size changes: if the overall dataset size changes by more than |X| from the last archive, throw a validation error.
- file size changes: if any one file size's changes by more than |X| from the last archive, throw a validation error.

To do:
- [x] Figure out how to configure the threshold of change for each dataset
- [x] Pick a more reasonable threshold than 10% (which seems too low)


src/pudl_archiver/archivers/classes.py
```diff
@@ -58,8 +58,12 @@ class AbstractDatasetArchiver(ABC):
     directory_per_resource_chunk: bool = False
 
     # Configure which generic validation tests to run
-    check_missing_files: bool = True
-    check_empty_invalid_files: bool = True
+    fail_on_missing_files: bool = True
```
jdangerx, line 61: non-blocking: In the future, if we add more validation tests, we might want to refactor this so that the `fail_on_x` and the `required_for_run_success` flags aren't duplicated - but that's not really an issue at this juncture IMO.

tests/unit/archive_base_test.py
```diff
@@ -405,6 +405,407 @@ def test_check_missing_files(datapackage, baseline_resources, new_resources, suc
...
 
 
+@pytest.mark.parametrize(
+    "baseline_resources,new_resources,success",
+    [
+        (
+            # Test when file size changes by > threshold for any one file
+            [
+                Resource(
+                    name="resource0",
```
jdangerx, line 415: It'll make the test cases way easier to understand if we make this more compact with a factory function that only takes the varying parameters. I think it shouldn't be much more than this:

```
def _resource_w_size(name, size):
    return Resource(
        name=name,
        path=f"https://www.example.com/{name}",
        ...
        bytes=size
    )
```

src/pudl_archiver/archivers/classes.py
```diff
@@ -245,7 +249,86 @@ def _check_missing_files(
...
+            required_for_run_success=self.fail_on_missing_files,
+        )
+
+    def _check_file_size(
+        self,
+        baseline_datapackage: DataPackage | None,
+        new_datapackage: DataPackage,
+    ) -> validate.DatasetSpecificValidation:
+        """Check if any one file's size has changed by |>allowed_file_rel_diff|."""
+        baseline_resources = set()
```
jdangerx, line 261: These two start out as `set`s but are treated as `dict`s later on, which has some surprising behavior:

```
>>> ss = set()
>>> ss.update({1: "a"})
>>> ss
{1}
```

We should probably just initialize them as dicts here.

src/pudl_archiver/archivers/classes.py
```diff
@@ -245,7 +249,86 @@ def _check_missing_files(
...
+        too_changed_files = set()
+        notes = None
+
+        if baseline_datapackage is not None:
```
jdangerx, line 265: non-blocking: it might be clearer that `baseline_datapackage is None` implies that this will always pass if you return early from that case:

```
if baseline_datapackage is None:
    return validate.DatasetSpecificValidation(...)
```

vs. populating this value only if baseline datapackage is not None. Then you can also avoid having to initialize `baseline_resources` to an empty value before assigning it an actual value.

I think it's more lines of code, because of the various params getting passed into `DatasetSpecificValidation`, but it's a little bit less Actual Logic.

src/pudl_archiver/archivers/classes.py
```diff
@@ -245,7 +249,86 @@ def _check_missing_files(
...
+    def _check_dataset_size(
+        self,
+        baseline_datapackage: DataPackage | None,
+        new_datapackage: DataPackage,
+    ) -> validate.DatasetSpecificValidation:
+        """Check if a dataset's overall size has changed by more than |>allowed_dataset_rel_diff|."""
+        notes = None
+        baseline_size = None
+
+        if baseline_datapackage is not None:
```
jdangerx, line 308: non-blocking: Similarly, it might be nice to return early from the `baseline_datapackage is None` case here, trading a bit of verbosity for less logic/state to carry around.

Actually in this function we get more out of the early return, since we then guarantee that `baseline_size is not None` in the rest of the function body. And then that `if baseline_size is not None` check can go away.

src/pudl_archiver/archivers/classes.py
```diff
@@ -265,12 +348,21 @@ def validate_dataset(
             Bool indicating whether or not all tests passed.
         """
         validations: list[validate.ValidationTestResult] = []
+
+        # Run baseline set of validations for dataset using datapackage
         validations.append(
             self._check_missing_files(baseline_datapackage, new_datapackage)
         )
+        # TO DO: pass config for %age off!
```
jdangerx, line 356: This config is now passed through the class attr, right?

tests/unit/archive_base_test.py
```diff
@@ -405,6 +405,407 @@ def test_check_missing_files(datapackage, baseline_resources, new_resources, suc
     assert validation_result.success == success
 
 
+@pytest.mark.parametrize(
+    "baseline_resources,new_resources,success",
+    [
+        (
+            # Test when file size changes by > threshold for any one file
```
jdangerx, line 412: non-blocking: You can name these test cases with `pytest.param(value, other_value, id="some name")` which will make test output a little bit nicer.

tests/unit/archive_base_test.py
```diff
@@ -405,6 +405,407 @@ def test_check_missing_files(datapackage, baseline_resources, new_resources, suc
...
+                    bytes=10,
+                    hash="",
+                ),
+            ],
+            True,
+        ),
+    ],
+)
+def test_check_file_size(datapackage, baseline_resources, new_resources, success):
+    """Test the ``_check_missing_files`` validation test."""
```
jdangerx, line 600: I think you don't need this docstring at all, we should have disabled the linting requirement in test. But if you want to keep the docstring, this (and the below docstring) still refers to `_check_missing_files`.

jdangerx (changes requested):
Just one blocking change to make the tests more readable, plus a few non-blocking code structure stuff that you can take or leave. Ping me aggressively for re-review!

src/pudl_archiver/archivers/classes.py
e-belfer: Ah I did this because `ruff` was screaming at me but I'll update it, that seems undesirable.

src/pudl_archiver/archivers/classes.py
e-belfer: The `required_for_run_success` flags are taking the `fail_on_x` parameters as boolean inputs at this point, so I'm not 100% following what would be the proposed change here?

tests/unit/archive_base_test.py
e-belfer: Done, great suggestion

tests/unit/archive_base_test.py
e-belfer: I think it's helpful for user legibility, but good catch! I've just updated them

src/pudl_archiver/archivers/classes.py
e-belfer: I've made a change halfway towards your suggestion, cleaning things up without fully duplicating the return call.

src/pudl_archiver/archivers/classes.py
e-belfer: I've moved things around here to move duplicate logic without returning twice. Let me know what you think!

e-belfer:
I bumped up the tolerances - I suspect as we continue to make new archives we'll fine tune the expected common threshold and find some dataset specific outliers but this seems like a reasonable starting point.

src/pudl_archiver/archivers/classes.py
jdangerx: Right now we have to make sure to pass the `fail_on_x` parameters in to the `required_for_run_success` flags - which is easy enough, but still error-prone. We *should* be able to rejigger the class structure so that we can't accidentally break it, but that definitely seems like too big of a refactor for this PR.

tests/unit/archive_base_test.py
jdangerx, line 1: non-blocking:

It seems unlikely the the baseline datapackage would have filesizes of 0, and if we *do* run into that, we'll throw a `DivisionByZero` error instead of failing silently.

But, it's probably smart to write down what we *expect* to happen if there's a filesize of 0 floating around. Do we want to error? Do we want to ignore that file? In either case, it might be nice to write a test that asserts that behavior

jdangerx (approved):
Sweet, looks much better!

One non-blocking comment about adding a divide-by-zero case. Should have caught it sooner, my bad!

I think you could merge this without the test, but it's *probably* worth adding that before merge. And I definitely don't need to re-review it afterwards, so I'm smashing that ✅

PR #237, opened 2023-12-18
Archive quarterly CEMS data in Annual files

jdangerx:
Partial fix for #230.

This just takes every year's four quarters and puts them into one ZIP file to avoid the 100 file limit in Zenodo.

You can check out the output [here](https://sandbox.zenodo.org/records/13474).

Here's the `resources` section of `datapackage.json` - I decided to format the partitions like `{"year": XXXX, "quarters": [1, 2, 3, 4]}`.

```json
{
    ...
    "resources": [
        {
            "profile": "data-resource",
            "name": "epacems-1995.zip",
            "path": "https://sandbox.zenodo.org/records/13474/files/epacems-1995.zip",
            "remote_url": "https://sandbox.zenodo.org/records/13474/files/epacems-1995.zip",
            "title": "epacems-1995.zip",
            "parts": {
                "year": 1995,
                "quarter": [
                    1,
                    2,
                    3,
                    4
                ]
            },
            "encoding": "utf-8",
            "mediatype": "application/zip",
            "format": ".zip",
            "bytes": 78118649,
            "hash": "8efa7789255fbfac75ed291d3a41f25b"
        },
        {
            "profile": "data-resource",
            "name": "epacems-1996.zip",
            "path": "https://sandbox.zenodo.org/records/13474/files/epacems-1996.zip",
            "remote_url": "https://sandbox.zenodo.org/records/13474/files/epacems-1996.zip",
            "title": "epacems-1996.zip",
            "parts": {
                "year": 1996,
                "quarter": [
                    1,
                    2,
                    3,
                    4
                ]
            },
            "encoding": "utf-8",
            "mediatype": "application/zip",
            "format": ".zip",
            "bytes": 75959903,
            "hash": "7cba31cfd16e73511831c6547b104381"
        }
    ],
    "created": "2023-12-18 17:02:54.530469",
    "version": "2.0.0"
}
```


src/pudl_archiver/archivers/classes.py
```diff
@@ -286,12 +286,14 @@ def dataset_validate_archive(
...
         return []
 
-    def valid_year(self, year: int | str):
+    def valid_year(self, year: int | str | None):
         """Check if this year is one we are interested in.
 
         Args:
             year: the year to check against our self.only_years
         """
+        if not year:
```
jdangerx, line 295: Just dealing with some complaints from the typechecker. I suppose I could move the "if not year" logic to the caller, too, but that would entail either pulling that calling code into a separate function or adding a weird conditional inside a list comprehension, which sucks.

src/pudl_archiver/archivers/epacems.py
```diff
@@ -15,6 +20,38 @@
 logger = logging.getLogger(f"catalystcoop.{__name__}")
 
 
+class BulkFile(BaseModel):
```
jdangerx, line 23: Wasn't *necessary* for this change, but found it slightly helpful to document the expected shape of the API responses in code - and made the type annotations more useful.

src/pudl_archiver/archivers/epacems.py
```diff
@@ -65,27 +104,33 @@ async def get_year_quarter_resource(
...
+            # Useful to debug at download time-outs.
+            logger.info(f"Downloading {year} Q{quarter} EPACEMS data from {url}.")
+
+            # Create zipfile to store year/quarter combinations of files
+            filename = f"epacems-{year}q{quarter}.csv"
+            file_path = self.download_directory / filename
+
+            await self.download_file(url=url, file=file_path, timeout=60 * 14)
+
+            with zipfile.ZipFile(
```
jdangerx, line 121: I didn't quite want to mess with how `download_file_to_zip` worked for *all* archivers yet, so I inlined this. But I think probably a more useful thing to do is have a couple useful utility functions like "download file (raw)," "download file (zip it)," and "download file (add to an existing archive)" that aren't coupled to the class hierarchy. Out of scope for this PR, I think.

src/pudl_archiver/archivers/epacems.py
```diff
@@ -65,27 +104,33 @@ async def get_year_quarter_resource(
...
+                archive_path,
+                "a",
+                compression=zipfile.ZIP_DEFLATED,
+            ) as archive, file_path.open("rb") as f:
+                archive.writestr(filename, f.read())
+
+            file_path.unlink()
```
jdangerx, line 128: Since we're limiting concurrency at the *resource* level we need to clean up the temporary CSV file manually here.

jdangerx (commented):
I'm still getting some wacky type complaints from the `get_resources() -> ArchiveAwaitable` call, and ended up spending quite some time trying to get the correct amount of `await`-ing in the right places for the download to work.

I have some hopes and dreams of trying out a refactor that would make this easier, but not sure if that's really the right place to put my time right now.

src/pudl_archiver/archivers/epacems.py
```diff
@@ -65,27 +104,33 @@ async def get_year_quarter_resource(
...
+
         return ResourceInfo(
-            local_path=archive_path, partitions={"year_quarter": f"{year}q{quarter}"}
+            local_path=archive_path,
+            partitions={
+                "year": year,
+                "quarter": sorted([file.metadata.quarter for file in files]),
```
jdangerx, line 134: @cmgosnell - this is just the first partition format I came up with, we can change it to suit whatever we want to use downstream. @e-belfer and I chatted about this and think maybe it should be `partitions={"year_quarter": sorted([f"{year}q{file.metadata.quarter}" for file in files])}` ? What do you think?

src/pudl_archiver/archivers/classes.py
e-belfer: Is this because sometimes there is no year partition? Not following the logic here of adding `None`

src/pudl_archiver/archivers/classes.py
```diff
@@ -321,6 +323,7 @@ async def download_all_resources(
         for resource_chunk in resource_chunks:
             for resource_coroutine in asyncio.as_completed(resource_chunk):
                 resource_info = await resource_coroutine
+                self.logger.warning(f"{resource_info=}")
```
e-belfer, line 326: Why `warning`? Is this leftover from testing?

src/pudl_archiver/archivers/epacems.py
```diff
@@ -34,29 +71,31 @@ async def get_resources(self) -> ArchiveAwaitable:
...
-            bulk_files = json.loads(resjson)
+            bulk_files = (BulkFile(**f) for f in json.loads(resjson))
             quarterly_emissions_files = [
                 file
                 for file in bulk_files
-                if (file["metadata"]["dataType"] == "Emissions")
-                and (file["metadata"]["dataSubType"] == "Hourly")
-                and ("quarter" in file["metadata"])
-                and (self.valid_year(file["metadata"]["year"]))
+                if (file.metadata.data_type == "Emissions")
```
e-belfer, line 78: so much neater, thank you!

src/pudl_archiver/archivers/epacems.py
```diff
@@ -34,29 +71,31 @@ async def get_resources(self) -> ArchiveAwaitable:
...
+                and self.valid_year(file.metadata.year)
             ]
             logger.info(f"Downloading {len(quarterly_emissions_files)} total files.")
             logger.debug(f"File info: {quarterly_emissions_files}")
-            for i, cems_file in enumerate(quarterly_emissions_files):
-                yield self.get_year_quarter_resource(
-                    file=cems_file, request_count=i + 1
-                )  # Count files starting at 1 for human legibility.
+            files_by_year = groupby(
+                sorted(quarterly_emissions_files, key=lambda bf: bf.metadata.year or 0),
```
e-belfer, line 86: Why 0 here?

src/pudl_archiver/archivers/epacems.py
e-belfer: The concurrency is for each epacems resource which itself is configured to run by year, so I'm not sure whether it would make a difference here? But shouldn't hurt.

src/pudl_archiver/archivers/epacems.py
e-belfer: +1 to this suggestion.

e-belfer (approved):
This works as expected, I have a few questions about specific lines where clarification would be helpful though.

src/pudl_archiver/archivers/classes.py
jdangerx: Yes indeed! Thank you.

src/pudl_archiver/archivers/epacems.py
jdangerx: This is also type stuff to handle "year might not be in metadata." But I think there's a way around it :)

.github/workflows/run-archiver.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           ZENODO_TOKEN_UPLOAD: ${{ secrets.ZENODO_TOKEN_UPLOAD }}
           ZENODO_TOKEN_PUBLISH: ${{ secrets.ZENODO_TOKEN_PUBLISH }}
         run: |
-          pudl_archiver --datasets ${{ matrix.dataset }} --summary-file ${{ matrix.dataset }}_run_summary.json
+          pudl_archiver --datasets ${{ matrix.dataset }} --summary-file ${{ matrix.dataset }}_run_summary.json --sandbox
```
e-belfer, line 54: We'll want to revert this before merging, just a flag!

src/pudl_archiver/archivers/epacems.py
```diff
@@ -61,48 +61,55 @@ class EpaCemsArchiver(AbstractDatasetArchiver):
     base_url = "https://api.epa.gov/easey/bulk-files/"
     parameters = {"api_key": os.environ["EPACEMS_API_KEY"]}  # Set to API key
 
+    def __filter_for_complete_metadata(
```
e-belfer, line 64: Thinking aloud here - the only issue here could be if the underlying format of the metadata changes (e.g., year becomes string type), and we'd silently lose all the files. But I guess this would be caught in our other validation tests (e.g. check missing files).

e-belfer (approved):
Changes look great, the handling of the metadata types seems much neater with the additional filtering method.

PR #240, opened 2023-12-19
Zip EIA 860M files by year

### LINKED ISSUE #239: Handle 100 file limit for EIA 860M Zenodo archiving
Zenodo started enforcing their 100-files-per-deposition rule via the API, and EIA 860M has more than 100 files in it, so we need to handle that.

```[tasklist]
## success criteria
- [x] zip all months for one year into a single file
- [x] datapackage.json associates list of `year_month` partitions with each zip file
```

jdangerx:
# Overview

Closes #239 - we can get around the 100-file limit on Zenodo by zipping things into fewer files.

Currently stacked on top of `multi-zip-cems` but could easily just get rebased on top of `main` once `stop-discarding-drafts` gets merged.  Though the `add_to_archive` method would maybe cause a conflict... but it should be all pretty straightforward.

We'll have to handle the new partition format downstream - I think @cmgosnell had volunteered for this?

# Testing

I ran `pudl_archiver --datasets eia860m  --sandbox --only-years 2020 2021` and got [this output](https://sandbox.zenodo.org/records/13857). Good signs:

- partitions in `datapackage.json` look good and include all the `202X-XX` months we expect
- zip files are ~90 MB each and include 12 8MB files each - makes sense, xlsx files are already compressed
- one file per year

### DIFF
--- a/src/pudl_archiver/archivers/epacems.py
+++ b/src/pudl_archiver/archivers/epacems.py
@@ -3,7 +3,6 @@ import datetime
 import json
 import logging
 import os
-import zipfile
 from collections.abc import Iterable
 from itertools import groupby
 
@@ -16,6 +15,7 @@ from pudl_archiver.archivers.classes import (
     ArchiveAwaitable,
     ResourceInfo,
 )
+from pudl_archiver.frictionless import ZipLayout
 
 logger = logging.getLogger(f"catalystcoop.{__name__}")
 
@@ -112,6 +112,7 @@ class EpaCemsArchiver(AbstractDatasetArchiver):
             files: the files we've associated with this year.
         """
         archive_path = self.download_directory / f"epacems-{year}.zip"
+        data_paths_in_archive = set()
         for file in files:
             url = self.base_url + file.s3_path
             quarter = file.metadata.quarter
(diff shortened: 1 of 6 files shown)


src/pudl_archiver/archivers/epacems.py
```diff
@@ -141,4 +140,5 @@ async def get_year_resource(
                     [f"{year}q{file.metadata.quarter}" for file in files]
                 ),
             },
+            layout=ZipLayout(file_paths=data_paths_in_archive),
```
e-belfer, line 143: @aesharpe Probably relevant for your PR!

tests/unit/eia_archiver_test.py
```diff
@@ -50,7 +50,7 @@ async def test_eia860m(mocker):
     )
     archiver = Eia860MArchiver(mock_session, only_years=[2019, 2022])
     resources = [res async for res in archiver.get_resources()]
-    assert len(resources) == 24
+    assert len(resources) == 2
```
e-belfer, line 53: Do we want to / can we assert anything about the contents/structure of these resources (e.g. that the number of files they contain == 12?)

e-belfer (approved):
This works as expected for me and the datapackage looks good! Appreciate the cleanup of the methods across archivers with the multi-zipping. Only thing is that I can't speak to the exact mechanics of the async changes made, defer to @zschira here.

tests/unit/eia_archiver_test.py
jdangerx: Just added some mocks to do this!

PR #243, opened 2023-12-19
Make message more compact & limit it to 3000 characters

jdangerx:
Archive-notify failed due to the 3001 character limit:

https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/7265550885/job/19795893314

I made the output a bit less verbose, truncated the output, and added the fallback `"text"` field mentioned in the error message above.

Tested by running 

```
pudl_archiver --datasets eia860m  --sandbox --only-years 2018 2020 --summary-file eia860m.json
```

And then 

```
./scripts/make_slack_notification_message.py --summary-files *-summary.json | pbcopy
```

And then pasting that output into https://app.slack.com/block-kit-builder/ (removing the top-level `text` field) to verify that it looked fine.

### DIFF
--- a/scripts/make_slack_notification_message.py
+++ b/scripts/make_slack_notification_message.py
@@ -17,6 +17,7 @@ hide large messages (such as a file diff) behind a "See more" action.
 import argparse
 import itertools
 import json
+from collections import defaultdict
 from pathlib import Path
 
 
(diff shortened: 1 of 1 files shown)


e-belfer (approved):
What an annoying code restriction. As written this works for me, so I'm approving it.

Non-blocking: If we're going to reformat for slack (which seems fine!) I wonder if just making a sorted list of file names under each type of change with entries - UPDATE, CREATE, and DELETE - would be easier to read and shorter. Seems like it'd both improve legibility and shorten the message further.

jdangerx:
You're thinking instead of some code-formatted JSON blob:

```
{
  "CREATED": [
    "foo.csv",
    "bar.csv"
  ],
  "UPDATED": ...
}
```

Having that actually formatted as a markdown list?

PR #244, opened 2023-12-22
Update artifact uploads due to breaking changes in v4

jdangerx:
They made artifacts immutable in the upload/download-artifacts@v4 versions.

Just naively applying the [migration guide](https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact) here.


e-belfer (approved):
Looks good to me, but of course we'll have to test it and see!

PR #251, opened 2024-01-09
Force zipfiles to be identical by fixing timestamps of zipped files

e-belfer:
The MD5 checksums of zipfiles we make (CEMS, 860M) change even when the files zipped in them stay the same. This is because the file modification time changes. This PR manually fixes the `datetime` of files when we add them to a zipfile repository. 

To test:
`pudl_archiver --datasets eia860m --sandbox --initialize --summary-file test.json --only-years 2015`
Run twice and compare checksums on the zipped file for each repository.

### DIFF
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -15,7 +15,7 @@ import pandas as pd
 
 from pudl_archiver.archivers import validate
 from pudl_archiver.frictionless import DataPackage, ResourceInfo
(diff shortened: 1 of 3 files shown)


jdangerx (approved):
Sweet, I'm glad this works!

It might make sense to extract this common logic of "add without metadata" to a little helper function. That makes the purpose of the code really clear, and also lets us cover this kind of funky logic with unit tests more easily. I'm happy to do that in a follow-up PR if we have another hour or so in the budget.

PR #252, opened 2024-01-11
Change PHMSA partitions to use `year` instead of `years`

e-belfer:
To make the excel extractor in `pudl` work consistently with the PHMSA data, update the `years` partition to `year` to match all other partitions. See draft dataset at: https://zenodo.org/deposit/10493790


cmgosnell (approved):
wow this looks a little complicated.... 👀 

loljk 🟢

PR #253, opened 2024-01-11
When PHMSA data has `_present` in file name, make partitions actually reflect years of data available correctly.

e-belfer:
PHMSA data has a `_present` filename that doesn't provide information about the data end date. Because the data isn't released right at the end of the year, our prior "assume from the prior year" assumption doesn't hold at the beginning of the year. To make this more robust, look inside all `_present` files and get the last year's worth of data from the file name contained inside, by sorting the file names and getting the year from the last portion of the filename string. Draft archive here: https://zenodo.org/uploads/10493790

### DIFF
--- a/src/pudl_archiver/archivers/phmsagas.py
+++ b/src/pudl_archiver/archivers/phmsagas.py
@@ -85,14 +85,14 @@ class PhmsaGasArchiver(AbstractDatasetArchiver):
 
         # If end year is present, open zipfile and get last year of excel data in it.
         if end_year == "present":
-            excel_files = sorted(
+            file_years = sorted(
                 [
-                    file
+                    int(file.split("_")[-1].replace(".xlsx", ""))
                     for file in ZipFile(download_path).namelist()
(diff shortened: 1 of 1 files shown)


src/pudl_archiver/archivers/phmsagas.py
```diff
@@ -76,15 +76,26 @@ async def get_zip_resource(
...
+        # If end year is present, open zipfile and get last year of excel data in it.
+        if end_year == "present":
+            excel_files = sorted(
+                [
+                    file
+                    for file in ZipFile(download_path).namelist()
+                    if file.endswith(".xlsx")
+                ]
+            )
+            end_year = int(excel_files[-1].split("_")[-1].replace(".xlsx", ""))
```
cmgosnell, line 95: okay okay your sorting them so the last one is the one with the max year? grabbed via the -1? this is not a blocking concern at all but i think it would be more readable if you attempted to grab the year out of each file and `end_year` was just the max

cmgosnell (approved):

Issue #256, opened 2024-01-18
EIA Bulk Electricity Archiver fails


zaneselvans:
When attempting to run create a fresh archive of the bulk EIA electricity data:

```console
pudl_archiver --datasets eia_bulk_elec --download-dir /tmp/eia_bulk_elec
```

I got the following error:

```
INFO:catalystcoop.pudl_archiver.depositors.zenodo:PUT https://zenodo.org/api/files/fe6cdea2-5553-4a26-b57a-4aad22403d97/datapackage.json - Uploading datapackage.json to bucket
Encountered exceptions, showing traceback for last one: ["('eia_bulk_elec', 1 validation error for RunSummary\nprevious_version\n  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]\n    For further information visit https://errors.pydantic.dev/2.5/v/string_type)"]
Traceback (most recent call last):
  File "/Users/zane/miniforge3/envs/pudl-cataloger/bin/pudl_archiver", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/zane/code/catalyst/pudl-archiver/src/pudl_archiver/cli.py", line 88, in main
    asyncio.run(archive_datasets(**vars(args)))
  File "/Users/zane/miniforge3/envs/pudl-cataloger/lib/python3.11/asyncio/runners.py", line 190, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-cataloger/lib/python3.11/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/miniforge3/envs/pudl-cataloger/lib/python3.11/asyncio/base_events.py", line 653, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl-archiver/src/pudl_archiver/__init__.py", line 108, in archive_datasets
    raise exceptions[-1][1]
  File "/Users/zane/code/catalyst/pudl-archiver/src/pudl_archiver/orchestrator.py", line 200, in run
    run_summary = self._summarize_run(
                  ^^^^^^^^^^^^^^^^^^^^
  File "/Users/zane/code/catalyst/pudl-archiver/src/pudl_archiver/orchestrator.py", line 234, in _summarize_run
[...]

PR #257, opened 2024-01-18
Add `eia_bulk_elec` to monthly automated archiver runs

e-belfer:
The issue found in #256 has been fixed through the creation of a new archive (v 3.0). This PR simply adds the `eia_bulk_elec` dataset to the monthly scheduled archiver run, as this dataset is relatively regularly updated. It also bumps `fiona` to `>=1.9.5` to match the version in PUDL and fix a CI issue.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -8,9 +8,5 @@ dependencies:
   - pip>=21,<24
   - python>=3.11,<3.12
   - setuptools>=66
-  # fiona is a transitive dependency which needs GDAL. so we install with conda
-  # TODO: we shouldn't have to install all this geo stuff, so once we break the
-  # dependency on `pudl` we should remove this too.
-  - fiona>=1.9.5
   - pip:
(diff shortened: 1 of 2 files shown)


environment.yml
```diff
@@ -11,6 +11,6 @@ dependencies:
   # fiona is a transitive dependency which needs GDAL. so we install with conda
   # TODO: we shouldn't have to install all this geo stuff, so once we break the
   # dependency on `pudl` we should remove this too.
-  - fiona>=1.8
+  - fiona>=1.9.5
```
zaneselvans, line 14: I'm not 100% sure if this is necessary (though presumably something was broken at some point and that's why it got added). But we've changed our dependency management / specs quite a bit in the PUDL repo since then. Maybe try remove this entirely just to see if it works?

e-belfer: Well, that didn't work!

zaneselvans (approved):

PR #261, opened 2024-01-25
Add explicit frictionless dependency since we import it direclty.

zaneselvans:
(I also sorted the dependencies...)


jdangerx:
Oh, woah, we were able to import frictionless because we depend on `ferc-xbrl-extractor` as a transitive dependency? Wack.

jdangerx (approved):

PR #267, opened 2024-01-29
Update archiver for EIA 176, write new archivers for Forms 191, 757A

### LINKED ISSUE #262: Update EIA 176 archiver to grab CSVs, rather than bundled data
We're currently downloading the 'bundled' EIA 176 data (https://www.eia.gov/naturalgas/ngqs/all_ng_data.zip). This data large CSVs with all years of data, contains only 2017 data for form 757 and a `LINE` column requiring translation for EIA 176. Instead, let's shift to querying the EIA website for CSVs for individual years of data from the form partitions, as discussed [here](https://github.com/catalyst-cooperative/pudl/issues/2603#issuecomment-1908880586).

For 191 and 757 data, this will look like querying the subsections of the form (e.g., 191 monthly) into individual CSVs, and zipping them. 

For 176, there are subsections of the form available but they don't seem to capture all lines. Instead, we'll use the "custom export" endpoint to iteratively download chunks of the form at a time, and then combine them all together into one annual CSV. This will look like querying the `items` endpoint, getting a list of all possible 176 line codes, and iteratively building URLs to query and compile all lines of the 176 form for a single year into one CSV, which is then zipped.

In the end, this should look like:
- three archivers, one each for EIA forms 176, 757, 191. 757 and 191 should be quite similar.
- one zipfile per year, containing one file per section with all form fields
- for 757: an archiver that tries all combinations of years and logs a note if they're not available (to capture irregular years of reporting for 757), or uses the `report` endpoint to generate start and end years

```[tasklist]
- [x] Write archiver for 176 that iteratively gets all items and compiles them
- [ ] https://github.com/catalyst-cooperative/pudl/pull/3304/
- [x] Write archiver for 191
- [x] Write near-identical archiver for 757
```

e-belfer:
Closes #262. Switch to scraping NGQV data, rather than relying on the bundled zipfile which requires manual line mapping. This PR produces one zipped CSV per year per form. This PR also reorganizes all EIA archivers into an `eia` subfolder to make subclassing the natural gas archiver logic more straightforward and to mirror the `ferc` format.

### Testing
See sandbox archives for [Form 176](https://sandbox.zenodo.org/uploads/24541), [Form 191](https://sandbox.zenodo.org/uploads/25172) and [Form 757A](https://sandbox.zenodo.org/uploads/25165).


src/pudl_archiver/archivers/eia191.py
```diff
@@ -0,0 +1,98 @@
...
+    EIANaturalGasData,
+    ResourceInfo,
+)
+from pudl_archiver.frictionless import ZipLayout
+from pudl_archiver.utils import add_to_archive_stable_hash
+
+logger = logging.getLogger(f"catalystcoop.{__name__}")
+
+
+class Eia191Archiver(AbstractDatasetArchiver):
```
zschira, line 20: As for sharing code between 191 and 757A, the only real differences I see between the two archivers are: the base url/`form` that get passed to `get_datasets`, the filter for `Monthly` in 191 `get_resources`, and the sorting by `company_name` and `field_name` for 191. With these minimal differences in mind, my inclination would be to create a base class that both of these archivers inherit from. The base class would inherit from `AbstractDatasetArchiver` and implement `get_datasets` and `get_year_resource`, leaving out the sorting that done in the 191 implementation (could this be done later in the ETL?). This way the `Eia191Archiver` and `Eia757AArchiver` could just set their `name` and `base_url`, then implement `get_resources` where 191 can filter for monthly data.

src/pudl_archiver/__init__.py
```diff
@@ -27,7 +29,18 @@ def all_archivers():
     for module in module_names:
         # AbstractDatasetArchiver won't know about the subclasses unless they are imported
         __import__(module)
-    return AbstractDatasetArchiver.__subclasses__()
+
+    def all_subclasses(cls):
```
e-belfer, line 33: @zschira This is the only part of the transition to subclassing the NGQV archiver that felt a bit questionable, but otherwise we wind up having a non-functional NGQV archiver class in the list and not including 191 or 757A. If you have suggestions on how to do this better I'm all ears.

zschira: Hmm yeah it's a bit annoying, but I might just leave it. We could pull out the `get_datasets` and `get_year_resource` methods to be standalone functions that get called by the archivers, but I'm not sure if it's actually worth it

zschira (approved):
Looks good to me! One small typo, and as for the weird subclassing thing, I think it might be fine for now. I think there's probably better ways to make the top-level process aware of all the archivers, but that's probably a project for a future day.

Issue #245, opened 2023-12-22
Zipfile validation not catching corrupted zipfiles


e-belfer:
The [most recent CEMS archive](https://zenodo.org/records/10420843) contains a corrupted file for the 2006 data. Yet the zipfile validation test did not catch this. `is_zipfile` is True for this file, while `ZipFile.test_file()` raises a `OSError`. It seems like we may want to revisit the `validate_zipfile` test method developed in #177, as it isn't catching as much as we need for it to be able to be our only failsafe.

PR #268, opened 2024-01-31
Update zipfile type check validation and add file checks into summary JSON files

e-belfer:
This PR addresses #245:

- adds `test_zip()` into the zipfile file check method.
- updates the validation summary to include file checks by default
- adds a test to ensure bad zipfiles fail our file check


zschira (approved):
Looks good!

PR #270, opened 2024-02-01
Make CEMS GHA runnable

e-belfer:
This PR breaks out the `run-archiver` method into 2 streams, and uses a large runner for the CEMS archive run.


zaneselvans:
If the CEMS is carved out to run as its own job, I think running it on a bigger runner with more disk space is probably the quickest, easiest fix. If it only runs once a month the cost will be negligible. It'll also have 2x the memory & CPU available, so we can probably increase the concurrency to 4 and it should run in ~half the time (so like half an hour?)

.github/workflows/run-archiver.yml
```diff
@@ -7,27 +7,77 @@ on:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: run-summaries-${{ matrix.dataset }}
+          path: ${{ matrix.dataset }}_run_summary.json
+
+  archive-run-cems:
+    defaults:
+      run:
+        shell: bash -l {0}
+    strategy:
```
jdangerx, line 70: non-blocking: you *could* rip out this whole `strategy` block and just hard-code `epacems` instead of using `${{ matrix.dataset }}` but 🤷 this works and is clear enough.

jdangerx (approved):
Reviewed the non-GHA stuff in #275; the actions stuff looks good too!

.github/workflows/run-archiver.yml
e-belfer: I imagine if we do start regularly archiving EQR it'll go in here as well, so I'll just leave it.

PR #275, opened 2024-02-08
Chunk downloads, handle session timeout, and fix CEMS zipping

e-belfer:
# Overview

Fixes persistent CEMS archiver failures.

- Add chunked file downloads
- Increase the session timeout
- Fix issues with zipfile compression introduced in #251 
- Fix GHA to upload run summary even if archiver fails
- Fix issues with new archivers (176, 191 and 757A) - add new archives to `datasetdoi.yaml`, improve JSON handling

Out of scope:
- Debug EIA server rejections of GHA queries that work locally, which makes automating the 176 archive currently impossible. See https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/7845007723/job/21408517832

# Testing

GHA works and produces new archives for all datasets currently in the list.

# To-do list

```[tasklist]
```


src/pudl_archiver/archivers/epacems.py
```diff
@@ -57,6 +57,7 @@ class EpaCemsArchiver(AbstractDatasetArchiver):
 
     name = "epacems"
     concurrency_limit = 2  # Number of files to concurrently download
+    allowed_file_rel_diff = 0.35  # Set higher tolerance than standard
```
jdangerx, line 60: Is there a particular reason why we want to set this tolerance higher? bc the files are big and few?

src/pudl_archiver/utils.py
```diff
@@ -70,4 +70,5 @@ def add_to_archive_stable_hash(archive: zipfile.ZipFile, filename, data: bytes):
         # Set fixed date to enable hash comparisons between archives
         date_time=(1980, 1, 1, 0, 0, 0),
     )
+    info.compress_type = zipfile.ZIP_DEFLATED
```
jdangerx, line 73: Huh! I guess the default value here is `zipfile.ZIP_STORED` which presumably causes some confusion in the decompression. TIL!

Also, surprising that compress_type can't be set at initialization time 🤷 

Might be useful to clarify these things in a comment.

```suggestion
    # default is ZIP_STORED, which means "uncompressed"; also this can't be set in the constructor as of 2024-02-09
    info.compress_type = zipfile.ZIP_DEFLATED
```

tests/unit/archive_base_test.py
```diff
@@ -196,43 +197,34 @@ async def test_download_zipfile(mocker, bad_zipfile, good_zipfile):
 
 
 @pytest.mark.asyncio
-async def test_download_file(mocker, file_data):
+async def test_download_file(mocker, tmp_path):
```
jdangerx, line 200: So this now downloads a real file with `requests`, and then tries to download it again with `AbstractDatasetArchiver.download_file`, and compares that they're the same - is that right?

And if I understand this all right, the point of ripping out the mocks is that we test that `ClientSession.get()` does what we think it does, vs. just assuming that the mocks are reasonable. Is that right also?

Seems smart to me - we gotta test the IO stuff somewhere, might as well test it in a very narrowly focused IO test. But we should change the docstring :)

jdangerx (changes requested):
Nice sleuthing y'all! Just a few small changes to clarify things / avoid using deprecated APIs. Ping me aggressively for re-review!

src/pudl_archiver/archivers/epacems.py
e-belfer: Yes, if we have one quarter of data and we add a second quarter it'll be substantially more than a 25% increase in file size

jdangerx (approved):

PR #284, opened 2024-02-20
Automate everything! Add remaining data sources to GHA

### LINKED ISSUE #276: Automate remaining archive runs
Right now in `run-archiver.yml` we run EIA 191, 757A, 860, 923, and EPACEMS monthly. Let's automate all our archiver runs to enable regular updates and debugging of the archivers. If not monthly, let's at least run these on a quarterly schedule:
- [x] `eia176`: Resolve server rejections, see https://github.com/catalyst-cooperative/pudl-archiver/pull/275
- [x] `eia861`: #284
- [x] `phmsagas`: #284 
- [ ] `eiawater`
- [ ] `ferc1`
- [ ] `ferc2`
- [ ] `ferc6`
- [ ] `ferc60`
- [ ] `ferc714`
- [x] `epacamd_eia`: #284 
- [x] `mshamines`: #284

e-belfer:
# Overview

Closes #276 

Makes the rest of our archives run monthly, and sorts `file_changes` to improve the legibility of reported changes.

Adding remaining data sources to the `run-archiver` method to prevent stale archives, and adding some quality of life improvements to the summary.

# Testing

Run `run-archiver` in GHA. All datasets should pass.

# To-do list

```[tasklist]
```

### DIFF
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -74,63 +74,63 @@ jobs:
           name: run-summaries-${{ matrix.dataset }}
           path: ${{ matrix.dataset }}_run_summary.json
 
-  archive-run-cems:
-    defaults:
-      run:
-        shell: bash -l {0}
-    strategy:
-      matrix:
-        dataset:
-          - epacems
-      fail-fast: false
-    runs-on:
-      group: large-runner-group
-      labels: ubuntu-22.04-4core
-    steps:
-      - uses: actions/checkout@v4
+  # archive-run-cems:
+  #   defaults:
+  #     run:
+  #       shell: bash -l {0}
+  #   strategy:
+  #     matrix:
+  #       dataset:
+  #         - epacems
+  #     fail-fast: false
+  #   runs-on:
+  #     group: large-runner-group
+  #     labels: ubuntu-22.04-4core
+  #   steps:
+  #     - uses: actions/checkout@v4
 
-      - name: Install Conda environment using mamba
-        uses: mamba-org/setup-micromamba@v1
-        with:
-          environment-file: environment.yml
-          cache-environment: true
-          condarc: |
-            channels:
-            - conda-forge
-            - defaults
-            channel_priority: strict
+  #     - name: Install Conda environment using mamba
+  #       uses: mamba-org/setup-micromamba@v1
+  #       with:
+  #         environment-file: environment.yml
+  #         cache-environment: true
+  #         condarc: |
+  #           channels:
+  #           - conda-forge
+  #           - defaults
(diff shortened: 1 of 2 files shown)


.github/workflows/run-archiver.yml
```diff
@@ -15,12 +15,22 @@ jobs:
...
           - eia860
           - eia860m
           - eia923
           - eia_bulk_elec
+          - eiawater
+          - ferc1
+          - ferc2
+          - ferc6
+          - ferc60
+          - ferc714
```
zaneselvans, line 30: Do we still expect the FERC XBRL archives will always appear completely new, even if nothing has changed in their contents?

IIRC they were doing something like autogenerating new IDs for every post in the RSS feed every time the feed was read, rather than using persistent unique IDs for each post. Did we find some way around that?

e-belfer: FERC archivers are certainly not working at present, but I'm tracking this in #285.

zaneselvans: Hmm, #285 looks like real brokenness in the FERC archivers.

If we haven't addressed the unique ID thing, we'd just see that all the FERC XBRL archives get updated every time the archiver is run, but I think we'd still be able to get an idea of how much new data there is from the change in the size of the archives, and saving interim data doesn't seem like a bad idea given how flaky FERC's data curation is!

.github/workflows/run-archiver.yml
```diff
@@ -18,9 +18,11 @@ jobs:
           - eia191
           - eia757a
           - eia860
+          - eia861
           - eia860m
           - eia923
           - eia_bulk_elec
+          - phmsagas
```
zaneselvans, line 25: I notice that this job is called `archive-run-eia` but it contains non-EIA datasets. Do we intend for the jobs to be broken out by data source? It seems like the workflow is primarily organized by whether a small vs. large runner is required.

It's not clear to me whether we really need to fix the cause of the "previous version" errors for `mshamines` and `epacamd_eia` that you note in #285 before adding those datasets in here (or ever). If approving the draft archives will make future versions work, I think it's just a one-time change in the available metadata and shouldn't be a problem in the future should it? (or rather, it'll be a problem, but just once for each dataset, with the same solution, of approving the new draft archive despite the error).

zaneselvans (changes requested):

.github/workflows/run-archiver.yml
e-belfer: Great catch. I'll rename this job and run both to see if manual approval fixes the error, then add them back.

zaneselvans (approved):

PR #289, opened 2024-03-08
Decouple depositors from zenodo to allow different storage backends

zschira:
This PR creates an abstract interface for Depositors to easily allow new storage backends for our archivers. It also implements the interface for the zenodo depositor.

### DIFF
--- a/.github/workflows/tox-pytest.yml
+++ b/.github/workflows/tox-pytest.yml
@@ -22,6 +22,7 @@ jobs:
         with:
           environment-file: environment.yml
           cache-environment: true
+          cache-environment-key: environment-${{ hashFiles('pyproject.toml') }}
           condarc: |
             channels:
             - conda-forge
--- a/src/pudl_archiver/__init__.py
+++ b/src/pudl_archiver/__init__.py
@@ -85,16 +85,11 @@ async def archive_datasets(
                 run_settings.only_years,
                 download_directory=run_settings.download_dir,
             )
-            depositor = await ZenodoPublishedDeposition.get_latest_version(
-                dataset=dataset,
-                session=session,
-                settings=run_settings,
-            )
             tasks.append(
                 orchestrate_run(
                     dataset,
                     downloader,
-                    depositor,
+                    run_settings,
                     session,
                 )
             )
--- a/src/pudl_archiver/__init__.py
+++ b/src/pudl_archiver/__init__.py
@@ -115,7 +115,7 @@ async def archive_datasets(
     if run_settings.summary_file is not None:
         run_summaries = [
             result.dict()
-            for _, result in results
+            for _, [result, published] in results
             if not isinstance(result, BaseException)
         ]
 
@@ -124,7 +124,9 @@ async def archive_datasets(
 
     # Check validation results of all runs that aren't unchanged
     validation_results = [
-        result.success for _, result in results if isinstance(result, RunSummary)
+        result.success
+        for _, [result, published] in results
+        if isinstance(result, RunSummary)
     ]
     if not all(validation_results):
         raise RuntimeError("Error: archive validation tests failed.")
--- a/src/pudl_archiver/__init__.py
+++ b/src/pudl_archiver/__init__.py
@@ -9,8 +9,7 @@ import aiohttp
 import pudl_archiver.orchestrator  # noqa: F401
 from pudl_archiver.archivers.classes import AbstractDatasetArchiver
 from pudl_archiver.archivers.validate import RunSummary
-from pudl_archiver.depositors.depositor import Depositor
-from pudl_archiver.depositors.zenodo.depositor import ZenodoDepositorInterface
+from pudl_archiver.depositors.zenodo.depositor import ZenodoPublishedDeposition
 from pudl_archiver.orchestrator import orchestrate_run
 from pudl_archiver.utils import RunSettings
 
@@ -85,8 +84,7 @@ async def archive_datasets(
                 run_settings.only_years,
                 download_directory=run_settings.download_dir,
             )
-            depositor = await Depositor.get_latest_version(
-                interface=ZenodoDepositorInterface,
+            depositor = await ZenodoPublishedDeposition.get_latest_version(
                 dataset=dataset,
                 session=session,
                 settings=run_settings,
--- a/src/pudl_archiver/__init__.py
+++ b/src/pudl_archiver/__init__.py
@@ -1,4 +1,5 @@
 """Tool to download data resources and create archives on Zenodo for use in PUDL."""
+
 import asyncio
 import json
 import logging
--- a/src/pudl_archiver/cli.py
+++ b/src/pudl_archiver/cli.py
@@ -26,9 +26,10 @@ def parse_main():
     parser.add_argument(
         "--only-years",
         nargs="*",
-        help="Years to download data for. Supported datasets: eia860, "
-        "eia860m, eia861, eia923, eiawater, epacems, ferc1, ferc2, ferc6, "
-        "ferc60, ferc714",
+        help="Years to download data for. Supported datasets: censusdp1tract, eia176, "
+        "eia191, eia757a, eia860, eia860m, eia861, eia923, eia930, eia_bulk_elec, "
+        "eiaaeo, eiawater, epacamd_eia, epacems, ferc1, ferc2, ferc6, ferc60, ferc714, "
+        "mshamines, nrelatb, phmsagas",
         type=int,
     )
     parser.add_argument(
--- a/src/pudl_archiver/cli.py
+++ b/src/pudl_archiver/cli.py
@@ -14,7 +14,7 @@ from pudl_archiver.utils import RunSettings
 logger = logging.getLogger("catalystcoop.pudl_archiver")
 
 
-def parse_main():
+def parse_main(args=None):
     """Process base commands from the CLI."""
     parser = argparse.ArgumentParser(description="Upload PUDL data archives to Zenodo")
     parser.add_argument(
@@ -79,24 +79,23 @@ def parse_main():
         type=str,
         default="zenodo",
     )
-    return parser.parse_args()
+    return parser.parse_args(args)
 
 
-def main():
+async def main(args=None):
     """Run desired archivers."""
     load_dotenv()
     logger.setLevel(logging.INFO)
     log_format = "%(asctime)s [%(levelname)8s] %(name)s:%(lineno)s %(message)s"
     coloredlogs.install(fmt=log_format, level=logging.INFO, logger=logger)
-    args = parse_main()
+    args = parse_main(args)
     datasets = args.datasets
     if args.all:
         datasets = ARCHIVERS.keys()
     del args.all
-    asyncio.run(
-        archive_datasets(datasets=datasets, run_settings=RunSettings(**vars(args)))
-    )
+
+    await archive_datasets(datasets=datasets, run_settings=RunSettings(**vars(args)))
 
 
 if __name__ == "__main__":
-    main()
+    asyncio.run(main())
--- a/src/pudl_archiver/cli.py
+++ b/src/pudl_archiver/cli.py
@@ -82,7 +82,7 @@ def parse_main(args=None):
     return parser.parse_args(args)
 
 
-async def main(args=None):
+async def archiver_entry(args=None):
     """Run desired archivers."""
     load_dotenv()
     logger.setLevel(logging.INFO)
@@ -97,5 +97,6 @@ async def main(args=None):
     await archive_datasets(datasets=datasets, run_settings=RunSettings(**vars(args)))
 
 
-if __name__ == "__main__":
-    asyncio.run(main())
+def main():
+    """Kick off async script."""
+    asyncio.run(archiver_entry())
--- a/src/pudl_archiver/cli.py
+++ b/src/pudl_archiver/cli.py
@@ -74,6 +74,11 @@ def parse_main():
         action="store_true",
         help="Regenerate metadata from PUDL data source rather than existing archived metadata.",
     )
+    parser.add_argument(
+        "--depositor",
+        type=str,
+        default="zenodo",
+    )
     return parser.parse_args()
 
 
--- a/src/pudl_archiver/depositors/__init__.py
+++ b/src/pudl_archiver/depositors/__init__.py
@@ -2,3 +2,61 @@
 
 Such as Zenodo, Zenodo, and Zenodo.
 """
+
+from dataclasses import dataclass
+
+import aiohttp
+
+from pudl_archiver.frictionless import DataPackage
+from pudl_archiver.utils import RunSettings
+
+from .depositor import (
+    DepositionAction,
+    DepositionChange,
+    DraftDeposition,
+    PublishedDeposition,
+)
+
+
+@dataclass
+class DepositionInterface:
+    """Wrap Published and Draft Deposition classes for a single depositor."""
+
+    published_interface: type[PublishedDeposition]
+    draft_interface: type[DraftDeposition]
+
+
+DEPOSITION_INTERFACES: dict[str, DepositionInterface] = {}
+
+
+def register_depositor(
+    depositor_name: str,
+    published_interface: type[PublishedDeposition],
+    draft_interface: type[DraftDeposition],
+):
+    """Function to register an implementation of the depositor interface."""
+    DEPOSITION_INTERFACES[depositor_name] = DepositionInterface(
+        published_interface=published_interface, draft_interface=draft_interface
+    )
+
+
+async def get_depositor(
+    dataset: str, session: aiohttp.ClientSession, run_settings: RunSettings
+) -> tuple[DraftDeposition, DataPackage | None]:
+    """Create draft deposition from scratch or previous version."""
+    deposition_interface = DEPOSITION_INTERFACES[run_settings.depositor]
+    if run_settings.initialize:
+        return await deposition_interface.draft_interface.initialize_from_scratch(
+            dataset, session, run_settings
+        ), None
+
+    published_deposition = (
+        await deposition_interface.published_interface.get_latest_version(
+            dataset, session, run_settings
+        )
+    )
+
+    original_datapackage_bytes = await published_deposition.get_file("datapackage.json")
+    original_datapackage = DataPackage.model_validate_json(original_datapackage_bytes)
+
+    return await published_deposition.open_draft(), original_datapackage
--- a/src/pudl_archiver/depositors/__init__.py
+++ b/src/pudl_archiver/depositors/__init__.py
@@ -13,30 +13,35 @@ from pudl_archiver.utils import RunSettings
 from .depositor import (
     DepositionAction,
     DepositionChange,
+    DepositorAPIClient,
     DraftDeposition,
     PublishedDeposition,
 )
 
 
 @dataclass
-class DepositionInterface:
+class DepositionBackend:
     """Wrap Published and Draft Deposition classes for a single depositor."""
 
+    api_client: type[DepositorAPIClient]
     published_interface: type[PublishedDeposition]
     draft_interface: type[DraftDeposition]
 
 
-DEPOSITION_INTERFACES: dict[str, DepositionInterface] = {}
+DEPOSITION_BACKENDS: dict[str, DepositionBackend] = {}
 
 
 def register_depositor(
     depositor_name: str,
+    api_client: type[DepositorAPIClient],
     published_interface: type[PublishedDeposition],
     draft_interface: type[DraftDeposition],
 ):
     """Function to register an implementation of the depositor interface."""
-    DEPOSITION_INTERFACES[depositor_name] = DepositionInterface(
-        published_interface=published_interface, draft_interface=draft_interface
+    DEPOSITION_BACKENDS[depositor_name] = DepositionBackend(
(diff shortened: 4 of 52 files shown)


src/pudl_archiver/utils.py
```diff
@@ -74,3 +77,56 @@ def add_to_archive_stable_hash(archive: zipfile.ZipFile, filename, data: bytes):
...
+    for task in tasks:
+        scheduled_tasks.append(asyncio.create_task(_run_task(task, result_queue)))
+        await asyncio.sleep(1 / rate_limit)
+
+    # Wait for all tasks to complete
+    await asyncio.gather(*scheduled_tasks)
+    await result_queue.put("tasks_complete")
+
+
+async def rate_limit_tasks(tasks: list[typing.Awaitable], rate_limit: int = 10):
```
zschira, line 102: I added this function for the sec archiver, but decided to put it in this infrastructure PR. It just launches download tasks at a specified rate until they're all done, and returns outputs as tasks complete.

src/pudl_archiver/utils.py
```diff
@@ -74,3 +77,56 @@ def add_to_archive_stable_hash(archive: zipfile.ZipFile, filename, data: bytes):
...
+    result_queue = asyncio.Queue()
+    asyncio.create_task(
+        _rate_limited_scheduler(tasks, result_queue, rate_limit=rate_limit)
+    )
+
+    while (result := await result_queue.get()) != "tasks_complete":
+        yield result
+
+
+class RunSettings(BaseModel):
```
zschira, line 121: Put all the CLI settings in here to simplify passing values around

src/pudl_archiver/depositors/depositor.py
```diff
@@ -0,0 +1,336 @@
...
+@dataclass
+class DepositionChange:
+    """Define a single change to a file in a deposition."""
+
+    action_type: DepositionAction
+    name: str
+    resource: io.IOBase | Path | None = None
+
+
+class AbstractDepositorInterface(ABC, BaseModel):
```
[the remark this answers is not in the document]
zschira: `AbstractDepositorInterface`'s aren't used directly, but instead wrapped in the `DraftDeposition` and `Depositor` classes.

src/pudl_archiver/depositors/depositor.py
```diff
@@ -0,0 +1,336 @@
...
+            session,
+            settings,
+        )
+        return cls(
+            deposition=interface,
+            settings=settings,
+        )
+
+    @asynccontextmanager
+    async def open_draft(self):
```
zschira, line 315: Returns a context manager representing an open draft deposition, so it can cleanly cleanup if there's an error during a run.

src/pudl_archiver/orchestrator.py
```diff
@@ -1,445 +1,58 @@
...
-        for deletion in self._get_deletions(draft, resources):
-            await self._apply_change(draft, deletion)
-
-        draft = await self.depositor.get_deposition_by_id(draft.id_)
-        new_datapackage, old_datapackage = await self._update_datapackage(
-            original=original, draft=draft, resources=resources
-        )
-        run_summary = self._summarize_run(
-            old_datapackage, new_datapackage, resources, draft.links.html
+async def orchestrate_run(
```
zschira, line 14: The orchestrator has been majorly simplified to just this top-level function, and most of the logic was moved to the `DraftDeposition` and the `ZenodoDepositorInterface` for zenodo specific stuff.

src/pudl_archiver/depositors/depositor.py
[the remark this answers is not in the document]
jdangerx: I think this is a good insight and we are close to getting it right - details in the overall review comment!

src/pudl_archiver/depositors/depositor.py
```diff
@@ -0,0 +1,336 @@
...
+        """Actually close the file for internal use."""
+        super().close()
+
+
+class DepositionAction(Enum):
+    """Enumerate types of changes which can be applied to deposition files."""
+
+    CREATE = (auto(),)
+    UPDATE = (auto(),)
+    DELETE = (auto(),)
```
jdangerx, line 52: If we allow the action to be `NO_OP` also, we can stop using `| None` in the type signatures 🙂

src/pudl_archiver/depositors/depositor.py
jdangerx: This is pretty slick!

src/pudl_archiver/depositors/zenodo/depositor.py
jdangerx, line 1: I'm assuming this was largely a lift-and-shift from `src/pudl_archiver/depositors/zenodo.py`, let me know if you want me to go through this with a fine-toothed comb. But I skimmed it and my sense is "yeah, there's a lot of Zenodo ugliness that we've successfully hidden from the rest of the code in here!"

jdangerx (changes requested):
I think the idea of "Draft deposition should only expose the *high-level* operations, and we should implement different backends that plug into this interface" is really good and provides us a lot of freedom to mess with the backend implementations. It also pushes a lot of backend-specific logic out of the orchestrator, which is great!

However, in this code we don't *quite* express that idea. Instead, we've created an interface at a lower level for the backends. Then we take this lower-level interface and string it together into the high level operations in `DraftDeposition`... which opens the door to something like:

* try using a new backend - implement all the `AbstractDepositionInterface` operations for it
* but there's some sort of weird side effect or order dependency that only exists in this backend, which means the logic in `DraftDeposition` has to change to account for that side effect.
* but `DraftDeposition` is also shared by *all* backends, so any changes to `DraftDeposition` need to play well with the various backend implementations.

In short, the implementations of `AbstractDepositionInterface`s have to know about the implementation of `DraftDeposition` itself, which is hidden coupling, which is confusing & causes code changes to turn out harder than expected.

I would probably reduce this risk by defining the interface at the `DraftDeposition` level - defining the `latest_for_dataset`, `add_resource`, `attach_datapackage`, `cleanup_after_error`, `publish`, `get_file`, `list_files`, `get_deposition_link` methods in some sort of `AbstractDraftDeposition`. Then we'd have `ZenodoDraftDeposition`, `GCSDraftDeposition`, `R2DraftDeposition`, etc. - all exposing the same high-level operations that the orchestrator can happily use.

This hews closer to what I think is your original idea, and gives us the flexibility we are looking for out of this PR without introducing additional coupling.

I *think* this is actually a pretty small change, since there's actually not *that* much logic required to glue the low-level operations (which correspond to the **Zenodo/GCS/R2 API design**) together into our high-level operations (which correspond to the **steps the orchestrator actually wants to take**).

An additional nice thing about this is that it deals with a lot of the confusing naming stuff we see in this code. `DraftDeposition.deposition` is an instance of `AbstractDepositorInterface` - but then `AbstractDepositorInterface` has an `open_draft` method... that all gets a little clearer when we only have the one type of thing that represents a draft deposition.

src/pudl_archiver/cli.py
```diff
@@ -82,10 +83,13 @@ def main():
     log_format = "%(asctime)s [%(levelname)8s] %(name)s:%(lineno)s %(message)s"
     coloredlogs.install(fmt=log_format, level=logging.INFO, logger=logger)
     args = parse_main()
+    datasets = args.datasets
     if args.all:
-        args.datasets = ARCHIVERS.keys()
+        datasets = ARCHIVERS.keys()
     del args.all
```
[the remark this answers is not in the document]
zschira: This is because all of the other args are passed to a `RunSettings` object as kwargs, so this just avoids having an extra arg.

src/pudl_archiver/depositors/depositor.py
```diff
@@ -0,0 +1,336 @@
...
+from pydantic import BaseModel, ConfigDict, PrivateAttr
+
+from pudl_archiver.archivers.validate import RunSummary
+from pudl_archiver.frictionless import DataPackage, ResourceInfo
+from pudl_archiver.utils import RunSettings, Url
+
+logger = logging.getLogger(f"catalystcoop.{__name__}")
+
+
+class _UploadSpec(BaseModel):
```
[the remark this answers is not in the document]
zschira: Changed to a `dataclass`

src/pudl_archiver/depositors/depositor.py
zschira: Good call, changed

src/pudl_archiver/depositors/zenodo/depositor.py
zschira: Yeah there's some reorg going on in here, but not really any substantive changes

zschira:
I took another crack at this and reorganized into `PublishedDeposition` and `DraftDeposition` base classes, and there are methods to go from a published deposition from a draft and vice-versa. I think this sort of encapsulates the main states of the depositor, which is basically that a deposition can be published and read only, or in a draft state and readable/writeable. The draft deposition will also return a new instance of its self every time you make an edit, which should update any state it maintains about its contents.

src/pudl_archiver/depositors/zenodo/depositor.py
```diff
@@ -0,0 +1,609 @@
...
+
+        if not self.settings.initialize:
+            self = self.model_copy(
+                update={"deposition": await self._get_deposition(self.doi)}
+            )
+        return self
+
+    async def open_draft(self) -> "ZenodoDraftDeposition":
+        """Open a new draft deposition to make edits."""
+        return await ZenodoDraftDeposition.new_draft(self)
```
jdangerx, line 308: I think it might be nice to have the "create a new draft from a published dep" logic live in the published dep, something like

```python
def open_draft(self):
    new_record_info = self.api_client.create_new_record(...)
    return ZenodoDraftDeposition(doi=..., api_client=...) # which then doesn't have to have any custom init stuff
```

jdangerx (changes requested):
I like the conceptual hierarchy we have now of "actual states *we* care about (`PublishedDeposition`, `DraftDeposition`)" vs. "interacting with Zenodo." Now if we want to make `GCSPublishedDeposition` and `GCSDraftDeposition`, I think we could actually do that without having to redo the interfaces. So that's a big win! I've a few smaller restructures to propose: adding an abstract unitialized state & using composition instead of inheritance in Zenodo-land.

I think it would be nice to include an "uninitialized" abstract state. Currently our `published -> draft` transition logic has a bit of extra logic in it to also handle the "what if `published` is actually not a published deposition at all, but one that doesn't exist yet?" case. Including a new state lets us have one function handle one state transition and avoids overloading the "published" state.

Incidentally, it is a little awkward that said `published -> draft` transition logic lives in `ZenodoDraftDeposition.new_draft(published_deposition)` - typically the transitions are methods on the source state, not the destination state.

Finally, the three Zenodo classes are still pretty tightly coupled in a way that will make it difficult to change when we migrate to the New Zenodo API. I think we can shove all the lower-level Zenodo interaction into one Zenodo API client class and have the higher-level Zenodo*Deposition classes hold references to an API client.

This would make it so that the higher-level classes don't expose lower-level operations, and the logic in those classes will operate at the same level of abstraction. Then, when the new Zenodo API is released, we can:
* write a new API client class for the new API
* inject *both* API clients to the high-level class
* continue to build the Deposition-level operations from API-level operations, phasing out the legacy API stuff but retaining the flexibility to use legacy API operations in the likely case that the new stuff is broken in some way.

src/pudl_archiver/__init__.py
```diff
@@ -94,22 +80,24 @@ async def on_request_end(session, trace_config_ctx, params):
...
-                dataset_settings_path=Path("dataset_doi.yaml"),
-                create_new=initialize,
-                dry_run=dry_run,
-                sandbox=sandbox,
-                auto_publish=auto_publish,
-                refresh_metadata=refresh_metadata,
+                run_settings.only_years,
+                download_directory=run_settings.download_dir,
+            )
+            depositor = await ZenodoPublishedDeposition.get_latest_version(
```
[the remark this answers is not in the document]
zschira: That's a good point, the uninitialized state was kind of being hidden here. I still decided not to add a class for this state, given that it doesn't really have any meaningful "state", and all it does is go to draft state. I felt like adding another base class would just lead to clutter, so instead I added a level function called `get_deposition` that acts as an entry point into the state machine. This function either goes `published->draft` or creates a `draft` from scratch, but it will always return a draft that you can start working on.

src/pudl_archiver/depositors/zenodo/depositor.py
```diff
@@ -0,0 +1,609 @@
...
+
+
+class DatasetSettings(BaseModel):
+    """Simple model to validate doi's in settings."""
+
+    production_doi: Doi | None = None
+    sandbox_doi: SandboxDoi | None = None
+
+
+class ZenodoBaseDepositionInterface(BaseModel):
```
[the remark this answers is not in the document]
zschira: Yeah I used inheritance here because I wanted to avoid some boiler plate, but ultimately this approach felt kind of tricky/error prone, so I'm good with using composition instead.

src/pudl_archiver/depositors/zenodo/depositor.py
zschira: Agreed, changed!

src/pudl_archiver/__init__.py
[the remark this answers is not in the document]
jdangerx: non-blocking, just musing: I guess the uninitialized state is actually sort of represented by the `DepositionInterface` wrapper class in this architecture - and then `get_deposition` actually represents the uninitialized -> draft transition.

src/pudl_archiver/depositors/depositor.py
```diff
@@ -0,0 +1,301 @@
...
+@dataclass
+class DepositionChange:
+    """Define a single change to a file in a deposition."""
+
+    action_type: DepositionAction
+    name: str
+    resource: io.IOBase | Path | None = None
+
+
+class PublishedDeposition(BaseModel, ABC):
```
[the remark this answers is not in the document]
jdangerx: I think now that we're using composition, both ways of instantiating the class leave us with usable class instances, which is the important bit :)

If you wanted to get rid of the class methods, though, we could move the "API client initialization" logic into the caller, something like this. Definitely not something we need to do, but see if this structure feels better to you 🤷 

```diff
diff --git a/src/pudl_archiver/depositors/__init__.py b/src/pudl_archiver/depositors/__init__.py
index 818cda2..7cb719b 100644
--- a/src/pudl_archiver/depositors/__init__.py
+++ b/src/pudl_archiver/depositors/__init__.py
@@ -44,16 +44,27 @@ async def get_deposition(
     dataset: str, session: aiohttp.ClientSession, run_settings: RunSettings
 ) -> tuple[DraftDeposition, DataPackage | None]:
     """Create draft deposition from scratch or previous version."""
-    deposition_interface = DEPOSITION_INTERFACES[run_settings.depositor]
+    deposition_backend = DEPOSITION_BACKENDS[
+        run_settings.depositor
+    ]  # renamed to better reflect the bundle of "zenodoeyness", maybe?
+    api_client = await deposition_backend.api_client.initialize_client(
+        dataset, session, run_settings.sandbox
+    )
     if run_settings.initialize:
-        return await deposition_interface.draft_interface.initialize_from_scratch(
-            dataset, session, run_settings
+        deposition = await api_client.create_new_deposition()
+        return await deposition_backend.draft_interface(
+            dataset_id=dataset,
+            settings=run_settings,
+            api_client=api_client,
+            deposition=deposition,
         ), None
+    deposition = await api_client.get_deposition(api_client.doi)
 
-    published_deposition = (
-        await deposition_interface.published_interface.get_latest_version(
-            dataset, session, run_settings
-        )
+    published_deposition = deposition_backend.published_interface(
+        dataset_id=dataset,
+        settings=run_settings,
+        api_client=api_client,
+        deposition=deposition,
     )
 
     original_datapackage_bytes = await published_deposition.get_file("datapackage.json")
diff --git a/src/pudl_archiver/depositors/zenodo/depositor.py b/src/pudl_archiver/depositors/zenodo/depositor.py
index 30256d9..2944c88 100644
--- a/src/pudl_archiver/depositors/zenodo/depositor.py
+++ b/src/pudl_archiver/depositors/zenodo/depositor.py
@@ -539,32 +539,6 @@ class ZenodoPublishedDeposition(PublishedDeposition):
     api_client: ZenodoAPIClient
     dataset_id: str
 
-    @classmethod
-    async def get_latest_version(
-        cls,
-        dataset: str,
-        session: aiohttp.ClientSession,
-        run_settings: RunSettings,
-    ) -> "ZenodoPublishedDeposition":
-        """Create a new ZenodoDepositor.
-
-        Args:
-            dataset: Name of dataset to archive.
-            session: HTTP handler - we don't use it directly, it's wrapped in self.api_client._request.
-            run_settings: Settings from CLI.
-        """
-        api_client = await ZenodoAPIClient.initialize_client(
-            dataset, session, run_settings.sandbox
-        )
-        deposition = await api_client.get_deposition(api_client.doi)
-
-        return cls(
-            dataset_id=dataset,
-            settings=run_settings,
-            api_client=api_client,
-            deposition=deposition,
-        )
-
     async def open_draft(self) -> "ZenodoDraftDeposition":
         """Open a new draft deposition to make edits."""
         draft_deposition = await self.api_client.get_new_version(
@@ -735,22 +709,5 @@ class ZenodoDraftDeposition(DraftDeposition):
             f"Failed while creating new deposition: {traceback.print_exception(e)}"
         )
 
-    @classmethod
-    async def initialize_from_scratch(
-        cls, dataset: str, session: aiohttp.ClientSession, run_settings: RunSettings
-    ) -> "ZenodoDraftDeposition":
-        """Create a brand new draft deposition."""
-        api_client = await ZenodoAPIClient.initialize_client(
-            dataset, session, run_settings.sandbox
-        )
-        deposition = await api_client.create_new_deposition()
-
-        return cls(
-            dataset_id=dataset,
-            settings=run_settings,
-            api_client=api_client,
-            deposition=deposition,
-        )
-
 
 register_depositor("zenodo", ZenodoPublishedDeposition, ZenodoDraftDeposition)
```

jdangerx (changes requested):
Sweet, this looks really good! No blocking architectural notes, a few things to consider & a few small hygienic improvements remain.

.github/workflows/tox-pytest.yml
```diff
@@ -22,6 +22,7 @@ jobs:
         with:
           environment-file: environment.yml
           cache-environment: true
+          cache-environment-key: environment-${{ hashFiles('pyproject.toml') }}
```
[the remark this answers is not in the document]
zschira: There was a GDAL conflict with PUDL that I fixed by adding an explicit GDAL requirement, but the CI was still failing because it was using the cached environment, so this will force it to recreate when `pyproject.toml` changes. Also, I've now removed the GDAL thing because Zane fixed it on the PUDL side.

src/pudl_archiver/depositors/depositor.py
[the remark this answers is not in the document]
zschira: Yeah I like removing the class methods so we don't have to implement them everytime! I also removed the `dataset_id` as a field on the zenodo api client , while I was creating an abstract interface for an api client, so this is just passed in where needed, and the api client itself truly maintains no state.

src/pudl_archiver/depositors/__init__.py
```diff
@@ -3,4 +3,60 @@
...
+from dataclasses import dataclass
+
+import aiohttp
+
+from pudl_archiver.frictionless import DataPackage
+from pudl_archiver.utils import RunSettings
+
+from .depositor import (
+    DepositionAction,
+    DepositionChange,
```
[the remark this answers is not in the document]
zschira: I decided to import them here so they're available to import elsewhere from `pudl_archiver.depositors` instead of `pudl_archiver.depositors.depositor`.

src/pudl_archiver/cli.py
```diff
@@ -73,21 +74,29 @@ def parse_main():
...
 
-def main():
+async def archiver_entry(args=None):
     """Run desired archivers."""
     load_dotenv()
     logger.setLevel(logging.INFO)
     log_format = "%(asctime)s [%(levelname)8s] %(name)s:%(lineno)s %(message)s"
     coloredlogs.install(fmt=log_format, level=logging.INFO, logger=logger)
-    args = parse_main()
+    args = parse_main(args)
```
jdangerx, line 91: Nit - it might be cleaner to not bother at all with arg parsing within this method, and just have it take various args as inputs. Then you get:

```python
def parse_main(): # no need to pass in an args object anymore, this always reads from sys.argv
    ...

async def archiver_entry(args:argparse.Namespace):
    ...
    datasets = args.datasets # just use the args object that was passed in

def main():
    args = parse_main()
    asyncio.run(archiver_entry(args))
```

And in test, where you use `archiver_entry(some_special_test_args)` that should all still work. This avoids passing the desired override args all the way into `parse_main()` just to have that pass things back out again.

jdangerx (approved):
Hooray for the new test and it doesn't look _so_ horrific to my eye :)

I think there's a little bit of extraneous argument passing complexity in the CLI which you can get rid of, but I'm also happy to merge everything as is.

src/pudl_archiver/cli.py
zschira: In this case I'm not sure how to parse the args from the test context before passing in to `archiver_entry`. I'm going to go ahead and merge for now since I'm leaving soon, but happy to change this later if there's something I'm missing.

PR #302, opened 2024-03-19
Archive EIA 930 data

### LINKED ISSUE #295: Archive bulk EIA930 data
* EIA Form 930 contains hourly data reported by various US balancing authorities and sub-BA regions.
* It is available for download as bulk CSVs, each of which covers 6 months of data
* The data starts in the 2nd half of 2015 and runs through 2024.
* The data is updated in close to real time.
* The [EIA Hourly Electric Grid Monitor](https://www.eia.gov/electricity/gridmonitor/dashboard/electric_overview/US48/US48) dashboard provides a pop-up with the download URLs.
* Prior to the 2nd half of 2018, the `SUBREGION` files do not exist.

Recent bulk download URLs take the following form:

* https://www.eia.gov/electricity/gridmonitor/sixMonthFiles/EIA930_BALANCE_2023_Jul_Dec.csv
* https://www.eia.gov/electricity/gridmonitor/sixMonthFiles/EIA930_BALANCE_2023_Jan_Jun.csv
* https://www.eia.gov/electricity/gridmonitor/sixMonthFiles/EIA930_INTERCHANGE_2023_Jul_Dec.csv
* https://www.eia.gov/electricity/gridmonitor/sixMonthFiles/EIA930_INTERCHANGE_2023_Jan_Jun.csv
* https://www.eia.gov/electricity/gridmonitor/sixMonthFiles/EIA930_SUBREGION_2023_Jul_Dec.csv
* https://www.eia.gov/electricity/gridmonitor/sixMonthFiles/EIA930_SUBREGION_2023_Jan_Jun.csv

To keep the number of archived files from approaching the Zenodo limits, we might archive these as zipfiles for half years, e.g. `eia930-2023h1.zip` and `eia930-2023h2.zip` by analogy with our quarterly filename formats.

e-belfer:
# Overview

Closes #295 

Archives raw EIA 930 data.

- Zips files into one zipfile per half-year, drawing on the file list provided by EIA. 
- Adds a CSV validation method. 
- Adds DOIs to `dataset.doi`
- Adds archive to GHA
- Switch from `2018h1` to `2018half1` for partition to avoid confusion with "hours" shorthand

# Testing

Review [sandbox archive](https://sandbox.zenodo.org/records/38410) and [draft production archive](https://zenodo.org/records/10840078?preview=1), attempt to make archive.

# To-do list

```[tasklist]
- [x] Get feedback on CSV validation method
- [x] Make production archive and add to `dataset_doi.yaml`
```

### DIFF
--- a/src/pudl_archiver/archivers/validate.py
+++ b/src/pudl_archiver/archivers/validate.py
@@ -1,6 +1,5 @@
 """Defines models used for validating/summarizing an archiver run."""
 
-import csv
 import logging
 import xml.etree.ElementTree as Et  # nosec: B405
 import zipfile
@@ -8,6 +7,7 @@ from io import BytesIO
 from pathlib import Path
 from typing import Any, Literal
 
+import pandas as pd
 import pyarrow as pa
 from pydantic import BaseModel
 
@@ -302,18 +302,13 @@ def _validate_xml(buffer: BytesIO) -> bool:
         Et.parse(buffer)  # noqa: S314
     except Et.ParseError:
         return False
-
     return True
 
 
 def _validate_csv(buffer: BytesIO) -> bool:
     try:
-        # pd.read_csv(buffer, nrows=100) # Alternate option and maybe preferable?
-        start = buffer.read(4096)
-        dialect = csv.Sniffer().sniff(str(start))
-        if dialect.delimiter == ",":
-            return True
+        sliver = pd.read_csv(buffer, nrows=100)  # Try reading in a data slice
+        return not sliver.empty
+    except (pd.errors.EmptyDataError, pd.errors.ParserError):
         return False
-    except csv.Error:
-        # Could not get a csv dialect -> probably not a csv.
-        return False
+    return True
--- a/src/pudl_archiver/archivers/validate.py
+++ b/src/pudl_archiver/archivers/validate.py
@@ -1,5 +1,6 @@
 """Defines models used for validating/summarizing an archiver run."""
 
+import csv
 import logging
 import xml.etree.ElementTree as Et  # nosec: B405
 import zipfile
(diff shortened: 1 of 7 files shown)


zschira:
The only thing that comes to mind for CSV validation is just to try and read the file and see if it works. We could make a function that takes some input like the expected delimiter, and if there should be any header rows to skip, then just tries to read the file and reports whether or not it encountered any errors.

We could theoretically use Pandera to check that CSV's conform to an expected schema, but this might be overkill at the archiving level.

src/pudl_archiver/archivers/validate.py
```diff
@@ -292,3 +304,16 @@ def _validate_xml(buffer: BytesIO) -> bool:
         return False
 
     return True
+
+
+def _validate_csv(buffer: BytesIO) -> bool:
```
e-belfer, line 309: @zschira This is currently saying, is the csv actually a comma-separated value? But it doesn't give us any information about whether the actual internals of the file are corrupted. Using `pd.read_csv()` on a small slice seems more robust and it sounds like from what you're saying in the issue you agree?

e-belfer: Updated to just read the first 100 lines of the CSV. Will run all existing archives against this to make sure skiprows isn't needed, I don't remember this being an issue with our existing CSVs.

cmgosnell (approved):
this looks good to me! appreciate the last minute `h` -> `half` change!

src/pudl_archiver/archivers/validate.py
zaneselvans: One system for validating the parameters of a CSV, but not its schema is the [CSV Dialect](https://specs.frictionlessdata.io/csv-dialect/) defined in the Frictionless specs. That seems like the appropriate level of validation to be doing here, with the schema information probably only coming in during the transform step somewhere.

e-belfer: This is closer to what I had before (using `csv.Sniffer()`) but it only tells us that this is an actual comma-separated string, not whether it's empty or has indecipherable characters in the header. I think trying to open the first 100 rows of the file seems just as or more useful at this stage, and will catch different types of errors so I'm inclined to stick with it for now. I'm also almost at budget for this step.

PR #303, opened 2024-03-20
Stop generating deprecated `remote_url` metadata field.

zaneselvans:
# Overview

We no longer use the `remote_url` field in our archives, and instead leave the `path` field pristine and infer the local paths, so we should stop inserting the cruft into new archives.

# Testing

- I removed the `remote_url` field from my handmade `gridpathratk` archive and it worked fine.

# To-do list

```[tasklist]
```

### DIFF
--- a/src/pudl_archiver/frictionless.py
+++ b/src/pudl_archiver/frictionless.py
@@ -81,7 +81,6 @@ class Resource(BaseModel):
     profile: str = "data-resource"
     name: str
     path: Url
-    remote_url: Url
     title: str
     parts: dict[str, Any]
     encoding: str = "utf-8"
@@ -104,7 +103,6 @@ class Resource(BaseModel):
         return cls(
             name=file.filename,
             path=file.links.canonical,
-            remote_url=file.links.canonical,
             title=filename.name,
             mediatype=mt,
             parts=parts,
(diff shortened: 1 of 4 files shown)


e-belfer:
Have you tested this on already existing archives, @zaneselvans

zaneselvans:
@e-belfer What do you mean by testing it on existing archives?  Like have I tried creating a new version of an existing archive and then downloaded it?

zaneselvans:
I was able to create [a new sandbox version](https://sandbox.zenodo.org/records/13549) of the `eia860` archive using this branch. The first archive failed the new archive validation step, but didn't give an informative error as to why.  The failure appeared to happen when it was comparing the new and old `datapackage.json` files.

I approved that draft, and tried to create another new version, and it succeeded without complaint.

My hunch is that when it was comparing the old and new `datapackage.json` files, it detected a difference between them because the  `remote_url` fields were not present in the new one, and this difference was incongruous with the fact that the data itself had not changed.

So it seems like we might get a round of "failed" archives which are actually fine, but which we'll need to check manually and approve to get beyond having the `remote_url`.

e-belfer (approved):
You answered my question - just wanted to know what the consequences of this deprecation would be, sounds like we'll just expect to do a bit of additional manual validation on April Fool's day.

PR #309, opened 2024-03-25
Update to use Python 3.12

zaneselvans:
# Overview

Update the PUDL archivers to use Python 3.12, so they are compatible with the main PUDL package.

Currently one failing test.... not sure why!
```console
pytest tests/unit/archive_base_test.py::test_download_file
```

```py
    @pytest.mark.asyncio
    async def test_download_file(mocker, tmp_path):
        """Test download_file.

        Tests that expected data is written to file on disk or in memory. We use Catalyst's
        README.md file to ensure that downloads produce the same content as a direct GET.
        """

        def _ground_truth_download(url):
            return requests.get(url, timeout=1000).content

        url = "https://github.com/catalyst-cooperative/pudl-archiver/blob/main/README.md"
        file_content = _ground_truth_download(url)

        # Initialize MockArchiver class
        archiver = MockArchiver(None)

        async with ClientSession() as session:
            archiver.session = session
            file = io.BytesIO()

            # Call method
            await archiver.download_file(url, file)
>           assert file.getvalue() == file_content
```
```
E           assert b'\n\n\n\n\n\...\n</html>\n\n' == b'\n\n\n\n\n\...\n</html>\n\n'
E
E             At index 16350 diff: b'A' != b'2'
E
E             Full diff:
E               (b'\n\n\n\n\n\n<!DOCTYPE html>\n<html\n  lang="en"\n  \n  data-color-mode="au'
E                b'to" data-light-theme="light" data-dark-theme="dark"\n  data-a11y-animated'
E                b'-images="system" data-a11y-link-underlines="true"\n  >\n\n\n\n\n  <head>\n '...
E
E             ...Full output truncated (5865 lines hidden), use '-vv' to show
```

# To-do list

```[tasklist]
```

### DIFF
--- a/tests/unit/archive_base_test.py
+++ b/tests/unit/archive_base_test.py
@@ -207,7 +207,7 @@ async def test_download_file(mocker, tmp_path):
     def _ground_truth_download(url):
         return requests.get(url, timeout=1000).content
 
-    url = "https://github.com/catalyst-cooperative/pudl-archiver/blob/main/README.md"
+    url = "https://raw.githubusercontent.com/catalyst-cooperative/pudl-archiver/main/README.md"
     file_content = _ground_truth_download(url)
 
     # Initialize MockArchiver class
--- a/environment.yml
+++ b/environment.yml
@@ -5,8 +5,8 @@ channels:
   - defaults
 dependencies:
   # Used to set up the environment
-  - pip>=21,<24
-  - python>=3.11,<3.12
+  - pip>=21,<25
+  - python>=3.12,<3.13
   - setuptools>=66
   # fiona is a transitive dependency which needs GDAL. so we install with conda
   # TODO: we shouldn't have to install all this geo stuff, so once we break the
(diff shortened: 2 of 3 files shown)


tests/unit/archive_base_test.py
```diff
@@ -207,7 +207,7 @@ async def test_download_file(mocker, tmp_path):
     def _ground_truth_download(url):
         return requests.get(url, timeout=1000).content
 
-    url = "https://github.com/catalyst-cooperative/pudl-archiver/blob/main/README.md"
+    url = "https://raw.githubusercontent.com/catalyst-cooperative/pudl-archiver/main/README.md"
```
zaneselvans, line 210: I'm not sure why the behavior on the old URL changed, but it seemed simpler to depend on the Markdown file itself rather than 6000 lines of HTML being dynamically generated by GitHub, and the test passes with the Markdown file so... 🤷🏼

e-belfer:
@zaneselvans I've been finding that that test periodically fails, and it's stochastic - typically resolves if you just rerun it. Seems like pulling the raw text might help resolve the issue, likely some kind of interaction with Zenodo is my guess.

jdangerx (approved):
Let 'er rip!

zaneselvans:
@e-belfer It was failing for me very reliably on the old URL, and succeeding very reliably on the new one but I have absolutely no idea what would have changed.

PR #313, opened 2024-04-01
Update issue templates for monthly dataset review

e-belfer:
Make monthly archive updates faster by creating a template for review. The next step will be to connect this to the GHA so this issue is automatically created after each monthly run.

See #314 for an example of how this looks as a rendered issue.


zaneselvans (approved):
Great!

PR #325, opened 2024-04-17
Fix auto publish and ferc 6/714 archivers

zschira:
This PR cleans up some minor bugs introduced in recent changes. It fixes auto-publishing behavior, which was publishing even when disabled, and fixes ferc 6/714 archivers, which were calling the `archive_taxonomy` function slightly incorrectly.

### DIFF
--- a/src/pudl_archiver/depositors/depositor.py
+++ b/src/pudl_archiver/depositors/depositor.py
@@ -269,6 +269,7 @@ class DraftDeposition(BaseModel, ABC):
             )
             return None
 
+        logger.info("Attempting to publish deposition.")
         return await self.publish()
 
     async def _apply_change(self, change: DepositionChange) -> "DraftDeposition":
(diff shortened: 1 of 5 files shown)


e-belfer (approved):
This fixed the issue I was having for both archives, and made a draft as expected. Thanks for the quick fix. One non-blocking request for the logging statements.

src/pudl_archiver/depositors/depositor.py
```diff
@@ -246,7 +246,7 @@ async def add_resource(
         return await self._apply_change(change)
 
     async def publish_if_valid(
-        self, run_summary: RunSummary, datapackage_updated: bool
+        self, run_summary: RunSummary, datapackage_updated: bool, auto_publish: bool
     ) -> PublishedDeposition | None:
         """Check that deposition is valid and worth changing, then publish if so."""
         logger.info("Attempting to publish deposition.")
```
e-belfer, line 252: Can we move this down to just above the last return statement? Otherwise we state we're attempting to publish a deposition before it meets our conditions for publication.

Issue #338, opened 2024-05-01
ferc1, ferc6, ferc60, ferc714 logon request error


zschira:
Several ferc archivers are receiving an error about getting a logon request when trying to download taxonomy files:

> Web file appears to be an html logon request, not retrieved: https://ecollection.ferc.gov/taxonomy/form1/2021-01-01/schedule/AccumulatedDeferredIncomeTaxes/form-1-sched-AccumulatedDeferredIncomeTaxes_2021-01-01_cal.xml

When I follow the link, I'm able to download the file without a problem, which leads me to believe this is some sort of rate limiting issue. I would guess that just reducing the concurrency of these archivers will solve this, but I also wonder if there's some sort of credentials we can get to logon and avoid these limits?

Issue #336, opened 2024-05-01
Publish archives for the month of 4/2024


zschira:
# Archiver run results

Most errors were acceptable validation errors (new years of data which have grown by >25% in size since last run). There seems to be some FERC rate limiting? error, plus nrelatb and phmsagas have legitimate errors that need investigating.

```[tasklist]
- [x] eia176 - Unchanged
- [x] eia191 - Published
- [x] eia757a - Unchanged
- [x] eia860 - Published
- [X] eia860m - Failed because eia860m-2024.zip increased in size by > 25%. This seems expected, so I've manually published.
- [x] eia861 - Unchanged
- [X] eia923 - Failed because eia923-2024.zip increased in size by > 25%. This seems expected, so I've manually published.
- [X] eia930 - Failed because eia930-2024half1.zip increased in size by > 25%. This seems expected, so I've manually published.
- [x] eiaaeo - Unchanged
- [x] eiawater - Unchanged
- [x] eia_bulk_elec - Published
- [x] epacamd_eia - Published
- [x] ferc1 - logon request error
- [x] ferc2 - Failed because ferc2-xbrl-2023.zip increased in size by > 25%. This seems expected, so I've manually published.
- [x] ferc6 - Same as ferc1
- [x] ferc60 - Same as ferc1
- [x] ferc714 - Same as ferc1
- [x] mshamines
- [x] nrelatb - When trying to get the latest version of the deposition, zenodo is returning an error with the message: "The record has been deleted"
- [x] phmsagas - Validation failure
- [x] epacems - Published
```

# Validation failures
For each run that failed because of validation test failures (seen in the GHA logs), add it to the tasklist. Download the run summary JSON by going into the "Upload run summaries" tab of the GHA run for each dataset, and follow the link. Investigate the validation failure.

If the validation failure is deemed ok after manual review (e.g., Q2 of 2024 data doubles the size of a file that only had Q1 data previously, but the new data looks as expected), go ahead and approve the archive and leave a note explaining your decision in the task list.

If the validation failure is blocking (e.g., file format incorrect, whole dataset changes size by 200%), make an issue to resolve it.

```[tasklist]
- [ ] phmsagas validation failure
```

# Other failures
For each run that failed because of another reason (e.g., underlying data changes, code failures), create an issue describing the failure and take necessary steps to resolve it.

```[tasklist]
- [ ] ferc1, ferc6, ferc60, ferc714 logon request error
[...]

PR #340, opened 2024-05-03
Fix datapackage errors and FERC failures

zschira:
This PR addresses issues discovered while trying to publish new archives outlined in [this](https://github.com/catalyst-cooperative/pudl-archiver/issues/336) issue. Specifically, it addresses problems with the `datapackage.json` files generated during this run, and problems with the FERC archivers. The datapackage issues arose fromrecent updates that caused us to accidentally overwrite values in the file causing there to be bad links for resources within the archive. The [FERC issue](https://github.com/catalyst-cooperative/pudl-archiver/issues/338) came up when we started occasionally getting errors where we would get a webpage that appeared to be a logon while trying to archive taxonomy files. This appears to be a rate limiting mechanism, and by reducing concurrency, we were able to avoid this issue in [this](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/8944169648/job/24570507508) archiver run.

### DIFF
--- a/src/pudl_archiver/depositors/depositor.py
+++ b/src/pudl_archiver/depositors/depositor.py
@@ -309,6 +309,10 @@ class DraftDeposition(BaseModel, ABC):
     def _datapackage_worth_changing(
         self, old_datapackage: DataPackage | None, new_datapackage: DataPackage
     ) -> bool:
+        # Copy datapackages so we can modify without causing problems down the line
+        new_datapackage = new_datapackage.model_copy(deep=True)
+        if old_datapackage is not None:
+            old_datapackage = old_datapackage.model_copy(deep=True)
         # ignore differences in created/version
         # ignore differences resource paths if it's just some ID number changing...
         if old_datapackage is None:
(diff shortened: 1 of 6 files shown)


src/pudl_archiver/depositors/depositor.py
```diff
@@ -309,19 +309,30 @@ async def _upload_file(self, upload: _UploadSpec):
...
                     r.path = re.sub(r"/\d+/", "/ID_NUMBER/", str(r.path))
-            if getattr(new_datapackage, field) != getattr(old_datapackage, field):
+            if getattr(new_datapackage_copy, field) != getattr(
+                old_datapackage_copy, field
+            ):
                 return True
-        return False
+
+        # Due to a bug in an earlier version some resources ended up with ID_NUMBER in path.
+        # These should be replaced
```
jdangerx, line 334: Right, they wouldn't be replaced without this line because we'd have `getattr(old_datapackage_copy, field) == "the string with ID_NUMBER in it anyways"`. Makes sense!

jdangerx (approved):
Making a copy makes sense! And glad you found a nice way to update the broken ones too.

zaneselvans:
Do we know how the bad archives with `ID_NUMBER` in them got published? I thought we had auto-publish disabled for the production server.

src/pudl_archiver/depositors/depositor.py
zschira: Yeah although I suppose this could be deleted now that they should all be fixed

zschira:
@zaneselvans that's on me, I published them and didn't notice the `ID_NUMBER` although I definitely should have checked for it given that we knew about this issue.

zaneselvans:
Do we understand where this intermittent

```
b'' != b'some other non-empty string' 
```

assertion failure in the unit tests is originating?

zschira:
Hmm, I hadn't seen this before, but it's happening in our big zenodo sandbox test where we go through a series of archiver runs with fake files and verify that the deposition looks as expected at each stage. For some reason, it looks like when we try to download files to verify their contents after a run, sometimes we're getting an empty string for reasons I don't understand. I'll create an issue.

Issue #346, opened 2024-05-15
Automatically create archive approval checklist


jdangerx:
### Overview

Filling out the archive approval checklist issue is tedious and seems automatable. Let's prioritize a small number of changes to make this quicker and less painful.

### Success Criteria
* [x] monthly archiving action automatically creates a new archive approval checklist issue
* [ ] monthly archive action populates the issue based on observed failures (including non-validation failures). stretch: the template is re-formatted based on the information (e.g., failed runs go into the failed run section). lower-hanging fruit goal is to add the updated slack template information as a comment on the issue.
* [ ] monthly archive action includes the highest priority information in the issue, to avoid having to open a million run summary JSONs. This includes: file changes (create/update/delete), size of file change in MB and percentage, change in partitions, and any validation failures.

### Current challenges
- We don't see any run failures in our existing automated reporting
- We only report run summaries report to Slack, not GH
- We don't show the size of changes in any of our reports, which makes assessing the scale of changes difficult.

PR #357, opened 2024-06-17
MVP improvements to automated archive runs

### LINKED ISSUE #347: Better reporting & notification of archive creation and validation failures
### Overview/Success criteria
* [ ] on archiver failure, enumerate all the failures without needing someone to re-run to debug them

```[tasklist]
### Next steps
* [ ] figure out where we want to enumerate these failures
```

e-belfer:
# Overview

Addresses subset of #346 and #347, focusing on the lowest hanging improvements.

There are two types of GHA failures we want to keep track of: run failures and validation test failures. These should both be reported in the Slackbot notifications. The archive run now also automatically opens an issue using the archive issue template, noting the date of the run and linking to the URL for the action. Finally, we can now easily configure what datasets we run on a manual workflow dispatch, whether to kick off the large runner for EPACEMS (and eventually maybe some other datasets), and whether to make a Github issue following the run.

- [x] Update the `make_slack_notification.py` to incorporate validation test failures.
- [x] Auto-create Github issue monthly with link to run results and date
- [x] Add options for manual dispatch: to only run workflow on certain datasets, kick off large runner, make Github issue

Out of scope:
- Figure out how to add other run failures to Slack message with enough info from the error to avoid having to luck at the run logs (e.g., "archive deleted")
- Figure out how to add run summary to Github issue (e.g., these datasets passed, these failed validation, these are differently broken)
- Figure out how to auto-populate tasklist with this run summary

# Testing

Action testing one dataset manually provided, Github issue creation and skipping large runners:
https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/9569277213
https://github.com/catalyst-cooperative/pudl-archiver/issues/360

Action testing large runners, 2 datasets manually provided: 
https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/9568539141

Action testing skipping of Github issue creation: 
https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/9569300760

# To-do list

```[tasklist]
- [ ] Squash commits
```

### DIFF
--- a/.github/ISSUE_TEMPLATE/monthly-archive-update.md
+++ b/.github/ISSUE_TEMPLATE/monthly-archive-update.md
@@ -8,7 +8,7 @@ assignees: e-belfer
 ---
 
 # Summary of results:
-See the job run results [here]({{ env.RUN_URL }}).
+See the job run logs and results [here]({{ env.RUN_URL }}).
 
 # Review and publish archives
 
@@ -55,6 +55,3 @@ For each run that failed because of another reason (e.g., underlying data change
 ```[tasklist]
 - [ ] dataset
 ```
-
-# Relevant logs
-[Link to logs from GHA run]({{ env.RUN_URL }})
--- a/.github/ISSUE_TEMPLATE/monthly-archive-update.md
+++ b/.github/ISSUE_TEMPLATE/monthly-archive-update.md
@@ -8,8 +8,7 @@ assignees: e-belfer
 ---
 
 # Summary of results:
-See the job run results [here][1].
-[1]: {{ env.RUN_URL }}
+See the job run results [here]({{ env.RUN_URL }}).
 
 # Review and publish archives
 
@@ -58,4 +57,4 @@ For each run that failed because of another reason (e.g., underlying data change
 ```
 
 # Relevant logs
-[Link to logs from GHA run][1]
+[Link to logs from GHA run]({{ env.RUN_URL }})
--- a/.github/ISSUE_TEMPLATE/monthly-archive-update.md
+++ b/.github/ISSUE_TEMPLATE/monthly-archive-update.md
@@ -1,11 +1,15 @@
 ---
 name: Monthly archive update
 about: Template for publishing monthly archives.
-title: Publish archives for the month of MONTH
+title: Publish {{ date | date('MMMM Do YYYY') }} archives
 labels: automation, zenodo
-assignees: ''
+assignees: e-belfer
 
 ---
+
+# Summary of results:
+See the job run results [here][1].
+[1]: {{ env.RUN_URL }}
 
 # Review and publish archives
 
@@ -50,8 +54,8 @@ If the validation failure is blocking (e.g., file format incorrect, whole datase
 For each run that failed because of another reason (e.g., underlying data changes, code failures), create an issue describing the failure and take necessary steps to resolve it.
 
 ```[tasklist]
-- [ ]
+- [ ] dataset
 ```
 
 # Relevant logs
-[Link to logs from GHA run]( PLEASE FIND THE ACTUAL LINK AND FILL IN HERE )
+[Link to logs from GHA run][1]
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -15,10 +15,6 @@ jobs:
       matrix:
         dataset:
           - ferc2
-          - ferc6
-          - ferc60
-          - ferc714
-          - mshamines
           - phmsagas
 
       fail-fast: false
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -24,7 +24,7 @@ jobs:
         shell: bash -l {0}
     strategy:
       matrix:
-        dataset: ${{ fromJSON(format('[{0}]', inputs.dataset || 'eia176,eia191,eia757a,eia860,eia860m,eia861,eia923,eia930,eiaaeo,eiawater,eia_bulk_elec,epacamd_eia,ferc1,ferc2,ferc6,ferc60,ferc714,mshamines,nrelatb,phmsagas,epacems')) }}
+        dataset: ${{ fromJSON(format('{{'dataset':[{0}]}}', github.event.inputs.dataset || 'eia176,eia191,eia757a,eia860,eia860m,eia861,eia923,eia930,eiaaeo,eiawater,eia_bulk_elec,epacamd_eia,ferc1,ferc2,ferc6,ferc60,ferc714,mshamines,nrelatb,phmsagas,epacems')) }}
 
       fail-fast: false
     runs-on: ubuntu-latest
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -89,3 +89,18 @@ jobs:
           payload: ${{ steps.all_summaries.outputs.SLACK_PAYLOAD }}
         env:
           SLACK_BOT_TOKEN: ${{ secrets.PUDL_DEPLOY_SLACK_TOKEN }}
+
+  make-github-issue:
+    runs-on: ubuntu-latest
+    needs:
+      - archive-run-small
+    if: ${{ always() }}
+    steps:
+      - name: Create an issue
+      - uses: actions/checkout@v3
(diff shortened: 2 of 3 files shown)


.github/workflows/run-archiver.yml
```diff
@@ -3,6 +3,22 @@ name: run-archiver
 
 on:
   workflow_dispatch:
+    inputs:
+      small_runner:
```
e-belfer, line 7: These variables will all only set if there's a manual workflow dispatch. If the action runs as scheduled, these variables should _I believe_ evaluate as empty strings, but the documentation on boolean behavior is sparse.

.github/workflows/run-archiver.yml
```diff
@@ -13,28 +29,7 @@ jobs:
...
-          - ferc1
-          - ferc2
-          - ferc6
-          - ferc60
-          - ferc714
-          - mshamines
-          - nrelatb
-          - phmsagas
-
+        dataset: ${{ fromJSON(format('[{0}]', inputs.small_runner || '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","ferc1","ferc2","ferc6","ferc60","ferc714","mshamines","nrelatb","phmsagas"')) }}
```
e-belfer, line 32: If scheduled, will default to the full list.

.github/workflows/run-archiver.yml
```diff
@@ -78,6 +73,7 @@ jobs:
           path: ${{ matrix.dataset }}_run_summary.json
 
   archive-run-large:
+    if: inputs.large_runner
```
e-belfer, line 76: If set as true in workflow dispatch, or triggered by scheduled run this _should_ run.

.github/workflows/run-archiver.yml
```diff
@@ -160,3 +155,19 @@ jobs:
           payload: ${{ steps.all_summaries.outputs.SLACK_PAYLOAD }}
         env:
           SLACK_BOT_TOKEN: ${{ secrets.PUDL_DEPLOY_SLACK_TOKEN }}
+
+  make-github-issue:
+    if: always() && inputs.create_github_issue != false
```
e-belfer, line 160: See discussion here for use of `always()` https://github.com/actions/runner/issues/491

jdangerx (changes requested):
Broadly this looks good! I have a few small questions that I'd love to see addressed before we merge. If you think that the implied changes aren't useful, feel free to say so and re-request review :)

In terms of testing - pretty tricky to do more than just "trigger manual runs and see if things work the way you expect." I think that level of testing is fine, and if the scheduled runs fail we can always handle that manually.

.github/workflows/run-archiver.yml
e-belfer: It's a bit tricky here because there should in fact be two variables - "small runner datasets" and "large runner datasets". The alternative is one variable with some kind of filtering, but I couldn't figure out how to do that neatly. But I can make a "small" and "large" dataset list.

.github/workflows/run-archiver.yml
e-belfer: My assumption was that an empty string would actually get evaluated as true, but I could be totally off-base here. I think your suggestion re: incorporating the type of run below is great and I'll incorporate it here.

.github/workflows/run-archiver.yml
e-belfer: I think that's a great idea, can incorporate it here.

.github/ISSUE_TEMPLATE/monthly-archive-update.md
```diff
@@ -1,12 +1,15 @@
 ---
 name: Monthly archive update
 about: Template for publishing monthly archives.
-title: Publish archives for the month of MONTH
+title: Publish {{ date | date('MMMM Do YYYY') }} archives
 labels: automation, zenodo
-assignees: ''
+assignees: e-belfer
```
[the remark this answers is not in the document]
e-belfer: I don't mind doing it, personally! Think it's less likely to fall through the cracks, and I'm on the team that would be handling the quick turnaround data integration anyways.

.github/workflows/run-archiver.yml
jdangerx: I *think* an unset variable here would get treated as a "falsey" value: https://docs.github.com/en/actions/learn-github-actions/expressions#literals 

> Note that in conditionals, falsy values (false, 0, -0, "", '', null) are coerced to false and truthy (true and other non-falsy values) are coerced to true.

And actually I bet unset variable is actually `null` instead of `''`, now that I look at those docs.

.github/workflows/run-archiver.yml
```diff
@@ -1,8 +1,28 @@
...
+env:
+  SMALL_DATASETS: '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","ferc1","ferc2","ferc6","ferc60","ferc714","mshamines","nrelatb","phmsagas"'
+  LARGE_DATASETS: '"epacems"' # Datasets requiring a large (paid) runner
+
 on:
   workflow_dispatch:
+    inputs:
+      small_runner:
+        description: 'Small runner: Comma-separated list of datasets to archive (e.g., "ferc2","ferc6").'
+        default: '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","ferc1","ferc2","ferc6","ferc60","ferc714","mshamines","nrelatb","phmsagas"'
```
e-belfer, line 13: While it would be neater, this doesn't like to take an environment variable as an input.

.github/ISSUE_TEMPLATE/monthly-archive-update.md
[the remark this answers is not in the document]
jdangerx: We'll just need to make sure to pick it up if you're out for some reason or another. Which should be easy if we're doing a good job of looking at the issues assigned to whoever is out...

.github/workflows/run-archiver.yml
jdangerx: Hm, since you're already doing `inputs.small_runner || env.SMALL_DATASETS` at the use site, it seems like you could just let this [default to `""`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputs) and then get the desired "default to all small datasets" behavior.

.github/workflows/run-archiver.yml
e-belfer: Either way, making this more explicit seems wise.

e-belfer:
Tested the large runner and github issue creation [here](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/9584137263), all looks good. Issue is [here](https://github.com/catalyst-cooperative/pudl-archiver/issues/361). @jdangerx you're right that I haven't come up with a way to test the scheduled run behavior, we could either kick one off or debug this at the start of the month as needed.

.github/workflows/run-archiver.yml
```diff
@@ -78,20 +75,20 @@ jobs:
...
+    if: ${{ github.event_name == 'schedule' || inputs.large_runner }}
     defaults:
       run:
         shell: bash -l {0}
     strategy:
       matrix:
-        dataset:
-          - epacems
+        # Note that we can't pass global env variables to the matrix, so we manually list the datasets here.
+        dataset: ${{ fromJSON(format('[{0}]', '"epacems"' )) }}
```
jdangerx, line 85: I suppose we could use the same pattern as with the small runners, and have `inputs.large_runner` actually represent a list of large-runner datasets we would like to run. But since we only have one of those for now I think this is fine!

jdangerx (approved):
🚢 - no need to futz around more in the tangled web of GHA if this works 😄 .

Sure there's some things that "could" be cleaner, but is the time saved down the line worth the upfront investment now? I don't think so at this point.

.github/workflows/run-archiver.yml
e-belfer: I wanted to prevent us from accidentally kicking off large runners for datasets that don't need them. Another way to do this would've been with an input box constrained against a preset list. But think that'll be more worth revisiting when another dataset gets merged in that calls for a large runner.

Issue #349, opened 2024-05-21
Update FERC archivers to handle multiple taxonomies in one year


jdangerx:
### Overview

FERC form 1 was reported using multiple taxonomies in 2023:

* 2023 Q1 through Q3 used a `2023-04-01` taxonomy
* 2023 Q4 used a `2023-11-01` taxonomy, which had some fields slightly renamed to address ambiguity introduced in the `2023-04-01` update. See catalyst-cooperative/pudl#3544 for details.

This means that in the extraction, we need to use different taxonomies per file. Which also means that we should *track* which taxonomy to use in the metadata.

### Success Criteria
* [ ] filename is in metadata (#178 )
* [ ] taxonomy version is also in metadata
* [ ] taxonomy archives include full date instead of just year

```[tasklist]
### Next steps
* [ ] #178 
* [ ] open the XML files, look for the taxonomy links, and parse out a version from them
* [ ] update taxonomy archive filenames
```

zaneselvans:
I think this should probably get bumped to High or Urgent priority so that we can unblock work on the new year of FERC Form 1 data.

PR #362, opened 2024-06-20
Ferc xbrl updates

zschira:
# Background

This PR updates the FERC archivers to decouple taxonomies from years of data because some years have filings which utilize taxonomies from more than one year. It now will create a single zipfile will all of the taxonomies zipped inside of it.

# Details
To get the full list of taxonomy versions, XBRL archivers will now first archive all of the filings and quickly look inside each one to see the taxonomy referenced inside. Then, it will archive all of the referenced taxonomies in a single zip file.

### DIFF
--- a/src/pudl_archiver/archivers/ferc/xbrl.py
+++ b/src/pudl_archiver/archivers/ferc/xbrl.py
@@ -302,9 +302,7 @@ async def archive_taxonomies(
             "taxonomy_versions": taxonomy_versions,
             "data_format": "XBRL_TAXONOMY",
         },
-        layout=ZipLayout(
-            file_paths=[f"{version}.zip" for version in taxonomy_versions]
-        ),
+        layout=ZipLayout(file_paths=taxonomy_versions),
     )
 
 
@@ -333,6 +331,7 @@ async def archive_year(
     # Track taxonomies referenced by all filings for archival
     taxonomies_referenced = set()
 
+    files_in_zip = ["rssfeed"]
     with zipfile.ZipFile(archive_path, "w", compression=ZIP_DEFLATED) as archive:
         for filing in tqdm(filings, desc=f"FERC {form.value} {year} XBRL"):
             # Add filing metadata
@@ -365,6 +364,7 @@ async def archive_year(
             taxonomies_referenced.add(filing_metadata.taxonomy_url)
 
             with archive.open(filename, "w") as f:
+                files_in_zip.append(filename)
                 f.write(response_bytes)
 
         # Save snapshot of RSS feed
@@ -381,6 +381,7 @@ async def archive_year(
             "data_format": "XBRL",
             "taxonomies_referenced": taxonomies_referenced,
         },
+        layout=ZipLayout(file_paths=files_in_zip),
     )
 
 
--- a/src/pudl_archiver/archivers/ferc/xbrl.py
+++ b/src/pudl_archiver/archivers/ferc/xbrl.py
@@ -6,7 +6,6 @@ import io
 import json
 import logging
 import re
-import time
 import zipfile
 from collections import defaultdict
 from collections.abc import Callable
@@ -21,7 +20,7 @@ import aiohttp
 import feedparser
 from arelle import Cntlr, ModelManager, ModelXbrl
 from dateutil import rrule
-from pydantic import BaseModel, Field, HttpUrl, field_validator, model_validator
+from pydantic import BaseModel, Field, HttpUrl, model_validator
 from tqdm import tqdm
 
 from pudl_archiver.archivers.classes import ResourceInfo
@@ -98,22 +97,17 @@ class FeedEntry(BaseModel):
 
     @model_validator(mode="before")
     @classmethod
-    def extract_url(cls, entry: dict):  # noqa: N805
-        """Get download URL for inline html in feed entry."""
+    def extract_url_timestamp(cls, entry: dict):  # noqa: N805
+        """Get download URL from inline html in feed entry and parse timestamp."""
+        # Get download URL
         link = XBRL_LINK_PATTERN.search(entry["summary_detail"]["value"])
         entry["download_url"] = link.group(1)
+
+        # Get published datetime
+        entry["published_parsed"] = datetime.datetime.strptime(
+            entry["published"], "%a, %d %b %Y %X %z"
+        ).astimezone(datetime.UTC)
         return entry
-
-    @field_validator("published_parsed", mode="before")
-    @classmethod
-    def parse_timestamp(cls, timestamp: time.struct_time):  # noqa: N805
-        """Parse timestamp to a standard datetime object.
-
-        The published timestamp is only available as a time.struct_time object in the
-        feed. Converting to a datetime object makes it much more usable within the
-        python ecosystem.
-        """
-        return datetime.datetime.fromtimestamp(time.mktime(tuple(timestamp)))
 
     def __hash__(self):
         """Implement hash so FeedEntry can be used in a set."""
--- a/src/pudl_archiver/archivers/ferc/xbrl.py
+++ b/src/pudl_archiver/archivers/ferc/xbrl.py
@@ -123,6 +123,41 @@ class FeedEntry(BaseModel):
         return self.download_url == other.download_url
 
 
+def _taxonomy_zip_name_from_url(url: str) -> str:
+    if not (match := TAXONOMY_URL_PATTERN.match(url)):
+        raise RuntimeError(f"{url} does not appear to be a taxonomy url.")
+    return f"{match.group(1).replace('_', '-')}.zip"
+
+
+class FilingMetadata(BaseModel):
+    """Combines RSS feed metadata with taxonomy referenced in filing."""
+
+    filename: str
+    rss_metadata: FeedEntry
+    taxonomy_url: str
+    taxonomy_zip_name: str
+
+    @classmethod
+    def from_rss_metadata(
+        cls, rss_metadata: FeedEntry, filename: str, filing_data: bytes
+    ) -> "FilingMetadata":
+        """Construct metadata from RSS feed and filing data to extract taxonomy URL."""
+        if not (match := TAXONOMY_URL_PATTERN.search(filing_data.decode().lower())):
+            raise RuntimeError(
+                f"Couldn't find taxonomy for filing {rss_metadata.download_url}"
+            )
+
+        taxonomy_url = match.group(0)
+        return cls(
+            filename=f"{rss_metadata.title}_form{rss_metadata.ferc_formname.as_int()}_{rss_metadata.ferc_period}_{round(rss_metadata.published_parsed.timestamp())}.xbrl".replace(
+                " ", "_"
+            ),
+            rss_metadata=rss_metadata,
+            taxonomy_url=taxonomy_url,
+            taxonomy_zip_name=_taxonomy_zip_name_from_url(taxonomy_url),
+        )
+
+
 FormFilings = dict[Year, set[FeedEntry]]
 """Type alias for a dictionary containing indexed filings for a single FERC form."""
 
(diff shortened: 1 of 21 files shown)


src/pudl_archiver/archivers/ferc/xbrl.py
```diff
@@ -205,63 +213,62 @@ async def archive_taxonomy(
...
+                    response_bytes = await retry_async(response.content.read)
+                    path = Path(url_parsed.path).relative_to("/")
+                    archive_files.add(path)
+
+                    with taxonomy_archive.open(str(path), "w") as f:
+                        f.write(response_bytes)
+
+            taxonomy_version = TAXONOMY_URL_PATTERN.match(taxonomy_entry_point).group(1)
+            taxonomy_versions.append(taxonomy_version)
+            with archive.open(f"{taxonomy_version}.zip", mode="w") as taxonomy_zip:
```
e-belfer, line 261: A nit: right now the file names look like `form-6_2023-04-01.zip`. Can we get them to be `form-6-2023-04-01.zip` instead to spare having to use both `-` and `_`?

e-belfer (commented):
This worked for me as expected on FERC 6, and looks good overall. Suspect @jdangerx might have more substantive comments. Just a question and one nit on my end, neither of which are blocking.

src/pudl_archiver/archivers/ferc/xbrl.py
```diff
@@ -205,63 +213,62 @@ async def archive_taxonomy(
...
-
-        resources.append(
-            ResourceInfo(
-                local_path=archive_path,
-                partitions={"year": year, "data_format": "XBRL_TAXONOMY"},
-                layout=ZipLayout(file_paths=archive_files),
+    taxonomy_versions = []
+    archive_path = output_dir / f"ferc{form.as_int()}-xbrl-taxonomies.zip"
+    with zipfile.ZipFile(archive_path, "w", compression=ZIP_DEFLATED) as archive:
+        for taxonomy_entry_point in taxonomies_referenced:
```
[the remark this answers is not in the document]
zschira: Yeah that's actually very common, so when this already uses a `set` to dedupe the taxonomy versions before trying to download them

src/pudl_archiver/archivers/ferc/xbrl.py
zschira: I kept the formatting from the ferc url's because taxonomies are referenced by URL, and I use these in the extractor to figure out which version to use when parsing filings. Wouldn't be hard to do the `_` to `-` change on the extractor side as well though, so I'm open to making the change.

jdangerx (changes requested):
We had [originally planned](https://github.com/catalyst-cooperative/pudl-archiver/issues/349) on adding the filename and referenced taxonomy to the filing-by-filing metadata in `rssfeed`. Then the extractor would read both of those from the file on disk instead of re-calculating both.

I see that in the [current extractor PR](https://github.com/catalyst-cooperative/ferc-xbrl-extractor/pull/233/files), we've:

* removed the `get_filing_name` function and replaced it with an inline `datetime.datetime.fromisoformat()` - which isn't the same as the filename logic that we have in this branch.
* added logic to look through the XML to find the right taxonomy per filing

Both of these changes add coupling between the `ferc-xbrl-extractor` and `pudl-archiver` repos - if we change how we name the files, we have to change two places; if we change how the taxonomies are archived, we have to change two places.

I'm not sure if there were issues you ran into with the original approach that led to the changes - if not, I think our lives will be made easier if we stick to the original plan. Happy to hop on a call to sort this out too!

zschira:
@jdangerx I actually added the filenames to the metadata, so the datetime thing doesn't have anything to do with the filenames anymore (now that I think about it, I think we can fully remove the `publication_times` thing).

As for adding the taxonomies to the metadata, I fully agree that it's a good idea (and I think it's actually a very small change here and in the extractor), and not 100% sure why I didn't do that. I think I might have had some tunnel vision. That being said, I don't think this change really decouples these repos given that the extractor still has to know where the metadata is and how it's structured. I'd be happy to chat about this, but I think ultimately this repo, the extractor, and PUDL are deeply coupled in ways that make them very frustrating to work on, but are probably bigger than can/should be addressed here.

jdangerx:
> @jdangerx I actually added the filenames to the metadata, so the datetime thing doesn't have anything to do with the filenames anymore (now that I think about it, I think we can fully remove the publication_times thing).

Huh! I must have missed that - I expected there to be a new `filename` field in the `FeedEntry`, but I must be looking in the wrong place for `rssfeed` updates:

```python
            if filing_name in metadata:
                metadata[filing_name].update({filename: filing.dict()})
            else:
                metadata[filing_name] = {filename: filing.dict()
            ...
        with archive.open("rssfeed", "w") as f:
            logger.info("Writing rss feed metadata to archive.")
            f.write(json.dumps(metadata, default=str).encode("utf-8"))
```
... now that I look at it, I guess we're replacing the goofy hexadecimal thing with the filename, yeah?


> That being said, I don't think this change really decouples these repos given that the extractor still has to know where the metadata is and how it's structured. I'd be happy to chat about this, but I think ultimately this repo, the extractor, and PUDL are deeply coupled in ways that make them very frustrating to work on, but are probably bigger than can/should be addressed here.

Yeah, I agree it doesn't make the coupling go away :) Just avoids digging the hole deeper. I think the extractor will always have a deep dependency on how archiver works, but the more we can shove that knowledge into the interface between the two (`rssfeed`) the easier that dependency will be to manage. If we touched these repos more I would maybe advocate for using some sort of thrift/protobuf thing to help track the schema changes so we can evolve the schema in non-breaking ways. But... we don't, so that feels a bit overkill.

zschira:
> now that I look at it, I guess we're replacing the goofy hexadecimal thing with the filename, yeah?

Yeah! I replaced the UUID's, which ended up being meaningless because they're not stable, with filenames. I might just wrap this all in a pydantic structure to make it a bit more readable.

> I think the extractor will always have a deep dependency on how archiver works, but the more we can shove that knowledge into the interface between the two (rssfeed) the easier that dependency will be to manage

Yeah I just find it really frustrating that both PUDL and the extractor have to know about the archive structure. Moving all of the extraction into PUDL would at least reduce the number of interfaces/amount of coupling, and I don't actually think it's a huge lift, but still definitely out of scope for this.

src/pudl_archiver/archivers/ferc/xbrl.py
```diff
@@ -108,11 +115,47 @@ def parse_timestamp(cls, timestamp: time.struct_time):  # noqa: N805
...
+    ) -> "FilingMetadata":
+        """Construct metadata from RSS feed and filing data to extract taxonomy URL."""
+        if not (match := TAXONOMY_URL_PATTERN.search(filing_data.decode().lower())):
+            raise RuntimeError(
+                f"Couldn't find taxonomy for filing {rss_metadata.download_url}"
+            )
+
+        taxonomy_url = match.group(0)
+        return cls(
+            filename=f"{rss_metadata.title}_form{rss_metadata.ferc_formname.as_int()}_{rss_metadata.ferc_period}_{round(rss_metadata.published_parsed.timestamp())}.xbrl".replace(
```
jdangerx, line 152: nit: this specific string formatting thing would be clearer as a separate function - or the `filename` can just be an `@property`.

src/pudl_archiver/archivers/ferc/xbrl.py
```diff
@@ -108,11 +115,47 @@ def parse_timestamp(cls, timestamp: time.struct_time):  # noqa: N805
...
+    """Combines RSS feed metadata with taxonomy referenced in filing."""
+
+    filename: str
+    rss_metadata: FeedEntry
+    taxonomy_url: str
+    taxonomy_zip_name: str
+
+    @classmethod
+    def from_rss_metadata(
+        cls, rss_metadata: FeedEntry, filename: str, filing_data: bytes
```
jdangerx, line 142: Seems like `filename` is unused in the function body, and we can remove this parameter.

src/pudl_archiver/archivers/ferc/xbrl.py
```diff
@@ -286,14 +330,12 @@ async def archive_year(
     metadata = {}
     archive_path = output_dir / f"ferc{form_number}-xbrl-{year}.zip"
 
+    # Track taxonomies referenced by all filings for archival
+    taxonomies_referenced = set()
```
jdangerx, line 334: non-blocking musing: instead of looking to see which taxonomies are *actually* referenced in all the filings, does it make sense to just hard-code "these are all the FERC taxonomies for each form, archive all of them"? Less in-code logic, but we have to update the map every time a new taxonomy becomes relevant.

src/pudl_archiver/archivers/ferc/xbrl.py
```diff
@@ -312,6 +354,15 @@ async def archive_year(
             filename = f"{filing.title}_form{filing.ferc_formname.as_int()}_{filing.ferc_period}_{round(filing.published_parsed.timestamp())}.xbrl".replace(
                 " ", "_"
             )
+            filing_name = f"{filing.title}{filing.ferc_period}"
+            if filing_name not in metadata:
+                metadata[filing_name] = []
+
+            filing_metadata = FilingMetadata.from_rss_metadata(
```
jdangerx, line 361: If you remove `filename` as a parameter we can also remove lines 354-356!

src/pudl_archiver/archivers/ferc/xbrl.py
```diff
@@ -312,6 +354,15 @@ async def archive_year(
             filename = f"{filing.title}_form{filing.ferc_formname.as_int()}_{filing.ferc_period}_{round(filing.published_parsed.timestamp())}.xbrl".replace(
                 " ", "_"
             )
+            filing_name = f"{filing.title}{filing.ferc_period}"
+            if filing_name not in metadata:
```
jdangerx, line 358: non-blocking: should we use a defaultdict here?

tests/unit/ferc_archiver_test.py
```diff
@@ -96,8 +84,4 @@ async def test_valid_years(module_name, test_spec, mocker):
     resources = [res async for res in archiver.get_resources()]
     # don't await these, just check to make sure they have right intention
     dbfs = [r for r in resources if r.__name__ == "get_year_dbf"]
-    xbrls = [r for r in resources if r.__name__ == "get_year_xbrl"]
```
jdangerx, line 99: Should we replace these checks with ones for `archive_xbrl_for_form`?

src/pudl_archiver/archivers/ferc/xbrl.py
zschira: Oh yeah, I actually wanted to just pass the already formatted `filename` in and then use that here instead of formatting again. This is used by the extractor, so I do need the `filename` in metadata

src/pudl_archiver/archivers/ferc/xbrl.py
zschira: I figured since we already have to look inside the filings to put the referenced taxonomy in metadata, that we might as well use that to get the list of taxonomies. I'm not particularly tied to that though.

src/pudl_archiver/archivers/ferc/xbrl.py
zschira: Removed by using `defaultdict`!

src/pudl_archiver/archivers/ferc/xbrl.py
zschira: Yes! For some reason I *always* forget that `defaultdict` exists even though it's super useful!

tests/unit/ferc_archiver_test.py
zschira: Probably!

zschira:
Not sure what's going on with the CI failure? Looks like something weird with PUDL dependency?

src/pudl_archiver/archivers/ferc/xbrl.py
jdangerx: Aha! Now it is being used.

jdangerx (approved):
This generally looks good! Not quite sure what's going on with CI here either... there's some dependency conflict with the geo stuff:

```
Exception: Python bindings of GDAL 3.9.1 require at least libgdal 3.9.1, but 3.9.0 was found
```

We have 

```
  - gdal>=3.9.0
```

In the `environment.yaml`, maybe we can try bumping that to 3.9.1? But the version of GDAL in the PUDL conda-locks is also 3.9.0, so maybe we can pin this to "==3.9.0"?

jdangerx (commented):
Oh, actually, it would be great if you changed the `published_parsed` to convert to UTC - I think that would make `rssfeed` just a hair nicer to read on the other end. But it's non-blocking since the published-parsed *is* tz-aware as far as I know.

zschira:
Looked into this more and the `published_parsed` seemed to be in UTC, but were actually timezone unaware. When trying to explicitly make this UTC it was modifying the timestamp (I was probably doing something weird), but there's also a `published` field in the raw feed entries that contain a datetime string with UTC offset, so I switched to parsing these explicitly, which works fine.

Issue #372, opened 2024-07-03
Improve archivers and automated workflow to support faster data integration


e-belfer:
In order to meet new deadlines for archiver creation and integration, we need to make several changes on the archiver side to enable quick automatic detection of new data archives.

- [x] Set up a daily archiver to check for early release (June, July) and final release data (Sept, Oct) for EIA 860, 861 and 923 data #373 and #374
- [x] Remove hard-coded year from NREL ATB archiver #377 
- [x] Remove hard-coded year from EIA 176 archiver #377

e-belfer:
Created new issue for AEO archiver, which will not be necessary to address until 2025. As such, closing this issue.

PR #373, opened 2024-07-03
Add GHA template to check early and final release data daily in release months, create issue

e-belfer:
# Overview

Addresses first part of #372.

We currently don't have a regular way of checking what day ER and FR data are released. This PR makes a new GHA workflow that runs the archivers every day in June and July (looking for ER data) and September and October (looking for FR data). This currently only runs on 860, 861 and 923.

Made a new GHA workflow and corresponding issue template.

# Testing

Once merged into main, we should run the workflow manually and check that it has kicked off the expected archives, and created the expected issue. We'll need to do this in order to be able to run the workflow without implementing some `on: push` workaround (e.g., as discussed [here](https://stackoverflow.com/questions/63362126/github-actions-how-can-i-run-a-workflow-created-on-a-non-master-branch-from-t?noredirect=1&lq=1#comment125886357_71057825)).

# To-do list

```[tasklist]
```

### DIFF
--- a/.github/workflows/early_final_release_checker.yml
+++ b/.github/workflows/early_final_release_checker.yml
@@ -1,0 +1,122 @@
+---
+name: early-final-release-checker
+
+on:
+  workflow_dispatch:
+    inputs:
+      small_runner:
+        description: 'Small runner: Comma-separated list of datasets to archive (e.g., "ferc2","ferc6").'
+        # We can't pass env variables to the workflow_dispatch, so we manually list all small datasets here.
+        default: '"eia860m","eia861","eia923"'
+        required: true
+        type: string
+      create_github_issue:
+        description: "Create a Github issue from this run?"
+        default: false
+        required: true
+        type: boolean
+  schedule:
+    - cron: "21 8 2-31 6-7,9-10 *" # 8:21 AM UTC, every day in June, July, Sept, Oct except the 1st
+
+jobs:
+  archive-run-small:
+    defaults:
+      run:
+        shell: bash -l {0}
+    strategy:
+      matrix:
+        # Note that we can't pass global env variables to the matrix, so we manually reproduce the list of datasets here.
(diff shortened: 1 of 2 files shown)


.github/workflows/early_final_release_checker.yml
```diff
@@ -0,0 +1,122 @@
...
+        default: '"eia860m","eia861","eia923"'
+        required: true
+        type: string
+      create_github_issue:
+        description: "Create a Github issue from this run?"
+        default: false
+        required: true
+        type: boolean
+  schedule:
+    - cron: "21 8 2-31 6-7,9-10 *" # 8:21 AM UTC, every day in June, July, Sept, Oct except the 1st
```
e-belfer, line 19: Every day except for the 1st day of the month, when the big archiver job will be running

.github/workflows/early_final_release_checker.yml
```diff
@@ -0,0 +1,122 @@
...
+
+  make-github-issue:
+    runs-on: ubuntu-latest
+    needs:
+      - archive-run-small
+      - archive-notify
+    env:
+      OUTPUT1: ${{needs.archive-notify.outputs.output1}}
+      # If Slack message includes new data, create a Github issue
+    if: ${{ always() && (contains(OUTPUT1, "CREATE") || contains(OUTPUT1, "UPDATE") || inputs.create_github_issue == true) }}
```
bendnorman, line 113: Is this just saying if the archiver found an update for the data, create a new issue? Is there another way to determine this rather than checking if the slack message contains the word update?

bendnorman (approved):
LGTM! We'll just have to merge it into to `main` to actually test it out.

.github/workflows/early_final_release_checker.yml
e-belfer: We could split the munging of the summary files into one step, save that as an output, and then process it both into a slack message and into the Github issue? I'd still be looking for the same text in the combined summary files, so it would just change what I'm pointing at. Otherwise, we'd need to repeat the munging step twice.

PR #374, opened 2024-07-04
Test and fix GHA for early and final release data

e-belfer:
# Overview

Now that #373 is merged, let's test it and fix it.

Get the right dataset, update various failures in how the workflow file was originally set up.

# Testing

Run the workflow. If there is no new data, it shouldn't create a GH issue unless the check issue box is selected at workflow startup.

# To-do list

```[tasklist]
```

### DIFF
--- a/.github/workflows/early_final_release_checker.yml
+++ b/.github/workflows/early_final_release_checker.yml
@@ -75,7 +75,7 @@ jobs:
       - archive-run-small
     if: ${{ always() }}
     outputs:
-      output1: ${{ steps.all_summaries.outputs }}
+      output1: ${{ steps.all_summaries.outputs.SLACK_PAYLOAD }}
     steps:
       - uses: actions/checkout@v4
       - name: Download summaries
--- a/.github/workflows/early_final_release_checker.yml
+++ b/.github/workflows/early_final_release_checker.yml
@@ -110,7 +110,7 @@ jobs:
     env:
       OUTPUT1: ${{needs.archive-notify.outputs.output1}}
       # If Slack message includes new data, create a Github issue
-    if: ${{ always() && (contains(OUTPUT1, "CREATE") || contains(OUTPUT1, "UPDATE") || inputs.create_github_issue == true) }}
+    if: ${{ always() && (contains(env.OUTPUT1, "CREATE") || contains(env.OUTPUT1, "UPDATE") || inputs.create_github_issue == true) }}
     steps:
       - uses: actions/checkout@v4
       - name: Create an issue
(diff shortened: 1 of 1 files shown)


.github/workflows/early_final_release_checker.yml
```diff
@@ -7,7 +7,7 @@ on:
       small_runner:
         description: 'Small runner: Comma-separated list of datasets to archive (e.g., "ferc2","ferc6").'
         # We can't pass env variables to the workflow_dispatch, so we manually list all small datasets here.
```
zaneselvans, line 9: Not blocking but a tiny comment cleanup:

```suggestion
        # We can't pass env variables to the workflow_dispatch, so we manually list the target datasets here.
```

.github/workflows/early_final_release_checker.yml
```diff
@@ -107,14 +107,12 @@ jobs:
...
       - archive-run-small
       - archive-notify
-    env:
-      OUTPUT1: ${{needs.archive-notify.outputs.output1}}
-      # If Slack message includes new data, create a Github issue
-    if: ${{ always() && (contains(OUTPUT1, "CREATE") || contains(OUTPUT1, "UPDATE") || inputs.create_github_issue == true) }}
     steps:
       - uses: actions/checkout@v4
       - name: Create an issue
         uses: JasonEtco/create-an-issue@v2.9.2
```
zaneselvans, line 113: Not blocking at all but I wonder if using the `gh` CLI for things like creating issues (or any other GitHub API interaction) would be a cleaner and more flexible solution?

https://docs.github.com/en/actions/using-workflows/using-github-cli-in-workflows

.github/workflows/early_final_release_checker.yml
```diff
@@ -107,14 +107,12 @@ jobs:
...
+        # If Slack message includes new data, create a Github issue
+        if: ${{ always() && ( inputs.create_github_issue == true || (contains(needs.archive-notify.outputs.output1, 'CREATE') || contains(needs.archive-notify.outputs.output1, 'UPDATE'))) }}
```
zaneselvans, line 115: Not blocking and I suspect this was cut-and-pasted from elsewhere but why is the `always()` necessary in this condition?

zaneselvans (approved):
Some questions but nothing blocking.

.github/workflows/early_final_release_checker.yml
e-belfer: It's to address a very unintuitive behavior of GHA. See https://github.com/actions/runner/issues/491

.github/workflows/early_final_release_checker.yml
e-belfer: Quite possibly, but no idea how this would interact with the environment variables I'm passing and I've already probably spent too much time on this, so I'll shelf this proposal for now.

PR #377, opened 2024-07-04
Update 176 and ATB archivers to not hardcode year

e-belfer:
# Overview

Addresses part of #372.

With hardcoded years, we currently have to remember to check whether a new year of data is out. The monthly archiver runs won't capture new years of data released.

Removes hardcoded year from NREL ATB and EIA 176 archivers. 

# Testing

Run both archivers. Verify all years expected to download are successfully downloaded.

# To-do list

```[tasklist]
```

### DIFF
--- a/src/pudl_archiver/archivers/eia/eia176.py
+++ b/src/pudl_archiver/archivers/eia/eia176.py
@@ -1,4 +1,9 @@
-"""Download EIA 176 data."""
+"""Download EIA 176 data.
+
+Unlike the other NGQV datasets, the "reports" provided for EIA 176 do not contain all
+fields of data. To get around this, we query the list of item codes for 176 from the
+portal, and then manually download subsets of data from all item codes for all years.
+"""
 
 import asyncio
 import logging
@@ -33,7 +38,7 @@ class Eia176Archiver(EiaNGQVArchiver):
     items_url = "https://www.eia.gov/naturalgas/ngqs/data/items"
 
     async def get_items(self, url: str = items_url) -> list[str]:
-        """Get list of item codes from EIA NQGS portal."""
+        """Get list of item codes from EIA NQGV portal."""
         logger.info("Getting list of items in Form 176.")
         items_response = await self.get_json(url)
         items_list = [item["item"] for item in items_response]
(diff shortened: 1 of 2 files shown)


src/pudl_archiver/archivers/eia/eia176.py
```diff
@@ -41,7 +42,11 @@ async def get_items(self, url: str = items_url) -> list[str]:
     async def get_resources(self) -> ArchiveAwaitable:
         """Download EIA 176 resources."""
         items_list = await self.get_items(url=self.items_url)
-        for year in range(1997, 2023):
+        datasets_list = await self.get_datasets(url=self.base_url, form=self.form)
+        dataset_years = set()
+        for dataset in datasets_list:
+            dataset_years.update([year.ayear for year in dataset.available_years])
```
zaneselvans, line 48: So this is getting a list of available years for the dataset from the API?  At first glance I thought maybe it was looking at the years specified in the data source metadata "available years".  A comment or note in the docstring would help avoid that confusion.

src/pudl_archiver/archivers/eia/eia176.py
```diff
@@ -41,7 +42,11 @@ async def get_items(self, url: str = items_url) -> list[str]:
     async def get_resources(self) -> ArchiveAwaitable:
         """Download EIA 176 resources."""
         items_list = await self.get_items(url=self.items_url)
-        for year in range(1997, 2023):
+        datasets_list = await self.get_datasets(url=self.base_url, form=self.form)
```
zaneselvans, line 45: Should this line still be referring to `self.base_url` which has been removed above?

zaneselvans (changes requested):
* Request for a bit more info in docstring.
* Question about maybe a variable name that needs to be updated?

src/pudl_archiver/archivers/eia/eia176.py
e-belfer: `base_url` is defined in the `EiaNGQVArchiver` class, it is the correct link here. I've updated the docstring to clarify.

src/pudl_archiver/archivers/eia/eia176.py
e-belfer: Noted, I've added a clarifying string.

zaneselvans (approved):

PR #433, opened 2024-10-02
Fix erroneous FERC XBRL updates

### LINKED ISSUE #399: Investigate partition "changes" when updating FERC archives
In #398 and the corresponding archive [run](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/10194817765), every FERC XBRL archive was marked as having its partition changed, even though the links are identical to what they were. In September, if this happens again, we should investigate it. This is a two-fold task:

- Investigate the "changed" partitions
- Sort the values in the taxonomies_zip, to prevent them from changing each month by order.

zschira:
# Overview

Closes #399.

### What problem does this address?
FERC XBRL resources have been marked as changed even when there don't appear to be actual changes. There seem to have been two primary causes of these changes:

1. XBRL archives have a partition key called `taxonomies_referenced` which lists any taxonomy referenced by filings in a year of data. These have been marked as changed when only the order of the list has actually changed.
2. XBRL archives contain an rssfeed metadata json file which has also been changed due to the ordering of filings in the json.

### What did you change in this PR?
This PR addresses both of these issues:
1. When comparing partition key value pairs, sort any values that are of type `list`
2. Sort keys in metadata json before writing to zipfile

# Testing
Problem `1.` is easy to test by creating a new archive and looking at the summary to see that there are no longer partition key changes. I also added a unit test to verify the behavior with list partition values.

Problem `2.` is harder to test, however. I manually verified that sorting the rssfeed metadata does lead to these files having the same hash, but because they are bundled in zipfiles it's hard to say conclusively that there are no other differences in the zipfiles that could be contributing to the erroneous changes. The easiest/most thorough way to check this would be to publish an archive, then rerun the archiver and see if it's still detecting changes.

To get more insight into the problem, I considered modifying the change detection to actually look inside zipfiles and produce file-by-file change summaries. However, given the current archiver architecture this is very difficult because we don't actually have the zipfiles locally when looking for changes. I do think this would be very useful functionality, but would require somewhat significant reworking.

### DIFF
--- a/src/pudl_archiver/archivers/ferc/xbrl.py
+++ b/src/pudl_archiver/archivers/ferc/xbrl.py
@@ -360,7 +360,7 @@ async def archive_year(
         # Save snapshot of RSS feed
         with archive.open("rssfeed", "w") as f:
             logger.info("Writing rss feed metadata to archive.")
-            f.write(json.dumps(metadata, default=str).encode("utf-8"))
+            f.write(json.dumps(sorted(metadata), default=str, indent=2).encode("utf-8"))
 
     logger.info(f"Finished scraping ferc{form_number}-{year}.")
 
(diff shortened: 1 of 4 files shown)


e-belfer (approved):
Looks good, thanks! Agreed re: the testing, we'll give it a go and see if further changes need to take place.

PR #438, opened 2024-10-04
Archive `vcerare` data

e-belfer:
# Overview

Closes #434.

What problem does this address?
Adds `vcerare` data to Zenodo. Because this data was provided to us and isn't available online, we upload our copy to a GCS bucket so we can archive from a stable source, as well as capture any file changes over time as new data is added.

What did you change in this PR?
- Archive `vcerare files from `sources.catalyst.coop` GCS bucket
- Add `.pdf` and `.md` filetypes to frictionless
- Add `vcerare` to  `run-archiver.yml` and add GCS configuration

**Note:** Zenodo does not allow you to set which files are previewed from the API, and it automatically previews the "first file", which it determines by alphabetical sorting - this will always be `datapackage.json` for this repository. We'll need to manually check "preview" next to the README file before publishing when we update this dataset.

**Note 2:**  This PR contemplated adding contributor types to the Zenodo payload. We were setting them in `zenodo_role` but not passing them to the Zenodo API. I configured this to specify type for `creator`, as is possible in the GUI, but this is only possible for collaborators in the API. So, unless we reconfigure the way we handle creators this isn't possible - for now, we'll just made any adjustments manually in the GUI since we rarely refresh metadata.

# Testing

How did you make sure this worked? How can a reviewer verify this?
Run `pudl_archiver --datasets vcerare --sandbox --initialize`. See also: https://zenodo.org/records/13919960

# To-do list

```[tasklist]
- [x] Add GCS credentials to Github, mirroring `pudl-usage-metrics`
- [x] Review the PR yourself and call out any questions or issues you have
- [x] Debug gdal problem - see https://github.com/conda-forge/gdal-feedstock/pull/991 or pin in PUDL?
```

PR #439, opened 2024-10-04
Archive `gridpathratoolkit` data from GCS

e-belfer:
# Overview

Closes https://github.com/catalyst-cooperative/pudl/issues/3895

Archives `gridpathratoolkit` data by uploading the original files given to us by Moment Energy Insights to a private `sources.catalyst.coop` bucket, and then directing the archiver to grab the data from there. This way, we can update the data in the GCS bucket and the archiver will automatically grab the new files and update the `datapackage.json` file.

Mimic the archiver in #438, but for the GridPath data.

# Testing

Run `pudl_archiver --datasets vceregen --sandbox --initialize`

# To-do list

```[tasklist]
- [ ] Verify sandbox archive looks as expected and publish production archive
```

### DIFF
--- a/src/pudl_archiver/archivers/gridpathratoolkit.py
+++ b/src/pudl_archiver/archivers/gridpathratoolkit.py
@@ -1,0 +1,57 @@
+"""Archive GridPath RA toolkit renewable generation data.
+
+This dataset was produced by Moment Energy Insights (now Sylvan Energy Analytics).
+It is archived from files stored in the private sources.catalyst.coop bucket.
+"""
+
+import logging
+from pathlib import Path
+
+from google.cloud import storage
+
+from pudl_archiver.archivers.classes import (
+    AbstractDatasetArchiver,
+    ArchiveAwaitable,
+    ResourceInfo,
+)
+
+logger = logging.getLogger(f"catalystcoop.{__name__}")
+
+
+class GridPathRAToolkitArchiver(AbstractDatasetArchiver):
+    """GridPath RA Toolkit renewable generation profiles archiver."""
+
+    name = "gridpathratoolkit"
+    bucket_name = "sources.catalyst.coop"
+
+    async def get_resources(self) -> ArchiveAwaitable:
+        """Download VCE renewable generation resources."""
+        bucket = storage.Client().get_bucket(self.bucket_name)
+        blobs = bucket.list_blobs(prefix=f"{self.name}/")  # Get all blobs in folder
+
+        for blob in blobs:
+            # Skip the folder, which appears in this list
+            if not blob.name.endswith("/"):
+                yield self.get_gcs_resource(blob)
+
+    async def get_gcs_resource(self, blob: storage.Blob) -> tuple[Path, dict]:
+        """Download VCE renewable generation profile files from GCS.
+
+        There are three types of files: a documentation PDF, a single CSV and a series
+        of zipped annual files that are named vceregen_{year}.zip.
+        """
+        # Remove folder name (identical to dataset name) and set download path
+        file_name = blob.name.replace(f"{self.name}/", "")
+        path_to_file = self.download_directory / file_name
+        # Download blob to local file
+        logger.info(f"Downloading {blob.name} to {path_to_file}")
+        blob.download_to_filename(path_to_file)
+
+        # The partition should be the filename without the filetype extension.
+        # E.g., solar_capacity_aggregations.csv has part: solar_capacity_aggregations
+
+        part = file_name.split(".")[0]  # Remove .csv/.zip extension
+        return ResourceInfo(
+            local_path=path_to_file,
+            partitions={"part": part},
+        )
--- a/src/pudl_archiver/archivers/vceregen.py
+++ b/src/pudl_archiver/archivers/vceregen.py
@@ -73,8 +73,8 @@ class VCEReGenArchiver(AbstractDatasetArchiver):
                 partitions={"fips": True},
             )
 
-        # Handle documentation
-        if file_name.endswith(".pdf"):
+        # Handle documentation and README
+        if file_name.endswith(".pdf") or file_name.endswith(".md"):
             return ResourceInfo(local_path=path_to_file, partitions={})
 
         raise AssertionError(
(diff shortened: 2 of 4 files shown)


zaneselvans:
Now that we have this private bucket that we can use as the original source, I think it would be better to archive the entire collection of [GridPath RA Toolkit data](https://drive.google.com/drive/folders/1UbhONethtjyx9y0M4zqvK6HARyTtwnp5) with all the original filenames, and rename the files that we're archiving on Zenodo programmatically? There's a lot of additional data in there, and I created the zipfiles and named them by hand to fit our expected partition names and now looking back it's not 100% clear what the mapping was for a couple of the files...

## Solar
* `gridpath-ra-toolkit/TemporalData/HourlySolar_byProject.zip` --> `original_solar_capacity.zip`
* `gridpath-ra-toolkit/MonteCarlo_Inputs.zip/temporal_data/solar/` --> `aggregated_solar_capacity.zip`
* `gridpath-ra-toolkit/MonteCarlo_Inputs.zip/temporal_data/solar_syn/` --> `aggregated_extended_solar_capacity.zip`
* ??? --> `solar_capacity_aggregations.csv`

## Wind
* `gridpath-ra-toolkit/TemporalData/HourlyWind_byProject.zip` --> `original_wind_capacity.zip`
* `gridpath-ra-toolkit/MonteCarlo_Inputs.zip/temporal_data/wind/` --> `aggregated_wind_capacity.zip`
* `gridpath-ra-toolkit/MonteCarlo_Inputs.zip/temporal_data/wind_syn/` --> `aggregated_extended_wind_capacity.zip`
* ??? --> `wind_capacity_aggregations.csv`

## Weather
* `gridpath-ra-toolkit/TemporalData/DailyWeatherData_cleaned.csv` --> `daily_weather.csv`

I was nervous to archive all of the original files on Zenodo since some of them were based on the WECC ADS which doesn't have clear licensing terms.

e-belfer:
> I was nervous to archive all of the original files on Zenodo since some of them were based on the WECC ADS which doesn't have clear licensing terms.

Just to clarify - you're proposing archiving them all on GCS but only uploading the ones that are already published, with the rename happening during the archiving?

zaneselvans:
> Just to clarify - you're proposing archiving them all on GCS but only uploading the ones that are already published, with the rename happening during the upload to Zenodo?

Yes, exactly -- put the whole collection of data that's associated with the GridPath RA Toolkit in our private GCS bucket, and then use that as the original source (since it is), with any file renaming or zipping/unzipping that's required done in the source-to-Zenodo archiving process, like we do with our other sources.  I only renamed the files by hand because the entire archiving process was manual before and I wanted them to have a simple naming / partitioning convention.

e-belfer:
That makes sense to me - do you want to go ahead and add the raw files to the correct source folder in GCS, and I can update this archiver?

zaneselvans:
* I've uploaded the contents as published in GridLab's Google Drive to `gs://sources.catalyst.coop/gridpathratoolkit/` (plus the PDF of the GridPath RA Toolkit report)
* I've temporarily moved the old manually archived files to `gs://sources.catalyst.coop/gridpathratoolkit-manual/` (we can remove after the new archive is working)

I spent a little time trying to identify which of the original files correspond to the `{wind,solar}_capacity_aggregation.csv` files in the manual archive and wasn't successful, so you'll probably be as fast to do it as I am with some finding and grepping.

src/pudl_archiver/archivers/gridpathratoolkit.py
```diff
@@ -0,0 +1,160 @@
...
+                yield self.get_and_zip_resources(original_file, bucket)
+
+    async def get_gcs_resource(
+        self, original_file: str, bucket: storage.Bucket
+    ) -> tuple[Path, dict]:
+        """Download VCE renewable generation profile files from GCS.
+
+        There are three types of files: a documentation PDF, a series of zipped annual
+        files, a series of CSV files,  and a series of files to download from within a
+        zipped file.
```
zaneselvans, line 80: Update docstring for GridPath RA Toolkit, not VCE.

src/pudl_archiver/archivers/gridpathratoolkit.py
```diff
@@ -0,0 +1,160 @@
...
+
+Wind data:
+* gridpath-ra-toolkit/TemporalData/HourlyWind_byProject.zip --> original_wind_capacity.zip
+* gridpath-ra-toolkit/MonteCarlo_Inputs.zip/temporal_data/wind/ --> aggregated_wind_capacity.zip
+* gridpath-ra-toolkit/MonteCarlo_Inputs.zip/temporal_data/wind_syn/ --> aggregated_extended_wind_capacity.zip
+* ??? --> wind_capacity_aggregations.csv
+
+
+Weather data:
+* gridpath-ra-toolkit/TemporalData/DailyWeatherData_cleaned.csv --> daily_weather.csv
```
zaneselvans, line 22: Probably want to either update this to reflect the new file layout, or just remove it since it's now contained in the archiver.

src/pudl_archiver/archivers/gridpathratoolkit.py
```diff
@@ -0,0 +1,133 @@
...
+            # We download the entire zipfile to avoid having to authenticate using
+            # a second GCS library, since GCS doesn't support fsspec file paths.
+            logger.info(f"Downloading {blob.name} to {path_to_file}")
+
+            blob.download_to_filename(path_to_file)
+
+            # The partition should be the filename without the filetype extension.
+            # E.g., solar_capacity_aggregations.csv has part: solar_capacity_aggregations
+
+        part = file_name.split(".")[0]  # Remove .csv/.zip extension
```
zaneselvans, line 87: Not blocking, but the `str.split("char")[idx]` pattern for munging paths stored as strings is a little hard to read and also prone to breaking when we get an unexpected filename / path. For local filesystem paths the [`pathlib.Path`](https://docs.python.org/3.12/library/pathlib.html#methods-and-properties) class has good methods for grabbing various path components, and the [`universal_pathlib`](https://github.com/fsspec/universal_pathlib) package (which I think is already installed as a dependency for `fsspec`) extends `pathlib.Path` to handle remote filesystem-like URLs.

zaneselvans (approved):
Nothing blocking. I suggested a docstring and a slight change in logic to improve readability and made a note about potentially using `Path` or `UPath` rather than strings for path manipulation / parsing.

src/pudl_archiver/archivers/gridpathratoolkit.py
e-belfer: Good call, updated this!

PR #462, opened 2024-10-30
Add archiver for EIA MECS data

zschira:
# Overview

This PR adds an archiver for the [Manufacturing Energy Consumption Survey (MECS)](https://www.eia.gov/consumption/manufacturing/data/2018/#r1). This data is archived exclusively on GCS, so this PR also adds an `fsspec` based storage backend. Using `fsspec` means this backend is compatible with most cloud storage providers as well as the local filesystem, which is very useful for testing.

# To-do list

```[tasklist]
- [x] Add MECS archiver
- [x] Add fsspec based storage backend
```


src/pudl_archiver/depositors/depositor.py
```diff
@@ -362,3 +362,29 @@ async def attach_datapackage(
             )
             await self.create_file("datapackage.json", datapackage_json)
         return new_datapackage, update
+
+
+@dataclass
+class DepositionBackend:
```
zschira, line 368: This is just moved to solve a circular import

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,272 @@
...
+            for child in deposition_path.iterdir()
+            if child.name != deposition_path.name
+        ]
+        return cls(
+            deposition_path=deposition_path,
+            file_list=file_list,
+        )
+
+
+class FsspecAPIClient(DepositorAPIClient):
```
zschira, line 72: This module implements the `fsspec` based storage backend for archivers. The `upath` package allows working with cloud storage/local filesystems interchangeably using `UPath` objects that behave like normal python `Path`'s.

src/pudl_archiver/frictionless.py
```diff
@@ -160,3 +160,32 @@ def new_datapackage(
             description=data_source.description,
             version=version,
         )
+
+    @classmethod
```
zschira, line 164: This is a hacky method to work around relying on metadata from PUDL. It's pretty ugly/bad, but I think it's ok for the time being until we come up with a better solution for our metadata handling.

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,272 @@
...
+
+    base_path: str = "gs://archives.catalyst.coop"
+
+
+class LocalAPIClient(FsspecAPIClient):
+    """FsspecAPIClient variant using local file system as backend."""
+
+    base_path: str = "file://archives.catalyst.coop"
+
+
```
zschira, line 264: I've created a depositor backend for both GCS and the local filesystem. These can technically be used for any dataset, although the depositor could probably use more development before more significant use.

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,272 @@
...
+)
+
+logger = logging.getLogger(f"catalystcoop.{__name__}")
+
+
+def _resource_from_upath(path: UPath, parts: dict[str, str]) -> Resource:
+    """Create a resource from a single file with partitions.
+
+    Args:
+        file: Deposition file metadata returned by Zenodo api.
```
e-belfer, line 29: This seems like leftover cruft from Zenodo, as it's not a required input. Conversely looks like `path` is missing here.

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,272 @@
...
+class FsspecAPIClient(DepositorAPIClient):
+    """Implement API for fsspec based depositors."""
+
+    base_path: str
+
+    @classmethod
+    async def initialize_client(
+        cls,
+        session: aiohttp.ClientSession,
+        sandbox: bool,
```
e-belfer, line 81: Does `sandbox` have any meaning in this context? If not, does it need to be in the parameters here? Might be worth raising an assertion if it is passed to note that this isn't implemented for this storage backend.

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,272 @@
...
+
+    async def cleanup_after_error(self, e: Exception):
+        """Cleanup draft after an error during an archive run."""
+        logger.error(
+            f"Failed while creating new deposition: {traceback.print_exception(e)}"
+        )
+
+    async def delete_deposition(self) -> None:
+        """Delete an un-submitted deposition."""
+        pass
```
e-belfer, line 199: Do we want to do something more forceful than passing here? E.g. log or raise an error? Or actually delete the files in the folder? In what context is this getting called?

e-belfer (changes requested):
A few questions about the design of this thing, and a few suggestions about places to add more bumpers to ward against unintended uses of this backend.

At a high-level, I feel like I'm missing something - where does the selection of the depositor actually occur? If it's the `pudl_archiver` flags we should document this in the CLI - if not, it'd be helpful to better document what config determines this.

In general, fsspec is a bit out of my wheelhouse and I suspect @jdangerx or @bendnorman might have better design feedback here.

I have not tested this locally as I don't want to overwrite the existing outputs.

src/pudl_archiver/depositors/fsspec.py
zschira: FIxed!

src/pudl_archiver/depositors/fsspec.py
zschira: Good call, now it will raise a `NotImplementedError` if sandbox is set.

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,272 @@
...
+        """Open a new draft to make edits."""
+        return FsspecDraftDeposition(
+            deposition=self.deposition,
+            settings=self.settings,
+            api_client=self.api_client,
+            dataset_id=self.dataset_id,
+        )
+
+
+class FsspecDraftDeposition(DraftDeposition):
```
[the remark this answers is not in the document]
zschira: Draft/published isn't a meaningful distinction right now since we haven't implemented any kind of versioning for the fsspec backend, but I think that's more a sign that this backend is still immature and we wouldn't want to use it for our main PUDL archives at this point.

Ideally, we will eventually expand this backend to support versioning, so there would be a published archive which is read only and will always point to a static version, and a draft which can be modified and eventually published.

src/pudl_archiver/depositors/fsspec.py
zschira: Right now this will only get called if no changes are detected after attempting to create a new archive. Since this backend doesn't support versioning this doesn't really make sense, so I've changed this to raise an error.

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # Create updated datapackage
+        resources = [
+            _resource_from_upath(
+                self.deposition.deposition_path / fname, resource_info[fname].partitions
+            )
+            for fname in self.deposition.file_list
+            if fname != "datapackage.json"
+        ]
+        datapackage = DataPackage.mecs(
```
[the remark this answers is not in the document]
zschira: I've changed this to call a generic `new_datapackage` method and added some slightly janky logic to that method to handle the `mecs` case. In general, it would be good to not have to get metadata from PUDL so we could handle this more cleanly, but that's a bigger rabbit hole.

zschira:
@e-belfer I added some documentation about the depositor backend options in the `README` including a warning that this is still very experimental and should be used with caution.

In general, I think this is still in a very early/janky state that is good enough to get files on GCS for right now, but would still need significant work to be something we depend on more officially.

README.md
```diff
@@ -80,6 +80,12 @@ There are also four optional flags available:
   the new Zenodo DOI to the `dataset_doi.yaml` file.
 - `--all`: shortcut for archiving all datasets that we have defined archivers
   for. Overrides `--datasets`.
+- `--depositor`: select backend storage system. Defaults to `zenodo`, which is
+  the only fully featured backend at this point, but we are experimenting with an
+  `fsspec` based backend which would allow using various cloud storage backends like
+  GCS. Available options include `gcs`, `local`, and `zeonodo`. `gcs` and `local` use
```
jdangerx, line 86:
```suggestion
  GCS. Available options include `gcs`, `local`, and `zenodo`. `gcs` and `local` use
```

i'm helpful!

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,285 @@
...
+
+import logging
+import traceback
+from typing import BinaryIO
+
+import aiohttp
+from pydantic import BaseModel, ConfigDict
+from upath import UPath
+
+from ..frictionless import MEDIA_TYPES, DataPackage, Resource, ResourceInfo
```
jdangerx, line 11: Can these be absolute instead of relative imports?

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,285 @@
...
+        title=path.name,
+        mediatype=mt,
+        parts=parts,
+        bytes=path.stat().st_size,
+        hash=compute_md5(path),
+        format=path.suffix,
+    )
+
+
+class Deposition(BaseModel):
```
jdangerx, line 47: non-blocking - this is what represents the deposition *contents* as opposed to sort of the "logical state" of the deposition as a whole (`Draft/PublishedDeposition`), yes?

src/pudl_archiver/depositors/fsspec.py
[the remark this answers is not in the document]
jdangerx: There's some logical notion of a deposition being "done" or not, which I think this captures, despite there not being a difference on disk right now.

I suppose we could rename `draft`/`published` as `readwrite`, `readonly` which might be clearer? 🤷

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,285 @@
...
+
+    async def get_deposition(self, dataset_id: str):
+        """Get latest version of deposition associated with dataset_id."""
+        return Deposition.from_upath(
+            deposition_path=UPath(self.base_path) / dataset_id,
+        )
+
+    async def create_new_deposition(self, dataset_id: str):
+        """Prepare new deposition associated with dataset_id."""
+        deposition_path = UPath(self.base_path) / dataset_id
```
jdangerx, line 104: Huh, cool, UPath just uses fsspec under the hood. Smart.

src/pudl_archiver/frictionless.py
jdangerx: That time might be a long time in the future... this is to avoid throwing this metadata into `pudl.metadata.resources` basically?

jdangerx (changes requested):
The usage of `fsspec`/`UPath` seems legit to me. Architecture wise I think the naming of `Deposition` and `DraftDeposition`/`PublishedDeposition` is confusing. I think that's my fault, sorry.

Anyways. Things look good overall, from an infrastructure perspective. Didn't pay too much attention to what you are actually doing to get the MECS data but it appears to be working.

One thing I'd request changes for is some sort of testing around the fsspec-based depositors - maybe mess with some temporary directories in an integration test?

zaneselvans:
Is there a reason why we aren't referring to the EIA MECS with its form number (846), like we do for all of the other EIA datasets?  [All the EIA Forms by number](https://www.eia.gov/Survey/index.php)

src/pudl_archiver/cli.py
```diff
@@ -64,8 +64,8 @@ def parse_main(args=None):
     )
     (
         parser.add_argument(
-            "--download-dir",
```
zschira, line 67: I removed the `download-dir` option because I don't think we're using it much and it could now be replaced with using the fsspec depositor with a local path to view resources locally.

I don't love that there's coupling between the `deposition-path` and `depositor` options, but I couldn't think of a better solution with spending more time than I really wanted on this.

e-belfer (approved):
Just want to confirm that running this only requires Zenodo API keys if using Zenodo API backend? Otherwise mostly small comments and a few non-blocking questions on my end, so I'll go ahead and approve it.

src/pudl_archiver/cli.py
zschira: I don't think it's impossible, but I felt like keeping both of these args around or overloading it to have 2 different meanings would cause more confusion than it's worth. There are still lots of avenues for reviewing data before publishing to zenodo like using the sandbox, turning `auto-publish` off and manually reviewing a draft, or using the new fsspec depositor to download locally first.

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -72,26 +73,37 @@ def from_upath(cls, deposition_path: UPath):
...
+                "There is no sandbox available for fsspec archiver. "
+                "An alternative for testing would be to use the 'local' depositor backend."
+            )
+        return cls(path=deposition_path)
 
     async def get_deposition(self, dataset_id: str):
         """Get latest version of deposition associated with dataset_id."""
         return Deposition.from_upath(
-            deposition_path=UPath(self.base_path) / dataset_id,
+            deposition_path=UPath(self.path),
```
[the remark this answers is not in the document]
zschira: Yeah that was my idea. We could definitely change back if that'd be preferred, but I thought it might be confusing if you pass a path and it ends up creating an unexpected sub-directory.

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -0,0 +1,268 @@
...
+
+    deposition_path: UPath
+    #: static list of files from deposition prior to any modifications
+    file_list: list[str]
+    version: str = "0.1"
+
+    @classmethod
+    def from_upath(cls, deposition_path: UPath):
+        """Construct deposition object from fsspec path to deposition."""
+        assert deposition_path.exists()
```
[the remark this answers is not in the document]
zschira: I don't think so, I've changed this now

src/pudl_archiver/cli.py
```diff
@@ -64,8 +64,8 @@ def parse_main(args=None):
...
-            help="Directory to download files to. Use tmpdir if not specified.",
+            "--deposition-path",
+            help="Configurable base path used by some depositor backends.",
             default=None,
         ),
     )
```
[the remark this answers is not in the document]
zschira: Added a help message, and it'll error out if you pass something else.

src/pudl_archiver/cli.py
```diff
@@ -77,6 +81,7 @@ def parse_main(args=None):
     parser.add_argument(
         "--depositor",
         type=str,
+        help="Specifies what backend engine will be used for archive storage. Current allowable options include zenodo and fsspec",
```
e-belfer, line 84: Last nit: We could use the `choices` parameter here to restrict to these valid options? https://docs.python.org/3/library/argparse.html#choices

src/pudl_archiver/frictionless.py
```diff
@@ -163,3 +175,32 @@ def new_datapackage(
...
+                    **{
+                        "name": "CC-BY-4.0",
+                        "title": "Creative Commons Attribution 4.0",
+                        "path": "https://creativecommons.org/licenses/by/4.0",
+                    }
+                )
+            ],
+            resources=sorted(resources, key=lambda x: x.name),  # Sort by filename
+            contributors=[CONTRIBUTORS["catalyst-cooperative"]],
+            created=str(datetime.utcnow()),
```
e-belfer, line 202:
```suggestion
            created=str(datetime.datetime.now(tz=datetime.UTC).isoformat()),
```
Should mirror what we do above, this currently returns an error for me when re-running the MECS archiver with an fsspec backend.

e-belfer (changes requested):
I ran into an issue with the `utcnow()` no longer existing as a method, otherwise all looks good.

src/pudl_archiver/cli.py
zschira: Good call, added!

src/pudl_archiver/depositors/fsspec.py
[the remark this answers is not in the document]
e-belfer: Oh no, that seems totally justifiable on second thought!

e-belfer:
Running ``pudl_archiver --dataset mecs --depositor fsspec --deposition-path=file:///PATH_TO_FILE --initialize`` where the path already had data in it and a folder from the last run, I'm getting the following traceback error:

2024-12-11 12:37:07 [    INFO] catalystcoop.pudl_archiver.orchestrator:32 Deleting mecs from deposition.
Encountered exceptions, showing traceback for last one: ["('mecs', ValueError('Cannot delete directory, set recursive=True'))"]
Traceback (most recent call last):
  ....
  File "/home/../pudl-archiver/src/pudl_archiver/orchestrator.py", line 33, in orchestrate_run
    draft = await draft.delete_file(filename)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/../pudl-archiver/src/pudl_archiver/depositors/fsspec.py", line 194, in delete_file
    (self.deposition.deposition_path / filename).unlink()
  File "/home/../miniforge3/envs/pudl-cataloger/lib/python3.12/site-packages/upath/core.py", line 1008, in unlink
    self.fs.rm(self.path, recursive=False)
  File "/home/../miniforge3/envs/pudl-cataloger/lib/python3.12/site-packages/fsspec/implementations/local.py", line 180, in rm
    raise ValueError("Cannot delete directory, set recursive=True")
ValueError: Cannot delete directory, set recursive=True

When I deleted the directory (from an older run), all worked as expected. Is there a scenario where we ever expect to have an archive with folders and want to enable this recursive deletion?

e-belfer (approved):
I will note that all the `publish_with_valid()` logging messages make no sense if you're running this locally, as they all refer to drafts and publishing still. But for now I think that's out of scope for this PR. Left one question about recursive deletion that would be helpful to at least confirm is not necessary to include here, otherwise all works as expected.

zschira:
@e-belfer I think having sub-directories within an archive would cause other problems, but I'm not really sure. I could take a look at enabling this if it's something we want to support, but I haven't really thought about enabling this before.

PR #465, opened 2024-11-01
Fix quoting error in workflow file

zaneselvans:
# Overview

There was a syntax error in the JSON that we feed into the monthly archive script as a matrix and so it failed to run November 1st. This PR literally adds a single character to that JSON blob to double-quote `"ferc1"`.

# Testing

I've kicked off a `workflow_dispatch` run manually it seems to be running as expected.

### DIFF
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -6,7 +6,7 @@ on:
     inputs:
       datasets:
         description: 'Comma-separated list of datasets to archive (e.g., "ferc2","ferc6").'
-        default: '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","epacems",ferc1","ferc2","ferc6","ferc60","ferc714","mshamines","nrelatb","phmsagas","vcerare"'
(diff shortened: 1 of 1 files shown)


.github/workflows/run-archiver.yml
```diff
@@ -6,7 +6,7 @@ on:
     inputs:
       datasets:
         description: 'Comma-separated list of datasets to archive (e.g., "ferc2","ferc6").'
-        default: '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","epacems",ferc1","ferc2","ferc6","ferc60","ferc714","mshamines","nrelatb","phmsagas","vcerare"'
+        default: '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","epacems","ferc1","ferc2","ferc6","ferc60","ferc714","mshamines","nrelatb","phmsagas","vcerare"'
```
zaneselvans, line 9: Not sure how I did it, but this one character error was my fault. Introduced when I switched us off the bigger paid runner.

zaneselvans: And it doesn't get caught by the action linter because it's an error in our little JSON blob.

jdangerx: Ugh. Slipped past both of us. Well, at least you caught it now!

jdangerx (approved):

PR #500, opened 2024-12-17
If no epacems_api_key, return `None` instead of throwing an exception

e-belfer:
# Overview

Right now if there's no `EPACEMS_API_KEY` set in the environment, an error is thrown whether or not a user is actually archiving EPA CEMS data. This came up in another repository importing the `pudl-archiver` code to run an archiver.

Switched from using `os.environ['EPACEMS_API_KEY']` which raises an error if this doesn't exist, to `os.environ.get("EPACEMS_API_KEY")` which will return `None`. This will still fail if the EPA CEMS API key is being called in the archiver, but will allow people to run other archivers without making a fake EPA CEMS API key environment variable.

# Testing

# To-do list

```[tasklist]
```


zschira (approved):
Looks good!

Issue #499, opened 2024-12-17
Re-design `pudl-archiver` to better separate dependencies from main `pudl` repository


e-belfer:
Our `pudl-archiver` and `pudl` repositories are linked together because we store metadata for datasets ingested in PUDL in the `pudl` repository. Right now, adding an archiver follows the following workflow:

1. Make a PR in the `pudl` repo to add the dataset's metadata to `pudl.metadata.sources`
2. Once that PR merges, recreate the `pudl-archiver` environment locally to reinstall PUDL and write an archiver.

**The problem**
When we ingest datasets we don't envision ending up in PUDL in the short-term, we need to do one of two things:
1. Add them into our PUDL metadata repository anyways.
2. Develop a hacky workaround, like in #462.

It's also impossible to use the `pudl-archiver` repository without installing PUDL. In order to be able to more broadly use our lovely archiving infrastructure to archive other datasets or to make it easier for external contributors to write archivers, we should more clearly separate these two repositories.

**Existing interconnections**
The `pudl-archiver` repository currently has the following dependencies on `pudl` functions, classes and dictionaries:

In `pudl_archiver.frictionless.py`:
```
from pudl.metadata.classes import Contributor, DataSource, License
from pudl.metadata.constants import CONTRIBUTORS
```

See in particular the `from_pudl_metadata` method in the `DataPackage` class.

**The desired solution**
- Metadata for every dataset should only be defined in one place - metadata _should not_ be duplicated in the `pudl` and `pudl-archiver` repositories. 
- There's a structured way to add archivers to the `pudl-archiver` repository without needing to open a PR in the main PUDL repo, if the archives are not clearly intended to land in PUDL.
- The `pudl` repository can be installed as an optional dependency for the `pudl-archiver` repository.

**Questions to resolve**
- If a dataset is ingested into PUDL, where should the source of truth for the metadata live?
- How can we test archivers made by external users without our Zenodo creds? -- Should we handle this in the CI, or have us test it manually?

**Some potential options to explore**
1. Use GHA to sync the source metadata files from `pudl` into the `pudl-archiver`, and add a second sources file for non-PUDL metadata. Doesn't resolve `CONTRIBUTORS` import.
Basically kicks the can down the road and creates a dependency in GH, not ideal.

2. Add a flag for archivers dependent on PUDL metadata, but make it possible for archives to rely on metadata provided in the `pudl-archiver` repository.

[...]

PR #506, opened 2025-01-07
Make it possible to add non-PUDL data sources and metadata to the archiver

e-belfer:
# Overview

A lightweight temporary solution to #499.

In concert with https://github.com/catalyst-cooperative/pudl/pull/4003, makes it possible to add non-PUDL metadata sources to the archiver directly.

- Added a new `metadata` folder that replicates the one in the PUDL repository, but only contains information about data sources that aren't integrated into PUDL (currently only MECS).
- Added a new `from_non_pudl_metadata()` method in the `DataPackage` class which calls the `NON_PUDL_SOURCES` dictionary (should work after the PR in the PUDL repository merges to make this possible).
- Updated the `README`
- Renamed `mecs` to `eiamecs` to make it more consistent with our other data sources
- Added some documentation on how to add additional contributors to `sources.py` in preparation for the hackathon, while I was at it!

# Testing

`pudl_archiver --dataset eiamecs --sandbox --initialize`

# To-do list

```[tasklist]
- [x] Fix `from_id()` to take `sources` in PUDL repo
- [ ] Once PUDL PR merges, delete cached env and test https://github.com/catalyst-cooperative/pudl/pull/4013
```


src/pudl_archiver/frictionless.py
```diff
@@ -177,30 +183,29 @@ def from_pudl_metadata(
...
+    def from_non_pudl_metadata(
+        cls,
+        name: str,
+        resources: Iterable[Resource],
+        version: str | None,
+    ):
+        """Create a datapackage for sources that won't end up in PUDL."""
+        # TODO: This is a slightly ugly workaround to avoid having to add sources into
+        # the from_id method - should I just fix this at the source?
+        data_source = DataSource(
```
zschira, line 195: I think if you change this to call `from_id` instead of `dict_from_id`, that *should* fix the broken tests.

zschira: Whoops, I see now that `from_id` doesn't take the `sources` dict right now. It's a bit of a pain to have to do another PUDL PR, but I think the most consistent API would be to add `sources` to the `from_id` method, since you're basically just copying what `from_id` does here.

zschira (changes requested):
Looks good. I think updating `from_id` to take `sources` as a parameter would be my preference to fix the test issues.

zschira (approved):
Sorry, got sidetracked earlier, but looks good now!

PR #507, opened 2025-01-09
Make census fips codes archiver

cmgosnell:
# Overview

working on https://github.com/catalyst-cooperative/pudl/issues/3884

- this adds an archiver which grabs that county level fips codes from the census

- i added an archiver module. I used the template in the readme and edited it mostly using patterns from the NREL ATB archiver
- I was surprised to have to add an `xls` media type. this was only required for the older files but i'm surprised we've never had to archive `xls` files. Did I miss something here?
- I had to do a semi-janky add of the [1990-2000/](https://www2.census.gov/programs-surveys/popest/geographies/1990-2000/).. txt file. 
- (i also added census metadata in pudl - i will link make that pr and link to it here)

# Testing

I ran `pudl_archiver --datasets censusfips --initialize --summary-file censusfips-summary.json`

to make a draft/sandbox archives here:
- https://zenodo.org/uploads/14648211
- https://sandbox.zenodo.org/records/151369

# To-do list

```[tasklist]
```

### DIFF
--- a/src/pudl_archiver/archivers/censusfips.py
+++ b/src/pudl_archiver/archivers/censusfips.py
@@ -26,22 +26,24 @@ class CensusFipsArchiver(AbstractDatasetArchiver):
             if not matches:
                 continue
             year = int(matches.group(1))
-            # don't go peep into the 1990-2000/ dir which just includes a txt file
-            if self.valid_year(year) and year != 2000:
+            if self.valid_year(year):
                 yield self.get_year_resource(year)
 
     async def get_year_resource(self, year: int) -> tuple[Path, dict]:
-        """Download excel file."""
+        """Download excel or txt file."""
         # before 2017, the files are xls. after that its xlsx
-        link_url = f"{BASE_URL}/{year}"
-        link_pattern = re.compile(rf"all-geocodes-v{year}.(xlsx|xls)")
-        file_names = await self.get_hyperlinks(link_url, link_pattern)
-        if len(file_names) != 1:
-            raise AssertionError(
-                f"We expected exactly one link for {year}, but we found: {file_names}"
-            )
-        file_name = file_names.pop()
+        if year != 2000:
+            link_url = f"{BASE_URL}/{year}"
+            link_pattern = re.compile(rf"all-geocodes-v{year}.(xlsx|xls)")
+            file_names = await self.get_hyperlinks(link_url, link_pattern)
+            if len(file_names) != 1:
+                raise AssertionError(
+                    f"We expected exactly one link for {year}, but we found: {file_names}"
+                )
+            file_name = file_names.pop()
+        elif year == 2000:
+            link_url = f"{BASE_URL}/1990-2000"
+            file_name = "90s-fips.txt"
         download_path = self.download_directory / file_name
         await self.download_file(f"{link_url}/{file_name}", download_path)
-
         return ResourceInfo(local_path=download_path, partitions={"year": year})
--- a/src/pudl_archiver/archivers/censusfips.py
+++ b/src/pudl_archiver/archivers/censusfips.py
@@ -9,7 +9,7 @@ from pudl_archiver.archivers.classes import (
     ResourceInfo,
 )
 
-BASE_URL = "https://www2.census.gov/programs-surveys/popest/geographies/"
+BASE_URL = "https://www2.census.gov/programs-surveys/popest/geographies"
 
 
 class CensusFipsArchiver(AbstractDatasetArchiver):
@@ -26,14 +26,15 @@ class CensusFipsArchiver(AbstractDatasetArchiver):
             if not matches:
                 continue
             year = int(matches.group(1))
-            if self.valid_year(year):
+            # don't go peep into the 1990-2000/ dir which just includes a txt file
+            if self.valid_year(year) and year != 2000:
                 yield self.get_year_resource(year)
 
     async def get_year_resource(self, year: int) -> tuple[Path, dict]:
         """Download excel file."""
         # before 2017, the files are xls. after that its xlsx
         file_extension = "xlsx" if year >= 2017 else "xls"
-        file_name = f"state-geocodes-v{year}.{file_extension}"
+        file_name = f"all-geocodes-v{year}.{file_extension}"
         url = f"{BASE_URL}/{year}/{file_name}"
         download_path = self.download_directory / file_name
         await self.download_file(url, download_path)
--- a/src/pudl_archiver/archivers/censusfips.py
+++ b/src/pudl_archiver/archivers/censusfips.py
@@ -33,10 +33,15 @@ class CensusFipsArchiver(AbstractDatasetArchiver):
     async def get_year_resource(self, year: int) -> tuple[Path, dict]:
         """Download excel file."""
         # before 2017, the files are xls. after that its xlsx
-        file_extension = "xlsx" if year >= 2017 else "xls"
-        file_name = f"all-geocodes-v{year}.{file_extension}"
-        url = f"{BASE_URL}/{year}/{file_name}"
+        link_url = f"{BASE_URL}/{year}"
+        link_pattern = re.compile(rf"all-geocodes-v{year}.(xlsx|xls)")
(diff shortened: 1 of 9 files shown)


src/pudl_archiver/archivers/censusfips.py
```diff
@@ -0,0 +1,42 @@
...
+    async def get_resources(self) -> ArchiveAwaitable:
+        """Download Census FIPS Codes resources."""
+        # the BASE_URL page has a bunch of links with YEAR/ at the end
+        link_pattern = re.compile(r"(\d{4})/$")
+        for link in await self.get_hyperlinks(BASE_URL, link_pattern):
+            matches = link_pattern.search(link)
+            if not matches:
+                continue
+            year = int(matches.group(1))
+            # don't go peep into the 1990-2000/ dir which just includes a txt file
```
e-belfer, line 29: I would just archive this file as `year=2001` based on the metadata in the file - we can decide whether or not to extract it on the PUDL side (we won't), but no reason not to grab it along with the rest.

e-belfer (changes requested):
I'd actually suggest archiving the 1990-2000 file, which notes that it was last updated for the 2001 population estimate in the metadata, as `year=2001`. We're not planning on extracting anything but the last file in PUDL anyways, so we won't have to deal with extraction but it's relatively straightforward to archive and I don't see an argument for not doing it?

src/pudl_archiver/archivers/censusfips.py
cmgosnell: okay! which is to say I'd grab this txt file in `get_year_resource`. hm which makes me wonder if we should also be grabbing all of the related txt files along with the excel files.

cmgosnell: oh the other txt files are just different versions of the same data so nvm!

src/pudl_archiver/archivers/censusfips.py
```diff
@@ -0,0 +1,42 @@
...
+    ResourceInfo,
+)
+
+BASE_URL = "https://www2.census.gov/programs-surveys/popest/geographies"
+
+
+class CensusFipsArchiver(AbstractDatasetArchiver):
+    """Census FIPS Codes archiver."""
+
+    name = "censusfips"
```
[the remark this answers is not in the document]
cmgosnell: The popest is the broader population estimates dataset right? This archive will be a smol portion from that broader dataset so that doesn't seem wrong but it does seem a little misleading. I kinda copied the named based on the other census dataset - which is also grabbing a smol specific piece of a broader dataset. but idk i don't feel clear on our conventions here when we are going after little slices like this and would be happy to go either way.

src/pudl_archiver/archivers/censusfips.py
cmgosnell: hm i also don't feel clear based on the metadata in the txt file that this should be `year=2001` vs `year=2000`

>  after the 1990 Census and are reflected in Census 2000 and the 2001 population estimates

src/pudl_archiver/archivers/censusfips.py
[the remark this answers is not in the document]
cmgosnell: i'm going to move to censuspep for Population Estimates Program which is not the folder name but is how the census webpages refer to this program/dataset

src/pudl_archiver/archivers/censuspep.py
```diff
@@ -0,0 +1,52 @@
...
+            link_pattern = re.compile(rf"all-geocodes-v{year}.(xlsx|xls)")
+            file_names = await self.get_hyperlinks(link_url, link_pattern)
+            if len(file_names) != 1:
+                raise AssertionError(
+                    f"We expected exactly one link for {year}, but we found: {file_names}"
+                )
+            file_name = file_names.pop()
+        elif year == 2000:
+            link_url = f"{BASE_URL}/1990-2000"
+            file_name = "90s-fips.txt"
```
e-belfer, line 49: Non-blocker: We do have the power to rename this file to be consistent with the others, and we do typically rename files to match a consistent `datasource_year.extension` format for our other datasets.

e-belfer (approved):
A non-blocking comment about the file names, which maybe we want to make consistent with each other and with our other datasets?

src/pudl_archiver/archivers/censuspep.py
cmgosnell: ah ty i was wondering about this! i believe we can rename it np!

cmgosnell:
some questions mostly about :
- do all or most of our archives expect that we will have one **zipped** file per partition? I believe i can use `download_and_zip_file` for that. but the docs for that and `download_file` are not consistent so it was a lil confusing to use. i can update the args to make em slightly more clear one i sorta understand em.
- post making one `--initialize` archive, do i have to store the doi somewhere? if i remove it from the command on the readme i get this error:

```python
File "/Users/christinagosnell/code/pudl-archiver/src/pudl_archiver/depositors/zenodo/depositor.py", line 467, in get_deposition
    concept_rec_id = concept_doi.split(".")[2]
                     ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'split'
```

- does the sandbox work at all?
```python
  File "/Users/christinagosnell/code/pudl-archiver/src/pudl_archiver/depositors/zenodo/depositor.py", line 526, in run_request
    raise ZenodoClientError(
pudl_archiver.depositors.zenodo.depositor.ZenodoClientError: ZenodoClientError(status=400, message=CSRF token missing or incorrect., errors=None)
[1]    96686 exit 1     pudl_archiver --datasets censuspep --initialize --summary-file  --sandbox
```

src/pudl_archiver/archivers/censuspep.py
```diff
@@ -0,0 +1,53 @@
...
+        url = f"{link_url}/{file_name}"
+        download_path = self.download_directory / f"{self.name}_{year}.zip"
```
e-belfer, line 51:
```suggestion
        download_path = self.download_directory / f"{self.name}-{year}.zip"
```
Is consistent with how we handle this in most of our other datasets (sometimes we aren't perfect about this)

src/pudl_archiver/archivers/classes.py
```diff
@@ -138,82 +138,75 @@ def get_resources(self) -> ArchiveAwaitable:
...
     async def download_zipfile(
-        self, url: str, file: Path | io.BytesIO, retries: int = 5, **kwargs
+        self, url: str, zip_path: Path | io.BytesIO, retries: int = 5, **kwargs
     ):
         """Attempt to download a zipfile and retry if zipfile is invalid.
 
         Args:
             url: URL of zipfile.
-            file: Local path to write file to disk or bytes object to save file in memory.
+            file_path: Local path to write file to disk or bytes object to save file in memory.
```
e-belfer, line 147:
```suggestion
            zip_path: Local path to write file to disk or bytes object to save file in memory.
```

src/pudl_archiver/package_data/zenodo_doi.yaml
```diff
@@ -1,6 +1,9 @@
 censusdp1tract:
   production_doi: 10.5281/zenodo.4127048
   sandbox_doi: 10.5072/zenodo.3138
+censuspep:
+  production_doi: 10.5281/zenodo.14624611
```
e-belfer, line 5: Have you published a production archive already, given that this PR is still under review?

e-belfer (changes requested):
It's impossible for me to test this without https://github.com/catalyst-cooperative/pudl/pull/4006 merging or me making environment changes, so I'd like to get that in first before I approve this!

Otherwise, looks good in principle other than one file name change.

src/pudl_archiver/package_data/zenodo_doi.yaml
cmgosnell: I did :-/ it was definitely jumping the gun. i did this on a friday when i had a good amount of time to get working on the integration into pudl and no sandbox key!

zaneselvans (approved):

PR #522, opened 2025-01-19
Update environment.yml

zaneselvans:
# Overview

Update gdal to v3.10.1 to match the main pudl repo

# Testing

Ran the tests here...


zaneselvans:
Merged this because the failure is Zenodo (error 500!) and it will *definitely* be broken without a matching version of gdal.

@e-belfer I'm worried Zenodo may be having another bad API period, just as we try to make a bunch of new archivers. 🫠

PR #528, opened 2025-01-20
Add FUNDING.yml with OC & GitHub Sponsorship links

zaneselvans:


README.md
```diff
@@ -231,28 +245,32 @@ async def get_year_resource(self, link: str, year: int) -> ResourceInfo:
...
+
 This method should handle the following steps:
-* identify the specific download link for the file(s) in the partition
-* rename the file to match our data conventions. We rename files to match the format
+
+- identify the specific download link for the file(s) in the partition
+- rename the file to match our data conventions. We rename files to match the format
 `datasource-partition.ext` - e.g. `eia860-1990.zip`.
-* construct the path to where we want to temporarily store the file locally, using
+- construct the path to where we want to temporarily store the file locally, using
```
cmgosnell, line 254: were all of these swaps from `*` to `-` necessary?

README.md
```diff
@@ -366,5 +389,60 @@ To use it:
...
+
+## Supporting PUDL
+
+These raw data archivers are part of the Public Utility Data Liberation Project (PUDL),
+a project of [Catalyst Cooperative](https://catalyst.coop). PUDL has been made possible
+by the generous support of our sustainers, grant funders, and volunteer open source
+contributors.
+
+If you would like to support the ongoing development of PUDL, please consider
+[becoming a sustainer](https://opencollective.com/pudl).
```
cmgosnell, line 404: I don't love the idea of adding this full content below with all of the tiers and sustainers that we will certainly occasionally forget to update in multiple places. My suggestion would be to just add the opening paragraph here but not include all of the logos and details here.

cmgosnell (changes requested):
this is not a hill i will die on but i don't think we should sprinkle all of the sustainers info all over the place. this isn't a particularly popping repo compared to pudl proper.

README.md
zaneselvans: It was rendering fine on GitHub but my markdown linter was complaining about them and some other minor formatting details, and I just accepted its suggestions.

README.md
zaneselvans: I wasn't thrilled with it either but I added it because we said we'd highlight them, and because I think it's also positive for us to show who else is involved in the project. But you're right that this repo gets far less traffic so it probably doesn't matter much. I'll remove.

cmgosnell (approved):
thanks for removing the logos and things! I like the idea in general but ya link to the oc page w/ all of that deets here feels fine bc this isn't our most traffic-y repo!

PR #533, opened 2025-01-22
Hackathon metadata 2

aesharpe:
Add metadata for three more datasets: 
- EPA MATS (this is kinda two datasets)
- IRA Energy Communities
- EPA Disadvantaged Communities (EJ Screen)


src/pudl_archiver/metadata/sources.py
```diff
@@ -287,4 +287,114 @@
         "license_pudl": LICENSES["cc-by-4.0"],
         "contributors": [CONTRIBUTORS["catalyst-cooperative"]],
     },
+    "eccoalclosure": {
```
cmgosnell, line 290:
```suggestion
    "doeiraec": {
```

src/pudl_archiver/metadata/sources.py
```diff
@@ -287,4 +287,114 @@
...
+        "title": "Coal Closure Energy Communities Dataset",
```
cmgosnell, line 291:
```suggestion
        "title": "DOE -- IRA Energy Community Data Layers",
```

src/pudl_archiver/metadata/sources.py
```diff
@@ -287,4 +287,114 @@
...
+                "ira",
+            }
+        ),
+        "license_raw": LICENSES["us-govt"],
+        "license_pudl": LICENSES["cc-by-4.0"],
+        "contributors": [CONTRIBUTORS["catalyst-cooperative"]],
+    },
+    "ecffemployment": {
+        "title": "MSA/Non-MSAs that only meets the Fossil Fuel Employment 0.17% Threshold Dataset",
+        "path": "https://edx.netl.doe.gov/dataset/ira-energy-community-data-layers",
```
cmgosnell, line 316: this is the same link from above

src/pudl_archiver/metadata/sources.py
```diff
@@ -287,4 +287,114 @@
...
+                "mines",
+                "closure",
+                "ira",
+            }
+        ),
+        "license_raw": LICENSES["us-govt"],
+        "license_pudl": LICENSES["cc-by-4.0"],
+        "contributors": [CONTRIBUTORS["catalyst-cooperative"]],
+    },
+    "ecffemployment": {
```
cmgosnell, line 314: the name should follow our convention of {source}{dataset}

src/pudl_archiver/metadata/sources.py
```diff
@@ -287,4 +287,114 @@
...
+        "title": "MSA/Non-MSAs that only meets the Fossil Fuel Employment 0.17% Threshold Dataset",
```
cmgosnell, line 315: can you make this title also follow the convention of "SOURCE NAME -- Full title"

cmgosnell (changes requested):

src/pudl_archiver/metadata/sources.py
```diff
@@ -287,4 +287,106 @@
...
+                "ira",
+            }
+        ),
+        "license_raw": LICENSES["us-govt"],
+        "license_pudl": LICENSES["cc-by-4.0"],
+        "contributors": [CONTRIBUTORS["catalyst-cooperative"]],
+    },
+    "epadcejp": {
+        "title": "EPA -- Disadvantaged Community Energy Justice Program Dataset",
+        "path": "https://www.arcgis.com/home/webmap/viewer.html?url=https://geopub.epa.gov/arcgis/rest/services/ejscreen/environmental_climate_justice_program/MapServer&source=sd",
```
zaneselvans, line 324: These paths are for human consumption, so pointing at a top level documentation page for the dataset rather than the map is probably preferable.  Maybe it's this one?
```suggestion
        "path": "https://www.epa.gov/ejscreen/technical-information-and-data-downloads",
```

For the actual archiving, the ArcGIS MapServer API endpoint might be where we want to point people:

https://geopub.epa.gov/arcgis/rest/services/ejscreen/environmental_climate_justice_program/MapServer

The compound URL you have is saying "use the ArcGIS website to display the data available at this MapServer endpoint"

src/pudl_archiver/metadata/sources.py
```diff
@@ -287,4 +287,115 @@
...
+        "description": (
+            "This dataset denotes whether a region is considered a disadvantaged community based on "
+            "the EPA's Energy Justice Program. An EPA Disadvantaged Community under the Energy Justice "
+            "Program is identified based on four key factors: environmental challenges like high pollution "
+            "exposure, economic hardships such as low income or high unemployment, social vulnerabilities "
+            "including health disparities and minority status, and geographic disadvantages in rural, urban, "
+            "or remote areas. These criteria ensure that support is directed to communities facing systemic "
+            "inequities and the greatest need for clean energy and environmental investments."
+        ),
+        "working_partitions": {"years": [2023]},  # not 100% sure on year
```
zaneselvans, line 334: This is a pretty new dataset that probably doesn't have any update frequency so years is probably not the right partition. But it's spatial so maybe by state? Depends how the data is available on the MapServer API.

aesharpe: We're not even sure whether we can download this data tbh so maybe best to just add a comment saying not sure what the partitions are and leave it at that

aesharpe: And whoever works on this dataset can update it.

zaneselvans: Yeah, Null partitions is probably fine. I'm pretty sure you can pull stuff from the MapServer -- it's a very friendly API IIRC.

zaneselvans:
I'm re-running the CI, but that file change detection error failure is something I was getting erratically on the other metadata PR and it didn't seem to be a real issue, so we can probably ignore it for now.

zaneselvans (approved):

PR #534, opened 2025-01-22
Archiver for EIA RECS

### LINKED ISSUE #519: Write an archiver for EIA Residential Energy Consumption Survey
### Motivation and context:

Briefly describe the dataset. What is it, and why do we want to archive it regularly?

Links:
- [Main dataset page](https://www.eia.gov/consumption/residential/index.php)
- [Most recent data page](https://www.eia.gov/consumption/residential/data/2020/)

Notes:
- this dataset is structured very similarly to EIA MECS in that it has a page for each year of data.
- The data gets released every 5 ish years.
- The older data is only published as PDFs. Stretch goal here would be to grab both the excel and pdf data but the excel data is way higher priority.

#### Requirements for archiving

To be archived on Zenodo, a dataset must be:
- [ ] published under an open license that permits reuse and redistribution
- [ ] less than 50Gb in size (when zipped)
- [ ] relevant to energy modelling and research

### Checklist for archive creation

Based on the [README documentation on creating a new archive](https://github.com/catalyst-cooperative/pudl-archiver#adding-a-new-dataset):

```[tasklist]
- [x] [Define the dataset's metadata](https://github.com/catalyst-cooperative/pudl-archiver#step-1-define-the-datasets-metadata)
- [ ] [Implement archiver interface](https://github.com/catalyst-cooperative/pudl-archiver#step-2-implement-archiver-interface)
- [ ] [Test archiver locally](https://github.com/catalyst-cooperative/pudl-archiver#step-3-test-archiver-locally)
- [ ] [Test uploading to Zenodo](https://github.com/catalyst-cooperative/pudl-archiver#step-4-test-uploading-to-zenodo)
- [ ] [Manually review archive before publication](https://github.com/catalyst-cooperative/pudl-archiver#step-5-manually-review-your-archive-before-publication)
- [ ] [Finalize archive](https://github.com/catalyst-cooperative/pudl-archiver#step-6-finalizing-the-archive) (only core Catalyst developers can complete this step)
- [ ] [Automate archiving](https://github.com/catalyst-cooperative/pudl-archiver#step-7-automate-archiving)
```
### Links to published archives:
Include a link to the published sandbox archive for review.

nilaykumar:
# Overview

Closes #519.

Implements archiver for [EIA RECS](https://www.eia.gov/consumption/residential/data/2020/).

Added archival of some of the 2020 .xlsx files.

# Testing

```sh
pudl_archiver --datasets eiarecs \
    --initialize \
    --summary-file eiarecs-summary.json \
    --depositor fsspec \
    --deposition-path /tmp/eia-recs/
```

# To-do list

```[tasklist]
- [ ] add other years (before 2020)
- [ ] add PDF
- [x] add microdata
- [ ] update gha workflows
```


src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,115 @@
...
+    ArchiveAwaitable,
+    ResourceInfo,
+)
+from pudl_archiver.frictionless import ZipLayout
+
+LINK_PATTERNS = [
+    # housing characteristics
+    {
+        "base_url": "https://www.eia.gov/consumption/residential/data",
+        "php_extension": "index.php?view=characteristics",
```
nilaykumar, line 17: The RECS [webpage](https://www.eia.gov/consumption/residential/data/2020/) is a little bit tricky. It has tabs containing different sequences of datafiles (and PDFs, etc.) that can't be reached directly from the base_url. Instead we need to tack on these strings to the end of the url, which are different for different tabs.

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,115 @@
...
+        "pattern": re.compile(r"State (.*)\.xlsx"),
+    },
+    # state data (consumption & expenditures)
+    {
+        "base_url": "https://www.eia.gov/consumption/residential/data",
+        "php_extension": "index.php?view=state",
+        "prefix": "state-ce",
+        "pattern": re.compile(r"ce(\d)\.(\d{1,2})\.(.*)\.xlsx"),
+    },
+    # microdata
```
nilaykumar, line 42: I'm skipping the microdata for now. Happy to come back and rewrite everything to be more general once I get a better sense from folks who have a sense of how deep we want to go down this rabbit hole.

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,115 @@
...
+            yield self.get_year_resources(year)
+
+    async def get_year_resources(self, year: int) -> list[ResourceInfo]:
+        """Download all excel tables for a year."""
+        # Loop through all download links for tables
+        tables = []
+        zip_path = self.download_directory / f"eia-recs-{year}.zip"
+        data_paths_in_archive = set()
+        # Loop through different categories of data (all .xlsx)
+        for pattern_dict in LINK_PATTERNS:
```
nilaykumar, line 71: We call `get_hyperlinks` multiple times: once for each tab of data.

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,115 @@
...
+            table_link_pattern = pattern_dict["pattern"]
+            for table_link in await self.get_hyperlinks(url, table_link_pattern):
+                table_link = f"{year_url}/{table_link}"
+                logger.info(f"Fetching {table_link}")
+                # Get table major/minor number from links
+                match = table_link_pattern.search(table_link)
+                # We've gotta do a bit of wrangling to get the output filename
+                # to match the url somewhat
+                n_groups = len(match.groups())
+                output_filename = f"eia-recs-{year}-{pattern_dict['prefix']}"
```
nilaykumar, line 84: This is kinda ugly wrangling to set the output filename appropriately. It would probably be better to specify directly in the dictionary of LINK_PATTERNS above, instead of weird if statements catching the behavior of the regexes in LINK_PATTERNS.

jdangerx: Hmm, from a cursory reading, this might do something similar-enough so as to be still human-readable? Otherwise defining some helper functions like `format_housing_characteristics_filename` and referencing them in the LINK_PATTERNS seems nice.

```python
output_filename += "-".join(g for g in match.groups() if g).replace(" ", "_")
```

src/pudl_archiver/archivers/eia/eiarecs.py
jdangerx: What a fun webpage!

src/pudl_archiver/archivers/eia/eiarecs.py
jdangerx: Sounds good - we can revisit later. I'm unaware of the context of this rabbit hole - assuming you all had a discussion during the hackathon? @cmgosnell

src/pudl_archiver/archivers/eia/eiarecs.py
jdangerx: Might be worth defining the shape of this dictionary as a dataclass, but definitely not a blocking concern.

jdangerx (commented):
This all looks pretty good - provided some feedback on the output file name munging but nothing's blocking. 

Your TODO list looks pretty good - though I'll make a draft in Zenodo just so we can have something at all archived ASAP.

jdangerx:
Discussed on Slack with Nilay and I'm taking this over the finish line.

[Draft record (currently including 2020 and 2015) on Zenodo](https://zenodo.org/uploads/14768379)

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,231 @@
...
+        f"https://www.eia.gov/consumption/residential/data/{year}/index.php?view={view}"
+    )
+
+
+# Each year, each tab's format changes. Rather than have complicated regexes that capture everything, just have lots of simple regexes
+YEAR_LINK_SETS = {
+    2020: {
+        "housing_characteristics": LinkSet(
+            url=_url_for(year=2020, view="characteristics"),
+            short_name="hc",
```
cmgosnell, line 45: I'd suggest using the view name (or some other something longer) instead of the short name in the file names so the files are more decipherable.

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,231 @@
...
+            short_name="microdata",
+            pattern=re.compile(r"(recs.*public.*)\.csv"),
+            extension="csv",
+        ),
+        "microdata-codebook": LinkSet(
+            url=_url_for(year=2020, view="microdata"),
+            short_name="microdata",
+            pattern=re.compile(r"(RECS 2020 Codebook.*v.)\.xlsx"),
+            extension="xlsx",
+        ),
```
cmgosnell, line 78: not blocking but i would suggest making these a little more generic so there aren't so many to maintain

they both have the `year` and they both have `recs` if your add `re.IGNORECASE` and you could `(\.csv|\.xlsx)` for different file types. this is kinda a high level overall suggestion to condense patterns when possible but not a hill i will die on.

src/pudl_archiver/archivers/eia/eiarecs.py
cmgosnell: yea if it were me i'd condense these two `LinkSet` attributes as just `view` and use the view name in the filename and construct the page url with `_url_for` in the `get_year_resources`

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+class EiaRECSArchiver(AbstractDatasetArchiver):
+    """EIA RECS archiver."""
+
+    name = "eiarecs"
+
+    async def get_resources(self) -> ArchiveAwaitable:
+        """Download EIA-RECS resources."""
+        for year in [2020, 2015, 2009]:
```
cmgosnell, line 180:
```suggestion
        for year in YEAR_LINK_SETS:
```

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,231 @@
...
+        link_sets = YEAR_LINK_SETS[year]
+        for link_set in link_sets.values():
+            for table_link in await self.get_hyperlinks(link_set.url, link_set.pattern):
+                table_link = urljoin(link_set.url, table_link).strip("/")
+                logger.info(f"Fetching {table_link}")
+                match = link_set.pattern.search(table_link)
+                matched_filename = (
+                    match.group(1)
+                    .replace(".", "-")
+                    .replace(" ", "_")
```
cmgosnell, line 204: we've typically attempted to always use dashes instead of _ for file names
```suggestion
                    .replace(" ", "-")
                    .replace("_", "-")
```

cmgosnell (changes requested):
the draft archive looks pretty good! i have a few small specific suggestions but in general when possible to make the link patterns more generic i'd suggest doing that so there aren't so so many to add for each year and view and sub-view file group.

its also kind of a bummer to not have `get_resources` be able to check if there is a new year of data. i had to poke around a lot to find somewhere for CBECS. I wonder if you could just use "https://www.eia.gov/consumption/residential/data/previous.php" and search for something like `^/consumption/residential/data/(\d{4})/$`

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,280 @@
...
+        Returns:
+            ResourceInfo: information about this year's zip file & its contents.
+        """
+        self.logger.info(f"Starting {year}")
+
+        tab_infos = await self.__select_tabs(url)
+
+        # most tabs for most years can be handled the same way
+        tab_handlers = {
+            "housing-characteristics": defaultdict(lambda: self.__get_tab_links),
```
jdangerx, line 91: In [1997](https://www.eia.gov/consumption/residential/data/1997/)/[1993](https://www.eia.gov/consumption/residential/data/1993/) there's a lot of context on the housing characteristics/consumption & expenditures tabs. We might want to archive the tab contents there too.

At which point we might want to just archive the tab contents for every GD tab - would make this "different defaults for each tab type, plus overrides" logic a lot simpler.

What do you think?

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,280 @@
...
+                )
+                logger.debug(f"Added {link} to {zip_path} as {output_filename}")
+                data_paths_in_archive.append(output_filename)
+            download_path.unlink()
+        return data_paths_in_archive
+
+    async def __get_tab_links(self, tab_info: TabInfo, zip_path: Path) -> list[str]:
+        """Get the data files for a single tab."""
+        soup = await self.__get_soup(tab_info.url)
+        links_in_tab = soup.select("div.tab-contentbox a[href]")
```
jdangerx, line 162: Sometimes the EIA page is actually missing links - e.g. in 1997, there are a bunch of files that exist on their server, but the links on the data display page don't point at them where they should:

* [C&E Top 4 states - appliance expenditures](https://www.eia.gov/consumption/residential/data/1997/pdf/four-states/appl_exp4states.pdf)
* [HC5 - appliances (households, not percents)](https://www.eia.gov/consumption/residential/data/1997/pdf/housing-characteristics/appl_household.pdf)
* [microdata questionnaire section a](https://www.eia.gov/consumption/residential/data/1997/pdf/puquex_a.pdf)

Do you think it's worth adding override logic for 1997? Personally leaning no.

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,280 @@
...
+            filename=filename,
+            blob=BytesIO(html.prettify().encode("utf-8")),
+        )
+        self.logger.info(f"{log_scope}: Added html to {zip_path} under {filename}")
+        return links + [filename]
+
+    async def __get_original_forms(self, year: int, zip_path: Path) -> list[str]:
+        """Get the survey forms that were used to collect the data.
+
+        These are all on the same page, which is different from the yearly RECS
```
jdangerx, line 225: Amusingly, the 2015 forms have some that are marked 'eia-475' instead of 'eia-457'. But I checked and they seem to be a survey of residential energy consumption.

jdangerx (commented):
A couple questions about some changes I was mulling over!

cmgosnell (changes requested):
i had a hard time getting to a level of understanding around what the heck the soup selections were returning and how to think about that. i've asked for a little bit more docs in a few key places.

I opened a few of the html files that were made and they definitely contain the important text, but the links within those files are janky. i'm not sure there a good way of getting around that given the circumstances - or if we'd want to put effort into that. but i jsut wanted to flag it as a thing.

an overall comment on much of this recs module contains much useful tooling that could be deployed within the `AbstractDatasetArchiver` or otherwise. this is non-blocking - i don't think we should attempt to integrate it within this pr but we should probably flag this.

but okay okay i made draft archives with this with a few changes (the most important to the archive itself imo being the `eia-recs` -> `eiarecs`, secondarily being plz clarify a few things in there) this looks pretty good and a cool new method for grabbing files!

src/pudl_archiver/archivers/eia/eiarecs.py
```diff
@@ -0,0 +1,280 @@
...
+                    name=re.sub(r"\W+", "-", tab.text.strip()).lower(),
+                    year=year,
+                )
+                for tab in unselected_tabs
+            }
+
+        first_unselected_tabs = await get_unselected_tabs(url)
+        another_tab_url = next(iter(first_unselected_tabs)).url
+        next_unselected_tabs = await get_unselected_tabs(another_tab_url)
+        return first_unselected_tabs.union(next_unselected_tabs)
```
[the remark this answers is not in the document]
jdangerx: Yes, and the issue is that the "data" dropdown links to `/data/YEAR` which then automatically opens a tab, and that tab is always housing characteristics which we care about :(

cmgosnell (approved):
these updates look good to me!

PR #538, opened 2025-01-23
generalize hyperlink extractor

marianneke:
# Overview

Closes #XXXX.

- Hyperlink extractor does not return URLs in tags other than `<a>` tags
- Hyperlink extractor also does not return information inside the tags, which could contain the actual filename, e.g. `<a href="...">filename.zip</a>`
- In the case of the new data archiver https://github.com/catalyst-cooperative/pudl-archiver/pull/539 the name would be needed for extracting properties such as dataset upload date.

- Hyperlink extractor returns all hyperlinks (based on some list of attributes rather than only the href attribute in the <a> tag)
- Return value is dict instead of set, with values {hyperlink: name} where name contains the text between tags, if present, and defaults to hyperlink itself.
- Check for name matches in both the name and the hyperlink instead of just the hyperlink

# Testing

WIP
Note that the change from datatype set -> dict should not be an issue downstream as long as the result of `get_hyperlinks` is looped over, since a simple for loop returns the keys, which are the hyperlinks, and ignores the values. This should be checked.
~Errors might occur if there is text between the tags but it lacks information that is present in the hyperlink, resulting in hyperlinks not being downloaded.~ solved by checking both name and link for pattern

# To-do list

```[tasklist]
- [ ] How to handle cases where the text between tags does not contain the filename (so we may want to default to the hyperlink for the name in that case)
- [ ] Use the name as the filename if downstream usecase does expect a valid filename, and the hyperlink does not contain the filename
```

### DIFF
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -49,15 +49,31 @@ class _HyperlinkExtractor(HTMLParser):
 
     def __init__(self):
         """Construct parser."""
-        self.hyperlinks = set()
+        self.hyperlinks = dict()
+        self.current_hyperlink = None
         super().__init__()
 
     def handle_starttag(self, tag, attrs):
         """Filter hyperlink tags and return href attribute."""
-        if tag == "a":
-            for attr, val in attrs:
-                if attr == "href":
-                    self.hyperlinks.add(val)
+        url_attrs = ["href", "src", "action", "data-url", "poster"]
+
+        for attr, val in attrs:
+            if attr in url_attrs:
+                self.current_hyperlink = val
+                if self.current_hyperlink not in self.hyperlinks:
+                    # By default, set the name identical to the hyperlink
+                    self.hyperlinks[self.current_hyperlink] = self.current_hyperlink
+                break
+
+    def handle_data(self, data):
+        """Capture text content and associate it with the current URL."""
+        if self.current_hyperlink and data.strip():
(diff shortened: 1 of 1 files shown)


src/pudl_archiver/archivers/classes.py
```diff
@@ -49,15 +49,31 @@ class _HyperlinkExtractor(HTMLParser):
 
     def __init__(self):
         """Construct parser."""
-        self.hyperlinks = set()
+        self.hyperlinks = dict()
```
zschira, line 52: I think to fix the test failure you just need to change this to:

```suggestion
        self.hyperlinks = {}
```

zschira (changes requested):
This all makes sense at a high level, I'm mostly just concerned with getting the CI passing.

I'm not 100% sure what still needs to be done for handling the filename stuff, but happy to take another look when ready, or discuss further if you have any questions. I will say, if we need to enable some more complex html extraction, I think it could be reasonable to bring in [beautiful-soup](https://beautiful-soup-4.readthedocs.io/en/latest/). Also, it's not really a big deal if we end up using different filenames from those coming from the upstream sources, so if we can construct our own meaningful files without extracting them from the html, that's a totally fine option.

zschira (approved):
This looks good to me, @jdangerx do you think we should we hold off on merging if you're testing beautiful soup, or go for it and we can replace with BS in the future?

jdangerx:
Merge it! Replacing with beautiful soup in the future seems fine.

Issue #529, opened 2025-01-22
Write an archiver for IRA Community Energy data


aesharpe:
### Motivation and context:

Data, geospatial data resources, and the linked mapping tool and web services reflect data for two types of potentially qualifying energy communities: 1) Census tracts and directly adjoining tracts that have had coal mine closures since 1999 or coal-fired electric generating unit retirements since 2009. These census tracts qualify as energy communities. 2) Metropolitan statistical areas (MSAs) and non-metropolitan statistical areas (non-MSAs) that are energy communities for 2023 and 2024, along with their fossil fuel employment (FFE) status.

Given that the IRA may or may not be repealed, it's important to retain information on what energy communities are comprised of for future research/policy ideas at the state level.

* [data access page](https://edx.netl.doe.gov/dataset/ira-energy-community-data-layers). The two files we are interested in are: 
     * [MSA_NMSA_EC_FFE_v2024_1.zip](https://edx.netl.doe.gov/dataset/dbed5af6-7cf5-4a1f-89bc-a4c17e46256a/resource/13454403-ef6b-479b-b720-d5e3eaefbb91)
     * [Coal_Closures_EnergyComm_v2024_1.zip](https://edx.netl.doe.gov/dataset/dbed5af6-7cf5-4a1f-89bc-a4c17e46256a/resource/4006c9da-f99c-4731-97b2-633cc1578994)

#### Requirements for archiving

To be archived on Zenodo, a dataset must be:
- [x] published under an open license that permits reuse and redistribution
- [x] less than 50Gb in size (when zipped)
- [x] relevant to energy modelling and research

### Checklist for archive creation

Based on the [README documentation on creating a new archive](https://github.com/catalyst-cooperative/pudl-archiver#adding-a-new-dataset):

- [x] [Define the dataset's metadata](https://github.com/catalyst-cooperative/pudl-archiver#step-1-define-the-datasets-metadata)
- [x] [Implement archiver interface](https://github.com/catalyst-cooperative/pudl-archiver#step-2-implement-archiver-interface)
- [x] [Test archiver locally](https://github.com/catalyst-cooperative/pudl-archiver#step-3-test-archiver-locally)
- [x] [Test uploading to Zenodo](https://github.com/catalyst-cooperative/pudl-archiver#step-4-test-uploading-to-zenodo)
- [x] [Manually review archive before publication](https://github.com/catalyst-cooperative/pudl-archiver#step-5-manually-review-your-archive-before-publication)
[...]

PR #537, opened 2025-01-23
Initial work on DOE IRA archiver

pima-riemann:
# Overview

Closes #529 

What problem does this address?

What did you change in this PR?

# Testing

How did you make sure this worked? How can a reviewer verify this?

# To-do list

```[tasklist]
- [ ] add other TODO items here if necessary! questions that need to answered, decisions that need to be made, tests that need to be run, etc.
- [ ] Update relevant documentation - like comments, docstrings, README, release notes, etc.
- [ ] Review the PR yourself and call out any questions or issues you have
```

PR #545, opened 2025-01-24
Archive DoE IRA Energy Community Data

e-belfer:
# Overview

Closes https://github.com/catalyst-cooperative/pudl-archiver/issues/529. A continuation of #537.

* Adds an archiver for DOE IRA energy community layers.
* Only adds "PUDL Raw" prefix to a dataset title in Zenodo if it is part of PUDL's source dictionary. Previously this was only implemented for frictionless datapackages in #506, but we want the changes to be applied across both Zenodo and the datapackage.

Note that we generally don't like to hardcode but in this case:
* we don't expect the data to be updated
* the actual download links aren't available on the webpage. They're hidden behind a registration wall.
* there are only four of them.

* Add doeiraec.py
* Update `title` in `depositors.zenodo.entities.DepositionMetadata`

# Testing

See the published sandbox archive: https://sandbox.zenodo.org/records/157935?preview=1

# Questions for the reviewers
Taking your feedback/advice on how to best name the second partition and accordingly the file names - it's feeling clunky currently, but also it's a lot of information to convey!

# To-do list

```[tasklist]
- [x] once approved, make prod archive
- [x] add dois to yaml
- [x] add to GHA
```


src/pudl_archiver/archivers/doe/doeiraec.py
```diff
@@ -0,0 +1,35 @@
+"""Download DoE IRA Community Energy data."""
```
e-belfer, line 1: Oops, this should get removed! It's a leftover from the original PR.

src/pudl_archiver/archivers/doeiraec.py
```diff
@@ -0,0 +1,54 @@
+"""Download DoE IRA Energy Community data.
```
aesharpe, line 1: Little nit, but the o should be uppercase 

```suggestion
"""Download DOE IRA Energy Community data.
```

src/pudl_archiver/depositors/zenodo/entities.py
```diff
@@ -144,8 +144,13 @@ def from_data_source(cls, data_source_id: str) -> "DepositionMetadata":
...
             # Otherwise, specify that data was archived from the data_source.path
-            # and can be found there.
-            title = f"PUDL Raw {data_source.title}"
+            # and can be found there. Only include PUDL Raw if this dataset does
+            # indeed wind up in PUDL.
+            title = (
+                f"PUDL Raw {data_source.title}"
+                if data_source_id in SOURCES
+                else f"{data_source.title}"
+            )
```
aesharpe, line 153: Should the name be start with `PUDL raw` if we aren't sure whether it will end up in PUDL?

src/pudl_archiver/archivers/doe/doeiraec.py
```diff
@@ -0,0 +1,35 @@
...
+    AbstractDatasetArchiver,
+    ArchiveAwaitable,
+    ResourceInfo,
+)
+
+BASE_URL = "https://edx.netl.doe.gov/dataset/ira-energy-community-data-layers"
+
+
+class DOEIRAECArchiver(AbstractDatasetArchiver):
+    """DOE IRA EC archiver."""
```
aesharpe, line 13: Probably good to spell out Energy Community as in the other spots

```suggestion
    """DOE IRA Energy Community archiver."""
```

aesharpe:
> Taking your feedback/advice on how to best name the second partition and accordingly the file names - it's feeling clunky currently, but also it's a lot of information to convey!

Yeah, these names are tough, but also say what they need to say....if we did decide to truncate them, I think we could get away with nixing the `energy-community` part and just have one say

      `msa-nonmsa-fossil-fuel-employment-energy-community-status` -- > `fossil-fuel-employment-status`

      `coal-closures-energy-communities` --> `coal-closures`

src/pudl_archiver/depositors/zenodo/entities.py
e-belfer: No it shouldn't, which is what this PR is doing (removing the PUDL raw prefix)

src/pudl_archiver/archivers/doeiraec.py
aesharpe: oh haha ignore this RE: your comment above

src/pudl_archiver/depositors/zenodo/entities.py
aesharpe: mmmmm I see ok.

aesharpe (approved):

aesharpe:
Didn't get to test the archiver because my internet crapped out, but the code itself looks reasonable!

PR #551, opened 2025-01-28
Make possible to create archivers from GHA

e-belfer:
# Overview

Sometimes the barrier to testing an archiving is our internet connection. GHA resolves this nicely, but we currently can only use it to update existing archives. This extends the existing GHA workflow to allow for the creation of new sandbox and production archive drafts. This is low risk because they are drafts, not published directly! This also deprecates the use of the `conda-test`, which hasn't been updated in some time and isn't currently working as expected (it is impossible to update the conda environment).

Created a new workflow file that initializes new archivers (prod and sandbox).

# Testing

We'll need to merge the workflow file to able to test it.

# To-do list

```[tasklist]
```

### DIFF
--- a/.github/workflows/create-archiver.yml
+++ b/.github/workflows/create-archiver.yml
@@ -1,0 +1,122 @@
+---
+name: run-archiver
+
+on:
+  workflow_dispatch:
+    inputs:
+      datasets:
+        description: 'Comma-separated list of datasets to archive (e.g., "ferc2","ferc6").'
+        default: '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","epacems","ferc1","ferc2","ferc6","ferc60","ferc714","gridpathratoolkit","mshamines","nrelatb","phmsagas","usgsuspvdb","vcerare"'
+        required: true
+        type: string
+      sandbox:
+        description: "Run archiver on the sandbox server?"
+        default: false
+        required: true
+        type: boolean
+
+jobs:
+  archive-run:
+    defaults:
+      run:
+        shell: bash -l {0}
+    strategy:
+      matrix:
+        # Note that we can't pass global env variables to the matrix, so we manually reproduce the list of datasets here.
+        dataset: ${{ fromJSON(format('[{0}]', inputs.datasets || '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","epacems","ferc1","ferc2","ferc6","ferc60","ferc714","gridpathratoolkit","mshamines","nrelatb","phmsagas","usgsuspvdb","vcerare"' )) }}
+      fail-fast: false
+    runs-on: ubuntu-latest
+    permissions:
+      contents: "read"
+      id-token: "write"
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set default GCP credentials
+        id: gcloud-auth
+        continue-on-error: true
+        uses: "google-github-actions/auth@v2"
+        with:
+          workload_identity_provider: "projects/345950277072/locations/global/workloadIdentityPools/gh-actions-pool/providers/gh-actions-provider"
+          service_account: "pudl-sources@catalyst-cooperative-pudl.iam.gserviceaccount.com"
+
+      - name: Install Conda environment using mamba
+        uses: mamba-org/setup-micromamba@v2
+        with:
+          environment-file: environment.yml
+          cache-environment: true
+          condarc: |
+            channels:
+            - conda-forge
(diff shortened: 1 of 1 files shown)


.github/workflows/create-archiver.yml
```diff
@@ -0,0 +1,122 @@
+---
+name: run-archiver
```
zaneselvans, line 2: I think we need to change the name of the workflow so it doesn't clash with the existing `run-archiver` workflow.

I think renaming the YAML file and the workflow both to `initialize-archive` or `initialize-archiver` might be more intuitive, since the thing that's different here is that we're doing `pudl_archiver --initialize` right? Using `create` makes it seem like there's maybe something other than `--initialize` going on.

.github/workflows/create-archiver.yml
```diff
@@ -0,0 +1,122 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      datasets:
+        description: 'Comma-separated list of datasets to archive (e.g., "ferc2","ferc6").'
+        default: '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","epacems","ferc1","ferc2","ferc6","ferc60","ferc714","gridpathratoolkit","mshamines","nrelatb","phmsagas","usgsuspvdb","vcerare"'
```
zaneselvans, line 9: Since this action will create new instances of every archive that is passed in regardless of whether they exist or not already, maybe we should have an empty default, rather than an all datasets default, to avoid accidentally creating dozens of new drafts that hang around being confusing with different DOIs but the same names as pre-existing real archives?

.github/workflows/create-archiver.yml
```diff
@@ -0,0 +1,122 @@
...
+
+jobs:
+  archive-run:
+    defaults:
+      run:
+        shell: bash -l {0}
+    strategy:
+      matrix:
+        # Note that we can't pass global env variables to the matrix, so we manually reproduce the list of datasets here.
+        dataset: ${{ fromJSON(format('[{0}]', inputs.datasets || '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","epacems","ferc1","ferc2","ferc6","ferc60","ferc714","gridpathratoolkit","mshamines","nrelatb","phmsagas","usgsuspvdb","vcerare"' )) }}
```
zaneselvans, line 26: And similarly here, can we default to nothing as a precaution? We only expect to run this manually, and for a single archiver at a time right?

.github/workflows/create-archiver.yml
```diff
@@ -0,0 +1,122 @@
...
+          conda config --show
+          printenv | sort
+
+      - name: Create archive for ${{ matrix.dataset }}
+        env:
+          ZENODO_SANDBOX_TOKEN_UPLOAD: ${{ secrets.ZENODO_SANDBOX_TOKEN_UPLOAD }}
+          ZENODO_SANDBOX_TOKEN_PUBLISH: ${{ secrets.ZENODO_SANDBOX_TOKEN_PUBLISH }}
+          EPACEMS_API_KEY: ${{ secrets.EPACEMS_API_KEY }}
+          ZENODO_TOKEN_UPLOAD: ${{ secrets.ZENODO_TOKEN_UPLOAD }}
+          ZENODO_TOKEN_PUBLISH: ${{ secrets.ZENODO_TOKEN_PUBLISH }}
```
zaneselvans, line 67: It's a little funny that we give both the sandbox and production tokens to all the jobs, regardless of whether they are sandbox or production. Shouldn't they be defined so that they match the expected archive destination?

zaneselvans (changes requested):

.github/workflows/create-archiver.yml
e-belfer: Oops yes, this was just a copy-paste rush. Good catch.

.github/workflows/initialize-archiver.yml
```diff
@@ -1,17 +1,17 @@
...
   workflow_dispatch:
     inputs:
       datasets:
         description: 'Comma-separated list of datasets to archive (e.g., "ferc2","ferc6").'
-        default: '"eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiawater","eia_bulk_elec","epacamd_eia","epacems","ferc1","ferc2","ferc6","ferc60","ferc714","gridpathratoolkit","mshamines","nrelatb","phmsagas","usgsuspvdb","vcerare"'
+        default: ""
         required: true
         type: string
       sandbox:
         description: "Run archiver on the sandbox server?"
```
zaneselvans, line 13:
```suggestion
        description: "Initialize a new archive on the Zenodo sandbox server?"
```

zaneselvans (approved):

PR #563, opened 2025-01-30
Add archiver for NREL Standard Scenarios

### LINKED ISSUE #561: Write an archiver for  NREL Standard Scenarios
### Motivation and context:

"NREL's Standard Scenarios are a suite of forward-looking scenarios of the U.S. power sector that are updated annually to support and inform energy analysis. "

[Main page](https://www.nrel.gov/analysis/standard-scenarios.html)
The data is downloadable on [this scenario viewer](https://scenarioviewer.nrel.gov/)
[These are also archived on this atb archive page](https://atb.nrel.gov/archive) but these links go to the scenario viewer (but it could be a good way to access the years.

#### Requirements for archiving

To be archived on Zenodo, a dataset must be:
- [ ] published under an open license that permits reuse and redistribution
- [ ] less than 50Gb in size (when zipped)
- [ ] relevant to energy modelling and research

### Checklist for archive creation

Based on the [README documentation on creating a new archive](https://github.com/catalyst-cooperative/pudl-archiver#adding-a-new-dataset):

- [ ] [Define the dataset's metadata](https://github.com/catalyst-cooperative/pudl-archiver#step-1-define-the-datasets-metadata)
- [ ] [Implement archiver interface](https://github.com/catalyst-cooperative/pudl-archiver#step-2-implement-archiver-interface)
- [ ] [Test archiver locally](https://github.com/catalyst-cooperative/pudl-archiver#step-3-test-archiver-locally)
- [ ] [Test uploading to Zenodo](https://github.com/catalyst-cooperative/pudl-archiver#step-4-test-uploading-to-zenodo)
- [ ] [Manually review archive before publication](https://github.com/catalyst-cooperative/pudl-archiver#step-5-manually-review-your-archive-before-publication)
- [ ] [Finalize archive](https://github.com/catalyst-cooperative/pudl-archiver#step-6-finalizing-the-archive) (only core Catalyst developers can complete this step)
- [ ] [Automate archiving](https://github.com/catalyst-cooperative/pudl-archiver#step-7-automate-archiving)

### Links to published archives:
Include a link to the published sandbox archive for review.

krivard:
# Overview

Closes #561.

* NREL Standard Scenarios data is not available for download from static links, only from a javascript-powered web app
* We reverse engineered the API calls made by the web app and reproduce them here; they're basic GET and POST requests with no credentials necessary

* Add options for POST to `pudl_archiver.archivers.classes._download_file()` and `pudl_archiver.archivers.classes.AbstractDatasetArchiver.download_file()`
* Split hyperlink extraction in `pudl_archiver.archivers.classes.AbstractDatasetArchiver` into two phases to permit extraction of links directly from an HTML string
* New archiver for NREL Standard Scenarios

# Testing

Ran locally in fsspec and checked md5s against Zenodo.

# To-do list

```[tasklist]
- [x] Confirm raw license
- [x] Actually test it locally
- [ ] ~~Add some delays between metadata fetches so we're nice to the NREL API server (or else they might add credentialing and our archiver would break)~~ 
- [x] Reduce general chaos -- pull up URLs and regex patterns, make error messages make sense, etc
- [x] Wait for #538 to merge (or maybe rebase this branch to that one)
- [x] Update documentation
- [x] Self-review
```

### DIFF
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -284,7 +284,6 @@ class AbstractDatasetArchiver(ABC):
             headers: Additional headers to send in the GET request.
         """
         # Parse web page to get all hyperlinks
-        parser = _HyperlinkExtractor()
 
         response = await retry_async(
             self.session.get,
@@ -295,6 +294,25 @@ class AbstractDatasetArchiver(ABC):
             },
         )
         text = await retry_async(response.text)
+        return self.get_hyperlinks_from_text(text, filter_pattern)
+
+    def get_hyperlinks_from_text(
+        self,
+        text: str,
+        filter_pattern: typing.Pattern | None = None,
+    ) -> list[str]:
+        """Return all hyperlinks from HTML text.
+
+        This is a helper-helper function to perform very basic HTML-parsing functionality.
+        It extracts all hyperlinks from an HTML text, and returns those that match
+        a specified pattern. This means it can find all hyperlinks that look like
+        a download link to a single data resource.
+
+        Args:
+            text: text containing HTML.
+            filter_pattern: If present, only return links that contain pattern.
+        """
+        parser = _HyperlinkExtractor()
         parser.feed(text)
 
         # Filter to those that match filter_pattern
@@ -311,7 +329,7 @@ class AbstractDatasetArchiver(ABC):
         if not hyperlinks:
             self.logger.warning(
                 f"The archiver couldn't find any hyperlinks{('that match: ' + filter_pattern.pattern) if filter_pattern else ''}."
-                f"Make sure your filter_pattern is correct, check if the structure of the {url} page changed, or if you are missing HTTP headers."
+                f"Make sure your filter_pattern is correct, and check if the structure of the page is not what you expect it to be."
             )
 
         return hyperlinks
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -260,9 +260,11 @@ class AbstractDatasetArchiver(ABC):
         # immediately after they're safely stored in the ZIP
         download_path.unlink()
 
-    async def get_json(self, url: str, **kwargs) -> dict[str, str]:
+    async def get_json(self, url: str, post: bool = False, **kwargs) -> dict[str, str]:
         """Get a JSON and return it as a dictionary."""
-        response = await retry_async(self.session.get, args=[url], kwargs=kwargs)
+        response = await retry_async(
+            self.session.post if post else self.session.get, args=[url], kwargs=kwargs
+        )
         response_bytes = await retry_async(response.read)
         try:
             json_obj = json.loads(response_bytes.decode("utf-8"))
@@ -301,12 +303,13 @@ class AbstractDatasetArchiver(ABC):
             },
         )
         text = await retry_async(response.text)
-        return self.get_hyperlinks_from_text(text, filter_pattern)
+        return self.get_hyperlinks_from_text(text, filter_pattern, url)
 
     def get_hyperlinks_from_text(
         self,
         text: str,
         filter_pattern: typing.Pattern | None = None,
+        context: str = "text",
     ) -> list[str]:
         """Return all hyperlinks from HTML text.
 
@@ -335,7 +338,7 @@ class AbstractDatasetArchiver(ABC):
         # Warn if no links are found
         if not hyperlinks:
             self.logger.warning(
-                f"The archiver couldn't find any hyperlinks {('that match: ' + filter_pattern.pattern) if filter_pattern else ''}."
+                f"In {context}: the archiver couldn't find any hyperlinks {('that match: ' + filter_pattern.pattern) if filter_pattern else ''}."
                 f"Make sure your filter_pattern is correct, and check if the structure of the page is not what you expect it to be."
             )
 
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -321,6 +321,7 @@ class AbstractDatasetArchiver(ABC):
         Args:
             text: text containing HTML.
             filter_pattern: If present, only return links that contain pattern.
+            context: String used in error messages to describe what text was being searched.
         """
         parser = _HyperlinkExtractor()
         parser.feed(text)
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -84,9 +84,7 @@ async def _download_file(
     post: bool = False,
     **kwargs,
 ):
-    method = session.get
-    if post:
-        method = session.post
+    method = session.post if post else session.get
     async with method(url, **kwargs) as response:
         with file.open("wb") if isinstance(file, Path) else nullcontext(file) as f:
             async for chunk in response.content.iter_chunked(1024):
@@ -337,7 +335,7 @@ class AbstractDatasetArchiver(ABC):
         # Warn if no links are found
         if not hyperlinks:
             self.logger.warning(
-                f"The archiver couldn't find any hyperlinks{('that match: ' + filter_pattern.pattern) if filter_pattern else ''}."
+                f"The archiver couldn't find any hyperlinks {('that match: ' + filter_pattern.pattern) if filter_pattern else ''}."
                 f"Make sure your filter_pattern is correct, and check if the structure of the page is not what you expect it to be."
             )
 
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -78,9 +78,12 @@ class _HyperlinkExtractor(HTMLParser):
 
 
 async def _download_file(
-    session: aiohttp.ClientSession, url: str, file: Path | io.BytesIO, **kwargs
+    session: aiohttp.ClientSession, url: str, file: Path | io.BytesIO, post: bool=False, **kwargs
 ):
-    async with session.get(url, **kwargs) as response:
+    method = session.get
+    if post: 
+        method = session.post
+    async with method(url, **kwargs) as response:
         with file.open("wb") if isinstance(file, Path) else nullcontext(file) as f:
             async for chunk in response.content.iter_chunked(1024):
                 f.write(chunk)
@@ -181,7 +184,7 @@ class AbstractDatasetArchiver(ABC):
         # If it makes it here that means it couldn't download a valid zipfile
         raise RuntimeError(f"Failed to download valid zipfile from {url}")
 
-    async def download_file(self, url: str, file_path: Path | io.BytesIO, **kwargs):
+    async def download_file(self, url: str, file_path: Path | io.BytesIO, post: bool=False, **kwargs):
         """Download a file using async session manager.
 
         Args:
@@ -189,7 +192,7 @@ class AbstractDatasetArchiver(ABC):
             file_path: Local path to write file to disk or bytes object to save file in memory.
             kwargs: Key word args to pass to retry_async.
         """
-        await retry_async(_download_file, [self.session, url, file_path], kwargs)
+        await retry_async(_download_file, [self.session, url, file_path, post], kwargs)
 
     async def download_and_zip_file(
         self, url: str, filename: str, zip_path: Path, **kwargs
--- a/src/pudl_archiver/archivers/nrelss.py
+++ b/src/pudl_archiver/archivers/nrelss.py
@@ -112,7 +112,7 @@ class NrelStandardScenariosArchiver(AbstractDatasetArchiver):
                 uuid=project_uuid,
                 file_ids=[
                     (f["id"], f"NRELSS {project_year}  {f['scenario']}  {f['location_type']}.csv".replace(" ","_"))
-                    for f in file_list["files"] if f["file_type"] == "CSV"
+                    for f in file_list["files"] if (f["file_type"] == "CSV" or project_year == 2020)
                 ],
                 year=project_year
             )
@@ -129,7 +129,7 @@ class NrelStandardScenariosArchiver(AbstractDatasetArchiver):
         zip_path = self.download_directory / f"{self.name}-{year}.zip"
         data_paths_in_archive = set()
         # report
-        self.logger.info(f"Downloading report {report[0]} from {report[1]}")
+        self.logger.info(f"Downloading report {year} {report[0]} from {report[1]}")
         download_path = self.download_directory / report[0]
         await self.download_file(report[1], download_path)
         self.add_to_archive(
@@ -143,7 +143,7 @@ class NrelStandardScenariosArchiver(AbstractDatasetArchiver):
         download_path.unlink()
         
         for file_id,filename in file_ids:
-            self.logger.info(f"Downloading file {file_id} {uuid}")
+            self.logger.info(f"Downloading file {year} {file_id} {uuid}")
 #             file_resp = await retry_async(
 #                 self.session.post,
 #                 ["https://scenarioviewer.nrel.gov/api/download/"],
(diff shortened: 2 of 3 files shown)


src/pudl_archiver/metadata/sources.py
```diff
@@ -416,4 +416,40 @@
...
+        "working_partitions": {
+            "years": list(range(2016, 2025)),
+        },
+        "contributors": [
+            CONTRIBUTORS["catalyst-cooperative"],
+        ],
+        "keywords": sorted(
+            {
+                "nrel",
+                "standard scenarios",
```
krivard, line 449: Other keywords that could go in here, cribbed from `pudl/metadata/sources.py@nrelatb`:

```
                + KEYWORDS["us_govt"]
                + KEYWORDS["electricity"]
```

src/pudl_archiver/metadata/sources.py
```diff
@@ -416,4 +416,40 @@
...
+            }
+        ),
+        "license_raw": LICENSES["cc-by-4.0"],
```
krivard, line 452: They have a weirdo disclaimer that says (approximately) "you have to cite us but you can't make it look like we endorse you" which seems close enough to cc-by?

src/pudl_archiver/metadata/sources.py
```diff
@@ -416,4 +416,40 @@
...
+            "power sector that are updated annually to support and inform energy analysis."
+            "The Standard Scenarios are simulated using the Regional Energy Deployment System"
+            "and Distributed Generation Market Demand Model capacity expansion models and are"
+            "updated each year to provide timely information regarding power sector evolution."
+            "The scenarios have been designed to capture a range of possible power system"
+            "futures and consider a variety of factors from high vehicle electrification to"
+            "major cost declines for electricity generation technologies (e.g., using cost"
+            "inputs from the Annual Technology Baseline)."
+            "For select scenarios, the models are run using the PLEXOS software and the"
+            "Cambium tool that assembles structured data sets of hourly cost, emissions, and"
```
krivard, line 433: We could consider pulling in the cambium results as well (as a second partition) but they're like 6GB each year

src/pudl_archiver/metadata/sources.py
```diff
@@ -416,4 +416,40 @@
...
+            "operational data for modeled futures. Results are available using the Scenario"
+            "Viewer and Data Downloader."
+        ),
+        "source_file_dict": {
+            "source_format": "CSV",
+        },
+        "working_partitions": {
+            "years": list(range(2016, 2025)),
```
krivard, line 441: There seem to be two geo levels available as well, with one set of files for `Nation` and one set for `States`. Should we include that as a partition, or leave it at years?

src/pudl_archiver/archivers/classes.py
```diff
@@ -271,6 +270,25 @@ async def get_hyperlinks(
...
+        This is a helper-helper function to perform very basic HTML-parsing functionality.
+        It extracts all hyperlinks from an HTML text, and returns those that match
+        a specified pattern. This means it can find all hyperlinks that look like
+        a download link to a single data resource.
+
+        Args:
+            text: text containing HTML.
+            filter_pattern: If present, only return links that contain pattern.
+        """
+        parser = _HyperlinkExtractor()
```
krivard, line 291: I believe I have carefully sliced this to not conflict with [Marianne's `get_hyperlink` changes](https://github.com/catalyst-cooperative/pudl-archiver/pull/538/files) but I'll handle any massaging necessary if not

src/pudl_archiver/metadata/sources.py
cmgosnell: partitions are meant to be able to uniquely grab one file from the archive. so if the plan is to put both the nation and states for one year into one zip file then you can keep only the years are the partitions. without knowing how the archiver will work or how the files are structured my inclination would be to squish both nation and states into one annual zip.

src/pudl_archiver/metadata/sources.py
cmgosnell: If its simple to also add in the camdium results i'd say add em & add a second partition of `project` or `scenario_type` or something! but this seems like a lower priority than grabbing just the standard scenarios.

src/pudl_archiver/metadata/sources.py
cmgosnell: I would ask our resident license scrutinizer @zaneselvans  on this one!

src/pudl_archiver/metadata/sources.py
cmgosnell: are the camdium results 6 GB zipped? if so w/ the standard scenarios (assuming they are a similar size) its pushing up against the 50 GB archive limit

src/pudl_archiver/metadata/sources.py
krivard: Resolved offline: since not every report year has the same location options, and particularly because 2020 only has "ALL" as a location option, we're not going to partition by geography

src/pudl_archiver/metadata/sources.py
krivard: Yeah zipped. The standard scenarios are like two orders of magnitude smaller since they don't include hourly data though, so it's less "will cambium push this archiver over the limit" and more "can we archive cambium on zenodo at all"

I'll go with "not right now" and write up cambium as a separate issue

src/pudl_archiver/metadata/sources.py
zaneselvans: For data from NREL my trick for finding the license is to look and see whether they have added it to the [OEDI Data Swamp](https://data.openei.org) which requires whoever is archiving it to add an explicit, well-defined license.

The [2024 Standard Scenarios are there](https://data.openei.org/submissions/6453) and have been released under CC-BY-4.0.  I don't see any of the earlier years obviously though.

src/pudl_archiver/metadata/sources.py
zaneselvans: Tiny formatting issue -- on multi-line concatenated strings like this if we don't add an explicit space at the end of each line in the quotes, they'll run together.  Triple quoted strings are another option.

src/pudl_archiver/archivers/classes.py
```diff
@@ -311,7 +338,7 @@ async def get_hyperlinks(
         if not hyperlinks:
             self.logger.warning(
                 f"The archiver couldn't find any hyperlinks{('that match: ' + filter_pattern.pattern) if filter_pattern else ''}."
```
e-belfer, line 340:
```suggestion
                f"The archiver couldn't find any hyperlinks {('that match: ' + filter_pattern.pattern) if filter_pattern else ''}."
```
Non-blocking: while we're here we can fix a spacing issue.

src/pudl_archiver/archivers/classes.py
```diff
@@ -311,7 +338,7 @@ async def get_hyperlinks(
...
-                f"Make sure your filter_pattern is correct, check if the structure of the {url} page changed, or if you are missing HTTP headers."
+                f"Make sure your filter_pattern is correct, and check if the structure of the page is not what you expect it to be."
```
e-belfer, line 341: I actually find having the URL here very useful for debugging what site is causing the problem in a context where I might be seeing one of these errors and hitting 100 sites, so much so that I'm almost inclined to ask you to pass URL as a parameter here so we can identify where the problem is coming from.

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+
+REPORT_URL_PATTERN = re.compile(
+    r"https://www.nrel.gov/docs/(?P<fy>fy\d{2}osti)/(?P<number>\d{5}\.pdf)"
+)
+
+
+class AbstractNrelScenarioArchiver(AbstractDatasetArchiver):
+    """Base class for archiving projects from the NREL Scenario Viewer."""
+
+    project_year_pattern: re.Pattern
```
e-belfer, line 24: Can you add docstrings to these parameters so we know what they mean?

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+            Tuple of:
+                - Project UUID
+                - Project year
+                - List of (filename, url) pairs for PDF reports
+                - List of (filename, id) pairs for files
+        """
+        project_uuid = scenario_project["uuid"]
+        m = self.project_year_pattern.search(scenario_project["name"])
+        if not m:
+            return (False,) * 3
```
e-belfer, line 57: The tuple has four items in it, is there a reason this only has 3? Also would be helpful to comment here to explain what you're doing and why is was necessary to handle this exception.

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+        report_links = self.get_hyperlinks_from_text(
+            scenario_project[self.report_section], REPORT_URL_PATTERN
+        )
+        if not report_links:
+            report_links = {self.handle_report_link_exceptions(project_year)}
+        if not report_links:
+            raise AssertionError(
+                f"We expect all projects to have a {self.report_section} with a link to the report PDF, but {project_year} does not:\n"
+                f"{scenario_project}"
+            )
```
e-belfer, line 69: Is the second `if not` a leftover here? I'm not sure why you're defining report_links if it doesn't exist and then looking for a non-existence report_links again.

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+        report_data = []
+        for report_link in report_links:
+            m = REPORT_URL_PATTERN.search(report_link)
+            if not m:
+                raise AssertionError(
+                    f"Bad report link {report_link} found in {scenario_project['name']} {project_uuid}: {scenario_project}"
+                )
+            report_data.append((f"{m.group('fy')}_{m.group('number')}", report_link))
```
e-belfer, line 77: Can you add a docstring here explaining what you're doing?

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+
+        file_list_resp = await retry_async(
```
e-belfer, line 79: Non-blocking: There is a `self.get_json()` method you could call from classes instead of reproducing it here.

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+            kwargs={"data": {"project_uuid": project_uuid}},
+        )
+        file_list = await retry_async(file_list_resp.json)
+
+        return (
+            project_uuid,
+            project_year,
+            report_data,
+            [
+                (
```
e-belfer, line 91: Can you add a docstring here explaining what's going on and giving an example of a filename change?

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+                    f"{self.name.upper()} {project_year}  {'  '.join(f[x] for x in self.file_naming_order)}.{f['file_type']}".replace(
+                        " ", "_"
```
e-belfer, line 93: Is there a reason not to just join these with underscores to begin with, rather than joining with spaces and then replacing spaces with _?

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+        for scenario_project in (
+            p for p in project_records if p["name"].startswith(self.project_startswith)
+        ):
+            (
+                project_uuid,
+                project_year,
+                report_data,
+                file_ids,
+            ) = await self.collect_project_info(scenario_project)
+            if project_uuid:
```
e-belfer, line 122: In what scenario would there not be one?

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+                )
+
+    async def get_project_resource(
+        self,
+        uuid: str,
+        year: int,
+        reports: list[tuple[str, str]],
+        file_ids: list[tuple[str, str]],
+    ) -> ResourceInfo:
+        """Download all available data for a project.
```
e-belfer, line 137: Might be helpful to note that a project in NREL terminology is one year of standard scenario data in ours.

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+                filename=filename,
+                blob=download_path.open("rb"),
+            )
+            # Don't want to leave multiple giant files on disk, so delete
+            # immediately after they're safely stored in the ZIP
+            download_path.unlink()
+
+        return ResourceInfo(
+            local_path=zip_path,
+            partitions={"years": year},
```
e-belfer, line 174: Were you able to test whether re-enabling the ZipLayout validation worked here? I remember you were having trouble with it before and we disabled it for expediency, but ideally we would return to using it if the files make this possible. If it's not possible, we should note somewhere in this archiver why we aren't using this.

src/pudl_archiver/archivers/nrelss.py
```diff
@@ -0,0 +1,194 @@
...
+    """Base class for archiving projects from the NREL Scenario Viewer."""
+
+    project_year_pattern: re.Pattern
+    project_startswith: str
+    report_section: str
+    file_naming_order: tuple[str]
+
+    def handle_report_link_exceptions(self, project_year: str) -> str:
+        """Provide hard-coded exceptions for years where the normal report URL locator doesn't work."""
+        return ""
```
e-belfer, line 31: Non-blocking: Do you expect many years to not have a report? Right now this just silently ignores the problem which seems ok if this is uneven across years, but if this is undesired behavior we want to catch another option would be to return a `NotImplementedError`, which I think would more accurately reflect how you're trying to use this in the Abstract class. That way you would define the default cambium behavior for this in the cambium archiver file, rather than here.

e-belfer (changes requested):
All the changes in the classes.py file look great and reasonable to me. I have one request re: a change in the logging for get_hyperlinks_from_text(), some requests for additional docstrings and a few non-blocking questions and changes. Think this should be mergeable with very minor updates!

src/pudl_archiver/archivers/classes.py
krivard: fair, will add

src/pudl_archiver/archivers/nrelss.py
krivard: it doesn't handle post; should I add POST handling to self.get_json?

src/pudl_archiver/archivers/nrelss.py
krivard: elements in the list contain spaces as well

src/pudl_archiver/archivers/nrelss.py
krivard: the CSV files have a large header that we don't have to way to tell pandas about in the ZipLayout validation checks. I'll add a comment about that.

src/pudl_archiver/archivers/nrelss.py
e-belfer: It'd be consistent with the other methods and I would definitely use it in the NREL siting archiver where I need to POST to get a JSON as well, so why not!

src/pudl_archiver/archivers/nrelss.py
e-belfer: As long as you aren't introducing double spaces into the names through this, that seems ok to me!

krivard:
One note: The archive on Zenodo was run with a version of the code that only looked at files with .CSV extensions. This version accepts all files. 

This causes one change to the 2022 archive, which in this version includes a file `nrelss_2022__all__all.zip` which is not present in the archive on Zenodo. The added file is 7.9 MB and is a compressed version of all the CSVs in 2022.

No other years are affected.

It is easiest to keep accepting all files, so that we don't have to add special case handling for Cambium... but if we want the outputs to exactly match what's already up there, then we should add the logic to drop the extra file.

e-belfer (approved):
Two more minor and non-blocking concerns, but the addition of the zipfile and erring on the side of grabbing all files seems totally fine to me

PR #573, opened 2025-02-03
Switch to a new prettier mirror and bump ruff pre-commit hook.

zaneselvans:
# Overview

Switch to using a different mirror of the `prettier` pre-commit hook that is actually being maintained.

# Testing

I ran the `prettier` pre-commit hook on all files locally and it didn't do anything unexpected.


e-belfer (approved):
Good catch, thanks for making the update.

PR #578, opened 2025-02-04
Fix EIA MECS on GHA

cmgosnell:
# Overview

Closes #XXXX.

eia mecs runs fine locally but has been getting access denied errors on the GHA. I also added the same little "check if this file we downloaded is actually an html redirect"

i added a header mirrored off of @krivard 's situation for doelead. 

# Testing

i ran the archive via [gha here](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/13142537392/job/36672871662)
which made a draft [archive here](https://zenodo.org/uploads/14784586)

# To-do list

```[tasklist]
```


e-belfer (changes requested):
Mostly this looks great. There's one link that is marked as skipped but when I click on it doesn't redirect as expected: https://www.eia.gov/consumption/manufacturing/data/2006/xls/RSETable3_2.xls

And indeed that table is missing from the zipped file. This seems undesirable. Any thoughts on why this is happening?

cmgosnell:
at first i couldn't figure it out and now I see our culprit.... one extra space:
image

cmgosnell:
okay @e-belfer  amazing catch! i believe it is now fixed here is the [draft archive](https://zenodo.org/uploads/14827265)

src/pudl_archiver/archivers/eia/eiamecs.py
```diff
@@ -72,7 +74,9 @@ async def get_year_resources(self, year: int) -> list[ResourceInfo]:
 
         # Loop through all download links for tables
         for table_link in await self.get_hyperlinks(year_url, table_link_pattern):
-            table_link = f"{year_url}/{table_link}"
+            # strip bc there is at least one file which has a space at the end of
+            # the href link
+            table_link = f"{year_url}/{table_link}".strip()
```
cmgosnell, line 79: this is what i changed re the one missing file bc " "

e-belfer (changes requested):
Ugh one more. Looks like 2002 has both RSE and rse in the table name, but rse isn't getting captured leading to duplicate file names in the run: https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/13245735503/job/36971681614

Should be a case of updating the pattern in `TABLE_LINK_PATTERNS`.

cmgosnell:
wow again amazing catch! (which is to say good job actually looking at the warnings!)

I added `re.IGNORECASE` and added the lower case version into the lil `rse_map`. And apparently the production server works on this archive bc i was able to make a new draft:

https://zenodo.org/uploads/14827265

PR #580, opened 2025-02-05
Make server selection on `initialize-archiver` clearer

e-belfer:
# Overview

Makes the choice of server more obvious in `initialize-archiver`

- Switch from boolean to choice in github action

# Testing

Run the action on this branch for both production and sandbox servers.
Sandbox: https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/13168342804/job/36753639628
Prod: https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/13168368880/job/36753714104

# To-do list

```[tasklist]
```

### DIFF
--- a/.github/workflows/initialize-archiver.yml
+++ b/.github/workflows/initialize-archiver.yml
@@ -9,11 +9,13 @@ on:
         default: ""
         required: true
         type: string
-      sandbox:
-        description: "Initialize a new archive on the Zenodo sandbox server?"
-        default: true
+      server:
+        description: "Which Zenodo server would you like to run on?"
+        options:
+          - sandbox
+          - production
         required: true
-        type: boolean
+        type: choice
 
 jobs:
   archive-run:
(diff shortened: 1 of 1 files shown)


.github/workflows/initialize-archiver.yml
```diff
@@ -9,11 +9,13 @@ on:
...
         required: true
         type: string
-      sandbox:
-        description: "Initialize a new archive on the Zenodo sandbox server?"
-        default: true
+      server:
+        description: "Which Zenodo server would you like to run on?"
+        options:
+          - sandbox
+          - production
```
zaneselvans, line 16: Can we have it default to `sandbox` to avoid proliferating drafts on production unnecessarily? It looks like [an available setting](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs).

```suggestion
      server:
        description: "Which Zenodo server would you like to run on?"
        default: 'sandbox'
        options:
          - sandbox
          - production
```

zaneselvans (approved):
Non-blocking suggestion to use `sandbox` as the default.

PR #592, opened 2025-02-24
Add blank GitHub Action to facilitate testing Cambium

krivard:
# Overview

We need a blank GHA in `main` to be able to run the complete version in the `cambium` branch to see if it works. We're leery of modifying the primary `run-archiver` action.

New GHA: `run-archiver-cambium`

# Testing

Not intended to work as-is! just be parseable so it shows up in Actions.

# To-do list


.github/workflows/run-archiver-cambium.yml
```diff
@@ -0,0 +1,17 @@
...
+  archive-run:
+    defaults:
+      run:
+        shell: bash -l {0}
+    runs-on: ubuntu-latest-4core
+    permissions:
+      contents: "read"
+      id-token: "write"
+    steps:
+      - uses: actions/checkout@v4
```
krivard, line 17:
```suggestion
      - uses: actions/checkout@v4

```

krivard:
Pre-commits fail because this branch doesn't have a `.github/actionlint.yml` to validate the runner name against. `cambium` does, and the pre-commits for the GHA succeed there, so I think we're OK here.

krivard:
Alternately I can cherry-pick all the GHA commits from https://github.com/catalyst-cooperative/pudl-archiver/pull/569 as-is and fully-fleshed-out, lmk

.github/workflows/run-archiver-large.yml
```diff
@@ -0,0 +1,17 @@
+---
+name: run-archiver-cambium
```
krivard, line 2:
```suggestion
name: run-archiver-large
```

jdangerx (approved):
approved pending actionlint fix -_-

PR #610, opened 2025-03-12
Fix EIA MECS DOIs

e-belfer:
# Overview

Responds to problems identified in March 10th run of #594: see https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/13760099248

- Fix DOIs for MECS to point at concept DOI, not first record. This was updating the archive against v1.0 each time the record was run, rather than successfully updating as compared to the latest record.

# Testing

# To-do list

```[tasklist]
```


krivard (approved):
Confirmed new DOIs resolve to latest, not toplevel, record 👍

PR #626, opened 2025-04-04
Add archiver for raw extracted sec 10k data

zschira:
# Overview

Adds an archiver that reads `deltalake` versions of extracted SEC 10k tables, then saves each table as a monolithic parquet file within the archive.

### DIFF
--- a/src/pudl_archiver/archivers/sec10k.py
+++ b/src/pudl_archiver/archivers/sec10k.py
@@ -37,5 +37,5 @@ class Sec10kArchiver(AbstractDatasetArchiver):
 
         return ResourceInfo(
             local_path=download_path,
-            partitions={"table_name": raw_name},
+            partitions={"table": raw_name},
         )
--- a/src/pudl_archiver/archivers/sec10k.py
+++ b/src/pudl_archiver/archivers/sec10k.py
@@ -1,0 +1,50 @@
+"""Download SEC10k extracted tables for arhival."""
+
+import deltalake
+import pandas as pd
+
+from pudl_archiver.archivers.classes import (
+    AbstractDatasetArchiver,
+    ArchiveAwaitable,
+    ResourceInfo,
+)
+
+# This can be removed once we change the upstream table names to reflect their raw status
+TABLE_NAME_MAP = {
+    "core_sec10k__filings": "raw_sec10k__quarterly_filings",
+    "core_sec10k__company_information": "raw_sec10k__quarterly_company_information",
+    "out_sec10k__parents_and_subsidiaries": "raw_sec10k__parents_and_subsidiaries",
+    "core_sec10k__exhibit_21_company_ownership": "raw_sec10k__exhibit_21_company_ownership",
+}
+
+
+def _year_quarters_from_dataframe(df: pd.DataFrame) -> dict:
+    return {"year_quarters": set(df["year_quarter"].unique())}
+
+
+class Sec10kArchiver(AbstractDatasetArchiver):
+    """Sec10k raw extracted archiver."""
+
+    name = "sec10k"
+    deltalake_version = 0
+
+    async def get_resources(self) -> ArchiveAwaitable:
+        """Archive monolithic parquet files for each raw SEC 10k table."""
+        for delta_name, raw_name in TABLE_NAME_MAP.items():
+            yield self.get_delta_table(delta_name, raw_name)
+
+    async def get_delta_table(self, delta_name: str, raw_name: str):
+        """Read configured version of table from deltalake on GCS and save parquet."""
+        table_url = f"gs://model-outputs.catalyst.coop/sec10k/{delta_name}"
+        download_path = self.download_directory / f"{raw_name}.parquet"
+        dt = deltalake.DeltaTable(table_url, version=self.deltalake_version)
+        df = dt.to_pandas()
+        df.to_parquet(download_path)
+        date_parts = {}
+        if delta_name != "out_sec10k__parents_and_subsidiaries":
+            date_parts = _year_quarters_from_dataframe(df)
+
+        return ResourceInfo(
+            local_path=download_path,
+            partitions={"table_name": raw_name} | date_parts,
+        )
--- a/src/pudl_archiver/archivers/sec10k.py
+++ b/src/pudl_archiver/archivers/sec10k.py
@@ -1,7 +1,6 @@
 """Download SEC10k extracted tables for arhival."""
 
 import deltalake
-import pandas as pd
 
 from pudl_archiver.archivers.classes import (
     AbstractDatasetArchiver,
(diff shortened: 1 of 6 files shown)


src/pudl_archiver/archivers/sec10k.py
```diff
@@ -0,0 +1,50 @@
...
+
+def _year_quarters_from_dataframe(df: pd.DataFrame) -> dict:
+    return {"year_quarters": set(df["year_quarter"].unique())}
+
+
+class Sec10kArchiver(AbstractDatasetArchiver):
+    """Sec10k raw extracted archiver."""
+
+    name = "sec10k"
+    deltalake_version = 0
```
zaneselvans, line 29: Does version 0 mean "the most recent version" or "the first version"? In general the archiver should be able to detect and download a new version without anybody needing to modify it.

src/pudl_archiver/archivers/sec10k.py
```diff
@@ -0,0 +1,50 @@
...
+        dt = deltalake.DeltaTable(table_url, version=self.deltalake_version)
+        df = dt.to_pandas()
+        df.to_parquet(download_path)
+        date_parts = {}
+        if delta_name != "out_sec10k__parents_and_subsidiaries":
+            date_parts = _year_quarters_from_dataframe(df)
+
+        return ResourceInfo(
+            local_path=download_path,
+            partitions={"table_name": raw_name} | date_parts,
```
zaneselvans, line 49: Do we really want to partition by table? Most of our data sources have many tables contained within a given partition -- whether they're stored in different spreadsheets, FERC database tables, etc.  Why not just partition by date and have each of the 4 tables available within each year partition?

The `out_sec10k__parents_and_subsidiaries` table contains a `report_date` column. Should that not be the basis for partitioning that table? Otherwise do you just always get all of that table?

zaneselvans: If we include partitions here that don't mean anything, I think the datastore infrastructure won't work correctly.  Looking at the FERC-714 bulk-CSV archiver (for 2006-2020 data) it sets the partitions to an empty dictionary.

src/pudl_archiver/archivers/sec10k.py
zschira: Version 0 is just the first version, and I'm a bit nervous about having it just grab the latest version because running the EX21 extraction model is very time consuming and will sometimes timeout and error, requiring manual retries on certain partitions, so I think that at this point it's good to have some human interaction to determine when a new version of data is actually ready for use downstream.

When I make updates to the upstream data, I can make sure to leave detailed documentation for the steps required to make a new version of the data, including updating the archiver.

src/pudl_archiver/archivers/sec10k.py
zschira: My thinking here was that in most cases when we have many tables within a single partition, the raw data is zipped together into a single '`Resource`', whereas here each table has it's own parquet file and is it's own `Resource`.

I guess if we wanted to make this more consistent with other datasets and have all the tables bundled in one `Resource` we could either zip the parquet files together or load the data into a duckdb file.

zaneselvans: Okay, I can see the per-file (table) partitions being convenient in this case even if it's different from almost all of our other archivers.  But what about the date component?

src/pudl_archiver/archivers/sec10k.py
zaneselvans: This sounds brittle. If the process for generating a new version can't complete reliably maybe it makes sense not to write it directly to the archiver pick up point (which is the current arrangement?)

Do we happen to know whether reading a given version out of DeltaLake is going to be bit-for-bit the same every time? If not then every time we run the archiver it'll detect changes and create a new version on Zenodo.

zschira: > This sounds brittle. If the process for generating a new version can't complete reliably maybe it makes sense not to write it directly to the archiver pick up point

Yeah right now we write to a 'workspace' when actually running the model, then mirror that data to the deltalake tables once we've completed a run, to provide static versioning. Unfortunately it's hard to have a perfectly reproducible process when running the model takes multiple days and is fairly costly to start over if something goes wrong at any point.

> Do we happen to know whether reading a given version out of DeltaLake is going to be bit-for-bit the same every time? If not then every time we run the archiver it'll detect changes and create a new version on Zenodo.

Yeah `deltalake` handles versioning by associating specific parquet file(s) with specific versions of a table, but the old parquet files are never deleted, so reading from a specific version should be completely stable.

src/pudl_archiver/archivers/sec10k.py
zschira: Fixed so every table now has `years` partitions

src/pudl_archiver/archivers/sec10k.py
zaneselvans: It sounds like there's some unavoidable babysitting that has to happen on the extraction side, and a manual decision that a new version is ready to be deployed, and we can put all of the manual steps on the extraction side and allow the archiver to run without human intervention, as is the expectation with all of the other archivers -- so long as a human is taking care to check that the new version is ready, and deploy it from the workspace to the delta lake / pickup side.

That sounds better to me than having to both manually watch the extraction, and then also add a manual step on the archiving side.

src/pudl_archiver/archivers/sec10k.py
zaneselvans: I though we were trying to get rid of the date partitions entirely and treat it monolithically like the older FERC-714 data. if there are date partitions here then the datastore will expect to be able to pull them individually but they won't exist.

src/pudl_archiver/archivers/sec10k.py
zschira: That sounds reasonable. By default `deltalake` will read the latest state of  a table, so I've removed the version altogether.

src/pudl_archiver/archivers/sec10k.py
zschira: Oh right. I had made it so there was a list of all years associated with each table, but that's probably unnecessarily complex and confusing, so I've removed them entirely.

zaneselvans: Okay great -- so it'll be partitioned by table, and we'll have years (or quarters) separately defined in the ETL Settings object, which can be used to select only a subset of the data for processing in CI, etc.

zaneselvans: I think you need to make the analogous changes to the data source in PUDL for this to work though.

zaneselvans: Also if the cached environment is screwing things up I think you can temporarily tell it not to do that in the GHA workflow file, and then it'll build it from scratch?

zaneselvans (approved):

PR #636, opened 2025-04-22
smol xbrl rssfeed fix (sorted was converting rssfeed dictionary to a list)

cmgosnell:
# Overview

in the standard monthly archiving process ella published a new ferc1 zenodo archive today. we decided to try to integrate it into pudl to work out integration kinks even though rn it looks like ~80% of the utilities have responded. and I’ve found the first  road block, but then i think it's fixed.

it looks like in the past archives the `rssfeed` files for each year are dictionaries. in this most recent archive the `rssfeed`  is just a list of respondents. it looks like this got changed ~7 months ago w/ a small fix. But the dictionary was being `sorted` which converted this into a list of just the filer names instead of the full dictionary.

kept sorting the dictionary but just used `sorted` to sort all of the keys.

Questions?
- was this change intentional somehow in someway that I'm missing??
- is there anything downstream (except me trying to run the ferc1_xbrl -> sqlite conversion) that I should look out for?

# Testing

i ran:
`pudl_archiver --datasets ferc1`
which gave me this draft:
https://zenodo.org/uploads/15264594
and i checked several of the xbrl years and all the `rssfeed` docs are all dictionaries!

# To-do list

```[tasklist]
```

### DIFF
--- a/src/pudl_archiver/archivers/ferc/xbrl.py
+++ b/src/pudl_archiver/archivers/ferc/xbrl.py
@@ -360,7 +360,16 @@ async def archive_year(
         # Save snapshot of RSS feed
         with archive.open("rssfeed", "w") as f:
             logger.info("Writing rss feed metadata to archive.")
-            f.write(json.dumps(sorted(metadata), default=str, indent=2).encode("utf-8"))
+            f.write(
+                json.dumps(
+                    {
+                        filing_name: metadata[filing_name]
+                        for filing_name in sorted(metadata)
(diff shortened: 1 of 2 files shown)


zaneselvans:
It looks like the unit tests are failing because they currently expect to see the list that is being generated by `sorted()` rather than a dictionary, so it seems like there was at least a little bit of interaction with this change when it happened. @zschira do you have any additional context you can share about why this change was made originally?

PR #645, opened 2025-04-29
Archive 2014-2018 VCE RARE data

e-belfer:
# Overview

Archives data needed for https://github.com/catalyst-cooperative/pudl/issues/4226.

- We have new 2014-2018 VCE RARE data - we need to archive it so we can publish it in PUDL.

- [x] Update data in GCS bucket, creating a new version folder to avoid overwriting changes to README and other files
- [x] Update the VCE RARE archiver to be able to point at a version folder, rather than storing all files from all versions in the same folder

# Testing

`pudl_archiver --dataset vcerare --sandbox --summary-file vcerare.json`
See also: https://sandbox.zenodo.org/records/203310

# To-do list

```[tasklist]
```

### DIFF
--- a/src/pudl_archiver/archivers/vcerare.py
+++ b/src/pudl_archiver/archivers/vcerare.py
@@ -22,11 +22,14 @@ class VCERAREArchiver(AbstractDatasetArchiver):
 
     name = "vcerare"
     bucket_name = "sources.catalyst.coop"
+    version = "v2"  # Version of the files to archive
 
     async def get_resources(self) -> ArchiveAwaitable:
         """Download VCE RARE resources."""
         bucket = storage.Client().get_bucket(self.bucket_name)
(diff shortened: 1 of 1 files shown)


cmgosnell (approved):
this all looks straightforward and the sandbox archive looks good. just to verify that i know whats going on:
- we manually upload these files on gcs
- we (probably you in this case) added v{n} folders within this vcerare bucket folder.
- if there are future updates presumably we'll also manually add a new version.
- if storing lots of versions on gcs gets to be too much we'll have the old zenodo archives and the gcs side of things is all manual or at least one-time uploads so its in our control.

zaneselvans:
IIRC, I think we may have edited the README in the first version before archiving it, since it had some crazy formatting and also said some untrue things. We might want to check the README this time around to see if it needs cleaning up.

Issue #663, opened 2025-05-12
eiaaeo: DOI is not concept DOI, causes updates to be made as compared to v1, not the latest version


e-belfer:
The archiver is always marked as creating the AEO 2025 file. This is because the DOI the archiver is comparing to is not the concept DOI! I also had to manually update the last published archive (v2.0.0) to v3.0.0 because v2.0.0 already existed. This should be a quick and straightforward fix.

zaneselvans:
Ah, I wonder if we used v1.0 instead of the concept DOI because Zenodo was claiming the original one had been deleted. This seems vaguely familiar.

PR #674, opened 2025-05-20
Fix bugs in PCAP and AEO archivers

### LINKED ISSUE #651: epapcap - Maricopa PDFs causing problems
Two PDFs in the PCAPs aren't downloading correctly:
https://azmag.gov/Portals/0/Environmental/CPRG/Maricopa-Pinal-County-Region-Priority-Climate-Action-Plan.pdf?ver=8G2HlJqwTD7IwdHfyGUlYA%3d%3d

and 
https://azmag.gov/Portals/0/Environmental/CPRG/Maricopa-Pinal-County-Region-Priority-Climate-Action-Plan.pdf?ver=8G2HlJqwTD7IwdHfyGUlYA%3d%3d

They are downloading as:
Maricopa-Pinal-County-Region-Priority-Climate-Action-Plan.pdf?ver=8G2HlJqwTD7IwdHfyGUlYA%3d%3d
and
Appendix-A-Maricopa-Pinal-County-Region-2020-GHG-EI-TSD.pdf?ver=WuCZG4HmHkVRQS7QDMwqWg%3d%3d

The extension after the .pdf is likely causing problems with the download, as these are unopenable files. We should strip the extension when saving.

e-belfer:
# Overview

Closes #651 and #663 

- AEO archives are always running compared to v1
- PCAP archives are not valid
- We've started using user-agent rotation across our archivers, but we do this slightly differently everywhere

- Updated AEO DOI to concept DOI
- Updated PCAP archiver to sanitize filenames
- Added a `get_user_agent` method to the classes, which picks a random user agent from a set list. This method could be improved over time, but it's equivalent to the most sophisticated version of rotation we currently have implemented.

# Testing

AEO:
https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/15143644378
As expected, we now see no change in EIA AEO.

PCAP:
When diff'ing the EPA PCAP from this run with the last published version, I see:

All other affected archivers: https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/15164066543

# To-do list

```[tasklist]
```

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -15,9 +15,8 @@ environments:
       - conda: https://conda.anaconda.org/conda-forge/noarch/ca-certificates-2025.4.26-hbd8a1cb_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/cyrus-sasl-2.1.27-h54b06d7_7.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/freexl-2.0.0-h9dce30a_2.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/gdal-3.10.3-py312h546fd74_9.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/gdal-3.11.0-py312h546fd74_4.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/geos-3.13.1-h97f6797_0.conda
-      - conda: https://conda.anaconda.org/conda-forge/linux-64/geotiff-1.7.4-h239500f_2.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/giflib-5.2.2-hd590300_0.conda
       - conda: https://conda.anaconda.org/conda-forge/linux-64/icu-75.1-he02047a_0.conda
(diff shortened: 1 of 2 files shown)


src/pudl_archiver/archivers/classes.py
```diff
@@ -33,6 +34,19 @@
 }
 VALID_PARTITION_RANGES: dict[str, str] = {"year_quarter": "QS", "year_month": "MS"}
 
+USER_AGENTS: list[str] = [
```
jdangerx, line 37: registering slight approval of this rotation approach in a github comment for posterity

cmgosnell: @e-belfer can you explain a bit about how you compiled these guys? in case someone in the future wants to add more or change em?

cmgosnell (approved):
I'm approving bc this looks chill and good even though i'm requesting a smol amount of documentation to be fleshed out for the user agent change. it seems like this is a consolidation/standardization of previous techniques that 

I've run a these few archivers locally and they look good! I got a failure for the 176 archive on the sandbox but it worked fine on prod.

Issue #630, opened 2025-04-15
Archives regularly incorrectly flagged as updated


e-belfer:
**Goal**: When an archiver runs, if the underlying data doesn't change the archive should _not_ be marked as updated.

Each month, the following archives are flagged as changed/updated despite no changes occurring:

- eiarecs - zipfile hash changes, file size does not. zipinfo file order changes
- usgsuspvdb
- usgsuswtdb
- all FERC archives - entry_id in RSS metadata is changing run to run

This looks like the following in the run summary (see e.g. [a recent run](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/14441094141)):

> {
>         "name": "eiarecs-2020.zip",
>         "diff_type": "UPDATE",
>         "size_diff": 0,
>         "partition_changes": []
>       }

This causes a small amount of time and energy to spent manually verifying that these archives do not need to be updated each time the archives are reviewed, and we would benefit from addressing the problem at the source!

With the recent merge of #636 we also now see <10 byte changes in the XBRL partitions, which most likely does not indicate actual changes in the underlying data.

Success metrics: Running an archiver and publishing to sandbox, then immediately re-running the same archiver on sandbox, we should see no change in the archive.

- [x] eiarecs #701 
- [x] #730
- [ ] usgsuspvdb - check after #654 fixed
- [ ] usgsuswtdb - check after #684 fixed

PR #701, opened 2025-06-16
Fix EIA RECS change detection

e-belfer:
# Overview

Addresses part of #630.

- EIA RECS always marks files as having changed - upon examination of the files, it's because the order of the zipped files changes from run to run, creating a zipfile with a different hash.

- Sort the order of adding files to the ZipFile, to ensure consistency across archives when the zipped files aren't actually changing.

# Testing

Run the EIA RECS sandbox archive twice back to back. The second archive should get marked as "no changes".

# To-do list

```[tasklist]
```


krivard (approved):
🚢 Looks legit!

PR #705, opened 2025-06-16
Remove dependabot updates for main, move from update to upgrade in `update-lockfiles.yml`

e-belfer:
# Overview

Closes #XXXX.

- Pixi and dependabot are both trying to update the same packages
- `pixi update` is resulting in errors, while `pixi upgrade` works just fine 

- Move from pixi update to pixi upgrade so that the package bounds in pyproject.toml get updated over time
- Drop dependabot updates to packages in main
- Update packages  

Hmm....
It turns out `pixi upgrade` updates bounds on the main dependencies, but doesn't handle the `optional-dependencies` packages as expected. Our options are:
- Add these to the main dependencies
- Reconfigure these as pixi environments
- Remove all package bounds on these environments, and just let pixi solve them

# Testing

# To-do list

```[tasklist]
```

### DIFF
--- a/src/pudl_archiver/archivers/eia/eiarecs.py
+++ b/src/pudl_archiver/archivers/eia/eiarecs.py
@@ -93,7 +93,7 @@ class EiaRECSArchiver(AbstractDatasetArchiver):
             local_path=zip_path,
             partitions={"year": year},
             layout=ZipLayout(
-                file_paths=paths_within_archive + original_forms_within_archive
+                file_paths=sorted(paths_within_archive + original_forms_within_archive)
             ),
         )
 
@@ -269,7 +269,10 @@ class EiaRECSArchiver(AbstractDatasetArchiver):
         first_unselected_tabs = await get_unselected_tabs(url)
         another_tab_url = next(iter(first_unselected_tabs)).url
         next_unselected_tabs = await get_unselected_tabs(another_tab_url)
-        return first_unselected_tabs.union(next_unselected_tabs)
+        # Sort tabs by name to ensure stability between archive runs
+        return sorted(
+            first_unselected_tabs.union(next_unselected_tabs), key=lambda x: x.name
+        )
 
     async def __skip(self, **kwargs) -> list[str]:
         return []
--- a/src/pudl_archiver/package_data/zenodo_doi.yaml
+++ b/src/pudl_archiver/package_data/zenodo_doi.yaml
@@ -45,6 +45,7 @@ eiacbecs:
   sandbox_doi: 10.5072/zenodo.160999
 eiarecs:
   production_doi: 10.5281/zenodo.14783267
+  sandbox_doi: 10.5072/zenodo.269461
 eianems:
   production_doi: 10.5281/zenodo.14767206
   sandbox_doi: 10.5072/zenodo.180515


zaneselvans:
I don't think it would be terrible to remove the upper version bounds on all the dependencies -- this is mostly what we do in PUDL, unless we encounter a real reason why there needs to be a max version, which we then manually update when the problem is resolved.  But this is rare.

That being said, it seems like the "right" way to deal with this is to define [multiple pixi environments](https://pixi.sh/latest/workspace/multi_environment/) and I think that's a feature I'd like to use in the PUDL repo when we switch over, so maybe not a bad idea to figure it out here?

pyproject.toml
```diff
@@ -5,19 +5,7 @@ authors = [{ name = "PUDL", email = "pudl@catalyst.coop" }]
...
-    "dask>=2025,<2026",
-    "deltalake>=0.25,<1.1",
-    "feedparser>=6.0",
-    "frictionless>=5,<6",
-    "pydantic>=2.7,<3",
-    "python-dotenv~=1.1.0",
-    "pyyaml>=6.0,<7",
-    "semantic_version>=2.8,<3",
-    "tqdm>=4.64",
+    "catalystcoop.pudl @ git+https://github.com/catalyst-cooperative/pudl.git", "arelle-release>=2.37.23,<3", "beautifulsoup4>=4.13.4,<5", "coloredlogs>=14.0,<15", "dask>=2025.5.1,<2026", "deltalake>=1.0.2,<2", "feedparser>=6.0.11,<7", "frictionless>=5.18.1,<6", "pydantic>=2.11.7,<3", "python-dotenv>=1.1.0,<2", "pyyaml>=6.0.2,<7", "semantic-version>=2.10.0,<3", "tqdm>=4.67.1,<5",
```
zaneselvans, line 8: Is there some way that we can keep it from reformatting the dependencies like this? It's so much easier to see wtf is going on in diffs when there's just one per line.  Or use a TOML formatter pre-commit hook to enforce a particular format?

pyproject.toml
```diff
@@ -202,5 +190,5 @@ lint = [{task="ruff"}, {task="pre-commit"}]
 ci = [{task="lint"}, {task="unit"}, {task="integration"}, {task="coverage"}]
 
 [tool.pixi.dependencies]
-gdal = "==3.11.0"
+gdal = ">=3.11.0,<4"
```
zaneselvans, line 193: Is this going to cause incompatibilities with the GDAL in PUDL when there's a minor update?

zaneselvans:
It seems to be having a bunch of problems with the solve. I wonder if this might be due to our trying to migrate an existing `pyproject.toml` over all at once, rather than using `pixi add`  to build up the dependencies and environments -- like maybe our existing specification has something that's internally inconsistent from pixi's point of view and it didn't come up before?

zaneselvans:
Poked at this a bit but didn't come up with a solution. I feel like I don't totally understand how features / solve groups / environments work and would need to do some more reading and experimentation. It seems like there's a lot of top level things specified in the `pyproject.toml` that could (should?) be specified using pixi as well, which might make it all work together more smoothly? Not sure.

zaneselvans:
Well, it's not really the cleanest way to do it from `pixi`'s point of view, but putting all of the dependencies into the default environment seems to make the solve work, and I think it should work with the `pixi upgrade` action -- though it still munges all of the dependency specs onto a single line for me.

Issue #242, opened 2023-12-19
Automatic archival slack notification doesn't show failures


jdangerx:
If one of the archivers fails, the slack notification doesn't tell us. All we see is that the failed archiver doesn't report back at all.

We should pass in the matrix jobs to the slack notification maker so that it can report on failed jobs as well.

```[tasklist]
## success criteria
- [ ] failed archiver run shows up in Slack notification loudly
```

PR #729, opened 2025-07-08
When runs fail, save and report run failure to Slack

### LINKED ISSUE #723: Convert `size_diff` to MB in reporting to speed interpretation of results

e-belfer:
# Overview

Closes #723, #242 and #473.

- When a run fails, report its failure to Slackbot
- When a run succeeds, add its file size change type to the report to enable rapid-fire assessments of the scale of the changes.
- Convert this size change from bytes to MB. We don't make this actual conversion in the datapackage, just in the report-out.

- Save logs from GHA
- Parse non-validation errors from these logs and report out to Slack
- Report out file size changes

Note that nothing in the datapackage itself or the actual validation process is changed, this is all affecting the GHA and reporting layers.

# Testing

See the following runs and their corresponding messages:

**To test 3 datasets with the creation of a Github issue:**
https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/16482859753/job/46601559377
created this issue:
https://github.com/catalyst-cooperative/pudl-archiver/issues/758
and this slack message:
https://catalystcooperative.slack.com/archives/C03FHB9N0PQ/p1753308860215419

**To test 3 datasets with the creation of a Slack message and check that prod works ok:**
https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/16482863421

**To test trimming of the Slack message:**
https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/16482964692

# To-do list
- [x] Restructure to add pixi env into second half of steps
- [x] Handle "if no GH issue created" case
- [x] Fix validation errors showing up twice

# out of scope
- Handle "if re-run" case
- Actually make a nice tasklist render for each dataset 
- Fully drop `early_release_checker` and make matrix choices depend on cron

### DIFF
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -52,7 +52,7 @@ jobs:
           service_account: "pudl-sources@catalyst-cooperative-pudl.iam.gserviceaccount.com"
 
       - name: Set up pixi
-        uses: prefix-dev/setup-pixi@v0.8.12
+        uses: prefix-dev/setup-pixi@v0.8.14
         with:
           cache: true
 
--- a/.github/workflows/early_final_release_checker.yml
+++ b/.github/workflows/early_final_release_checker.yml
@@ -53,7 +53,7 @@ jobs:
           service_account: "pudl-sources@catalyst-cooperative-pudl.iam.gserviceaccount.com"
 
       - name: Set up pixi
-        uses: prefix-dev/setup-pixi@v0.8.12
+        uses: prefix-dev/setup-pixi@v0.8.14
         with:
           cache: true
 
--- a/.github/workflows/initialize-archiver-large.yml
+++ b/.github/workflows/initialize-archiver-large.yml
@@ -44,7 +44,7 @@ jobs:
           service_account: "pudl-sources@catalyst-cooperative-pudl.iam.gserviceaccount.com"
 
       - name: Set up pixi
-        uses: prefix-dev/setup-pixi@v0.8.12
+        uses: prefix-dev/setup-pixi@v0.8.14
 
       - name: Log the pixi environment
         run: |
--- a/.github/workflows/initialize-archiver.yml
+++ b/.github/workflows/initialize-archiver.yml
@@ -44,7 +44,7 @@ jobs:
           service_account: "pudl-sources@catalyst-cooperative-pudl.iam.gserviceaccount.com"
 
       - name: Set up pixi
-        uses: prefix-dev/setup-pixi@v0.8.12
+        uses: prefix-dev/setup-pixi@v0.8.14
 
       - name: Log the pixi environment
         run: |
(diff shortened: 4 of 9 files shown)


scripts/make_slack_notification_message.py
```diff
@@ -71,20 +85,70 @@ def _format_summary(summary: dict) -> list[dict]:
     if file_changes := summary["file_changes"]:
         abridged_changes = defaultdict(list)
         for change in file_changes:
-            abridged_changes[change["diff_type"]].append(change["name"])
+            abridged_changes[change["diff_type"]].append(
+                # Convert the size diff to MB for speed of assessment, and round
+                {change["name"]: round(change["size_diff"] * (2**-20), 4)}
+            )
```
aesharpe, line 91: If you're converting from bytes to MB, wouldn't it be 1**-6?

aesharpe (commented):
I might pass this on to @cmgosnell or @krivard for another review, but wondering why the FERC 6 archiver change outputs look like this: 

```
 "UPDATE": [
    {
      "ferc6-xbrl-2014.zip": -0.0
    },
    {
      "ferc6-xbrl-2015.zip": -0.0
```

Is this because of the rounding? I'm assuming it's just a verrryyyy small number.

.github/workflows/run-archiver.yml
```diff
@@ -83,7 +83,15 @@ jobs:
           ZENODO_TOKEN_PUBLISH: ${{ secrets.ZENODO_TOKEN_PUBLISH }}
         if: ${{ inputs.server == 'sandbox' }}
         run: |
-          pixi run pudl_archiver --datasets ${{ matrix.dataset }} --sandbox --summary-file ${{ matrix.dataset }}_run_summary.json --clobber-unchanged
+          set -o pipefail
```
krivard, line 86: Maybe add a comment here saying why we need pipefail?

scripts/make_slack_notification_message.py
```diff
@@ -71,20 +85,70 @@ def _format_summary(summary: dict) -> list[dict]:
...
+                summaries.extend(json.loads(f.read()))
+    return summaries
+
+
+def _load_errors(error_files: list[Path]) -> list[str]:
+    errors = []
+    for error_file in error_files:
+        if (
+            error_file.exists() and error_file.stat().st_size > 0
+        ):  # Handle case where no files are found or file is empty
```
krivard, line 136: what causes the script to sometimes receive a filename that doesn't exist?

scripts/make_slack_notification_message.py
```diff
@@ -71,20 +85,70 @@ def _format_summary(summary: dict) -> list[dict]:
...
+
+
+def main(summary_files: list[Path], error_files: list[Path]) -> None:
+    """Format summary files for Slack perusal."""
+    summaries = _load_summaries(summary_files)
+    errors = _load_errors(error_files)
+
+    error_blocks = list(
+        itertools.chain.from_iterable(
+            _format_errors(e) for e in errors if _format_errors(e) is not None
```
krivard, line 149: Rather than calling `_format_errors` twice, consider

```suggestion
            filter(None, (_format_errors(e) for e in errors))
```

krivard (approved):
Fix the units thing for sure; other than that I'm just being nosy and this is fine

PR #756, opened 2025-07-22
Add playwright to PHMSA archiver

### LINKED ISSUE #725: phmsagas - Archiver repeatedly times out on `self.get_hyperlinks()`
Archiver times out repeatedly when running links = await self.get_hyperlinks(BASE_URL, link_pattern). Quick implementation of a user agent did not seem to resolve this immediately. The [URL in question](https://www.phmsa.dot.gov/data-and-statistics/pipeline/gas-distribution-gas-gathering-gas-transmission-hazardous-liquids) is up and working fine in the browser. See runs in https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/15994132343/job/45222775329

e-belfer:
# Overview

Closes #725 

- Archiver times out using both the existing get methods and those same methods with a randomized header.
- Archiver times out on both the main getting of links and on downloading each individual file.

- Switched to using playwright to download methods, copying the format from the FERC2 archiver.
- Added a new method `get_hyperlinks_via_playwright()`

# Testing

Run `pudl_archiver --dataset phmsagas --summary-json phmsa.json`

# To-do list

- [ ] Once approved, update the main archive

### DIFF
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -401,9 +401,9 @@ class AbstractDatasetArchiver(ABC):
 
         Args:
             url: URL of web page.
+            playwright_browser: async browser instance to use for fetching the URL.
             filter_pattern: If present, only return links that contain pattern.
             verify: Verify ssl certificate (EPACEMS https source has bad certificate).
-            headers: Additional headers to send in the GET request.
         """
         # Parse web page to get all hyperlinks
         page = await playwright_browser.new_page()


src/pudl_archiver/archivers/phmsagas.py
```diff
@@ -28,12 +30,25 @@ class PhmsaGasArchiver(AbstractDatasetArchiver):
 
     name = "phmsagas"
 
+    async def after_download(self) -> None:
+        """Clean up playwright once downloads are complete."""
+        await self.browser.close()
+        await self.playwright.stop()
```
krivard, line 36: Noting for posterity this is the second time we're using this pattern; once we hit three we should refactor to pull out a central implementation

krivard (approved):
Looks solid; needs a docstring fix but after that should be good to go

PR #891, opened 2025-10-20
Fix `fsspec` workflow default path on scheduled runs

zschira:
# Overview

The new workflow to run fsspec backed archivers (currently used to archive FERC EQR to GCS) has an issue where the default fsspec path (points to gs://archives.catalyst.coop) doesn't get used on scheduled runs. See [this run](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/18646689392). This pr updates the workflow to set the path when it is empty.


e-belfer (approved):
Ah yes, the classic GHA annoyance with the auto-runs not catching defaults set in the workflow_dispatch section. Good catch on this one!

PR #899, opened 2025-10-23
Don't run early final release checker on sundays

### LINKED ISSUE #897: Update schedule for `early-final-release-checker.yml`
We updated the schedule for the main `run-archiver.yml` to run on Sundays, which means we should update the release checker's schedule to not run on Sundays. This is as simple as updating the cron schedule.

zaneselvans:
# Overview

* Update final release checker to avoid Sundays so it doesn't collide with the primary archiver run.
* Also addressed some suggestions from `shellcheck` which came up from the github action workflow linter.

Closes #897


jdangerx (approved):
Just the comment fix! Otherwise good to go.

PR #930, opened 2025-11-10
Eqr archiver improvements

zschira:
# Overview

The automated EQR archiver runs have had some failures due to timeouts during downloads and file size diff validation test failures. I'm not sure the exact cause or solution to the timeout issue. It seems like we may be getting rate limited by FERC, in which case adding breaks between downloads might help, but that's complete speculation at this point. As for the file size diff, I've set `fail_on_file_size_change = False` in the EQR archiver, which should resolve this issue. It would probably be slightly better if we had a way to make this test still fail if any quarters other than the most recent has significant size changes, but that would be a much larger change.

I also modified the `run-fsspec-archiver` workflow so it only runs once a month since I think anything more frequent is unnecessary for EQR at this point.

### DIFF
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -550,6 +550,12 @@ class AbstractDatasetArchiver(ABC):
                     f"The following files have absolute changes in file size >|{self.allowed_file_rel_diff:.0%}|: {too_changed_files}"
                 ]
 
+            if len(self.ignore_file_size_increase_partitions) > 0:
+                notes += [
+                    "Size increases ignored for resource with the following partitions: {partitions}"
+                    for partitions in self.ignore_file_size_increase_partitions
+                ]
+
         return validate.DatasetUniversalValidation(
             name="Individual file size test",
             description=f"Check for files from previous version of archive that have changed in size by more than {self.allowed_file_rel_diff:.0%}.",
--- a/src/pudl_archiver/archivers/ferc/ferceqr.py
+++ b/src/pudl_archiver/archivers/ferc/ferceqr.py
@@ -3,9 +3,9 @@
 import asyncio
 import logging
 import re
-from datetime import date
 from pathlib import Path
 
+import pandas as pd
 from playwright.async_api import async_playwright
 
 from pudl_archiver.archivers.classes import (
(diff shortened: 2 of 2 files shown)


src/pudl_archiver/archivers/classes.py
```diff
@@ -506,28 +512,33 @@ def _check_file_size(
...
+                            - baseline_resources[resource_name].bytes_
+                        ) / baseline_resources[resource_name].bytes_
+
+                        # Check if resource is included in set that should be ignored
+                        # on size increase
+                        if any(
+                            baseline_resources[resource_name].parts == parts
+                            for parts in self.ignore_file_size_increase_partitions
+                        ) and (file_size_change > 0):
+                            continue
```
e-belfer: Would `baseline_resources[resource_name].parts in self.ignore_file_size_increase_partitions` work equivalently as a condition here? I'm finding this a bit hard to read.

src/pudl_archiver/archivers/classes.py
```diff
@@ -126,6 +127,11 @@ class AbstractDatasetArchiver(ABC):
     allowed_dataset_rel_diff: float = 0.15
     fail_on_data_continuity: bool = True
 
+    # Allow specific partitions to be ignored by the relative file size diff test
+    # if the size has increased. This is useful for new data that will grow as more
+    # becomes available
+    ignore_file_size_increase_partitions: [dict[str, Any]] = []
```
e-belfer: It would be helpful to add this to the notes field for this test so we have a record of it somewhere other than the ephemeral logs.

src/pudl_archiver/archivers/classes.py
zschira: Unfortunately, no. `dict`s aren't hashable, which is required to be able to do comparisons like that.

src/pudl_archiver/archivers/classes.py
```diff
@@ -506,28 +512,33 @@ def _check_file_size(
...
+                        # Check if resource is included in set that should be ignored
+                        # on size increase
+                        if any(
+                            baseline_resources[resource_name].parts == parts
+                            for parts in self.ignore_file_size_increase_partitions
+                        ) and (file_size_change > 0):
+                            continue
+                        # Note failure
+                        if abs(file_size_change) > self.allowed_file_rel_diff:
                             too_changed_files.update({resource_name: file_size_change})
```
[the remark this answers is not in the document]
zschira: Yeah, they should!

e-belfer (approved):
With those changes made this looks good to me. Agreed, the logs indicate things are working as expected even though the test wasn't disabled. I think it's ok to merge without re-testing, and we'll know very shortly whether it's not working as expected.

PR #955, opened 2025-12-12
Resolve multiple filename options in nrelefs

### LINKED ISSUE #652: nrelefs - inconsistent file names
A duplicate link to https://www.nrel.gov/docs/fy21osti/78783.pdf results in one of two file names getting applied to the same file. changing the size of the dataset ever so slightly. Need to use `sort` or another method to make this consistent over time.

krivard:
# Overview

Closes #652

* Make _HyperlinkExtractor record all anchor text it encounters for each link
* Handle what to do with multiple anchor text options in AbstractDatasetArchiver.get_hyperlinks_from_text(): permit returning all available anchor text options, but default to returning the anchor text that appears closest to the end of the file
* Thread that through AbstractDatasetArchiver.get_hyperlinks() and into NrelEFSArchiver
* Add a spotfix for [78783.pdf](https://www.nrel.gov/docs/fy21osti/78783.pdf) to pick the "presentation slides" option
* Scream if anyone else has more than one anchor text option, or if the "presentation slides" option for [78783.pdf](https://www.nrel.gov/docs/fy21osti/78783.pdf) is missing

While I was here, I also fixed:

* Check against list of URLs for `links_not_downloaded`, not list of versions
* Whitespace in warning for get_hyperlinks()

# Testing

* Ran locally (at least up until it had things to download; this archive is on the larger side so it often times out on my home internet) and witnessed no screaming

# To-do list

### DIFF
--- a/src/pudl_archiver/archivers/classes.py
+++ b/src/pudl_archiver/archivers/classes.py
@@ -82,14 +82,14 @@ class _HyperlinkExtractor(HTMLParser):
                 self.current_hyperlink = val
                 if self.current_hyperlink not in self.hyperlinks:
                     # By default, set the name identical to the hyperlink
-                    self.hyperlinks[self.current_hyperlink] = self.current_hyperlink
+                    self.hyperlinks[self.current_hyperlink] = [self.current_hyperlink]
                 break
 
     def handle_data(self, data):
         """Capture text content and associate it with the current URL."""
         if self.current_hyperlink and data.strip():
             # If data is present, replace the hyperlink name
-            self.hyperlinks[self.current_hyperlink] = data.strip()
+            self.hyperlinks[self.current_hyperlink].append(data.strip())
 
     def handle_endtag(self, tag):
         """Reset the current URL after processing the content of the tag."""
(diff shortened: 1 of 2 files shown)


src/pudl_archiver/archivers/classes.py
```diff
@@ -392,72 +399,98 @@ async def get_hyperlinks(
...
-                f"In {context}: the archiver couldn't find any hyperlinks {('that match: ' + filter_pattern.pattern) if filter_pattern else ''}."
+                f"In {context}: the archiver couldn't find any hyperlinks{(' that match: ' + filter_pattern.pattern) if filter_pattern else ''}. "
                 f"Make sure your filter_pattern is correct, and check if the structure of the page is not what you expect it to be."
             )
 
-        return hyperlinks
+        return (
+            hyperlinks
+            if all_names
+            else {link: names.pop() for link, names in hyperlinks.items()}
```
krivard: (`.pop()` returns the last item in the list, which is the one that appeared closest to the end of the file)

e-belfer:
Closing in favor of #1021, as we now consider this data archiver retired.

e-belfer:
@krivard I closed this because the archiver that was causing the issue is no longer running. If there are changes in here that would be generally beneficial, feel free to re-open and retag me for a more prompt review this time.

PR #963, opened 2025-12-22
Add FERC CID archiver

katie-lamb:
# Overview

Closes [this issue](https://github.com/catalyst-cooperative/pudl/issues/4832).

Archives the FERC Company Identifier table from [FERC's interactive data viewer page](https://data.ferc.gov/company-registration/ferc-company-identifier-listing/).

Added an archiver to page for the CID table data from FERC's interactive data viewer page and save the table as a CSV.
Accompanying metadata PR is [here](https://github.com/catalyst-cooperative/pudl/pull/4861)

Zenodo sandbox archive [here](https://sandbox.zenodo.org/records/420250)

# Testing

- [x] Ran locally
- [x] Merge in the PUDL data source metadata PR
- [x] Uploaded to Zenodo sandbox.
- [x] Preview Zenodo sandbox archive and hit publish

# To-do list

### DIFF
--- a/.github/workflows/run-archiver.yml
+++ b/.github/workflows/run-archiver.yml
@@ -6,7 +6,7 @@ on:
     inputs:
       datasets:
         description: 'Comma-separated list of datasets to archive (e.g., "ferc2","ferc6").'
-        default: '"censuspep","doeiraec","doelead","eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiaapi","eiacbecs","eiamecs","eianems","eiarecs","eiasteo","eiawater","epacamd_eia","epacems","epaegrid","epamats","epapcap","ferc1","ferc2","ferc6","ferc60","ferc714","gridpathratoolkit","mshamines","nrelatb","nrelefs","nrelsiting","nrelss","nrelsts","phmsagas","sec10k","usgsuspvdb","usgsuswtdb","vcerare"'
+        default: '"censuspep","doeiraec","doelead","eia176","eia191","eia757a","eia860","eia860m","eia861","eia923","eia930","eiaaeo","eiaapi","eiacbecs","eiamecs","eianems","eiarecs","eiasteo","eiawater","epacamd_eia","epacems","epaegrid","epamats","epapcap","ferc1","ferc2","ferc6","ferc60","ferc714","ferccid","gridpathratoolkit","mshamines","nrelatb","nrelefs","nrelsiting","nrelss","nrelsts","phmsagas","sec10k","usgsuspvdb","usgsuswtdb","vcerare"'
         required: true
         type: string
       server:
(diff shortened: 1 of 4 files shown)


src/pudl_archiver/archivers/ferc/ferccid.py
```diff
@@ -0,0 +1,88 @@
...
+
+class FercCIDArchiver(AbstractDatasetArchiver):
+    """FERC CID archiver."""
+
+    name = "ferccid"
+
+    async def get_resources(self) -> ArchiveAwaitable:
+        """Download FERC CID resources."""
+        # get CSV
+        # https://ferc.gov/sites/default/files/2025-06/FERC CID Listing 6-10-2025.csv
```
zaneselvans: I'm suspicious that this filename (and the XLSX one below) may be manually assigned, and not follow a consistent naming convention between releases, in which case we may want to use a more flexible regular expression.

This would also make the extraction of the date elements more complex / unreliable. It might be possible to use `pd.to_datetime()` on a subset of the string and have it to more generic parsing.

zaneselvans: I think their interactive table is a better primary source for this data because:

- It's been updated much more recently, and I suspect more consistently.
- It seems like the CSV/XLSX outputs are the same (while the lonely individual files are not)
- The lonely CSV/XLSX filenames look like they are probably not stable.
- The interactive interface also provides a data dictionary, which we should archive.
- The interactive interface is uniform across many interesting tables that it would be great to grab.

I haven't really worked with Playwright, but there are other archivers in here that use it to navigate javascripty interfaces.

src/pudl_archiver/archivers/ferc/ferccid.py
```diff
@@ -0,0 +1,169 @@
...
+                    payload = {
+                        "startRow": start,
+                        "endRow": start + page_size,
+                        "sortModel": [],
+                        "filterModel": {},
+                        "columns": columns,
+                        "castData": [],
+                    }
+                    # make the request from inside the page
+                    result = await page.evaluate(
```
zschira: Could you add some clarifying comments about what's going on here (in the docstring and/or inline)? From what I can tell we're continuously querying the actual api that supplies the data for the interactive table on the CID page until we've fetched all rows, but it would be nice to have a bit more clarification on exactly what we're doing/why.

Also, is there a reason we have to use `playwright` to query the API from within the page using Javascript? could we just query the API directly with python?

zschira (dismissed):
I ran this locally, and the outputs look good to me, so I think we can get this merged soon, but just wanted a bit of clarification on how we're querying the CID page.

src/pudl_archiver/archivers/ferc/ferccid.py
katie-lamb: It didn't work for me to query the API directly with python. I don't fully understand why, but it seems like you need to simulate being in a browser before making the request. I'll leave some comments.

src/pudl_archiver/archivers/ferc/ferccid.py
```diff
@@ -0,0 +1,169 @@
...
+                await page.wait_for_load_state("networkidle", timeout=timeout_ms)
+
+                all_rows: list[dict[str, Any]] = []
+                start = 0
+                total_count: int | None = None
+
+                while True:
+                    payload = {
+                        "startRow": start,
+                        "endRow": start + page_size,
```
jdangerx: Did you try not providing a `startRow`, `endRow`, and `castData`? That seems like what my browser is doing when I do the bulk API request... if that works we can avoid this whole pagination situation completely which would be nice.

jdangerx (commented):
I do think that hitting the API once for the bulk download is worth a shot - tried to get archiver stuff set up on my computer and ran into some issues with my specific setup, but left a suggestion... if that doesn't work I think the page-by-page approach with some extra documentation would be *fine*.

src/pudl_archiver/archivers/ferc/ferccid.py
```diff
@@ -0,0 +1,169 @@
...
+
+        match = DATE_LAST_UPDATED_PATTERN.search(row_text)
+        if not match:
+            raise RuntimeError(
+                f"Couldn't find date-like value near 'Data Last Updated'. Row text was: {row_text!r}"
+            )
+
+        return match
+
+    async def download_cid_csv_by_paging_api(
```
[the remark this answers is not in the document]
katie-lamb: I was able to do button clicks to get to this popup page, but then actually grabbing the data dictionary data after doing the button clicks was proving difficult. I can try entering JS world though if you think that's what the key was.

src/pudl_archiver/archivers/ferc/ferccid.py
jdangerx: This works on my machine:


```python
from playwright.sync_api import sync_playwright

SOURCE_URL = (
    "https://data.ferc.gov/company-registration/ferc-company-identifier-listing/"
)
DATASET_API_URL = "https://data.ferc.gov/api/v1/dataset/26/export/?format=csv"


def main():
    columns = [
        "Organization_Name",
        "CID",
        "Program",
        "Company_Website",
        "Address",
        "Address2",
        "City",
        "State",
        "Zip",
    ]
    payload = {"sortModel": [], "filterModel": {}, "columns": columns}
    with sync_playwright() as pw:
        browser = pw.chromium.launch(headless=False)
        page = browser.new_page()
        page.goto(SOURCE_URL, wait_until="domcontentloaded")
        page.wait_for_load_state("networkidle")
        breakpoint()
        result = page.evaluate(
            """
    async ({url, payload}) => {
    const resp = await fetch(url, {
        method: "POST",
        headers: { "Content-Type": "application/json" },
        body: JSON.stringify(payload),
        credentials: "include",
    });
    const text = await resp.text();
    return { status: resp.status, ok: resp.ok, text };
    }
    """,
            {"url": DATASET_API_URL, "payload": payload},
        )
        print(result)


if __name__ == "__main__":
    main()

```

jdangerx (approved):
Paired on this - looks good and the sandbox action worked. Let's make the prod DOI, update this, and merge!

PR #987, opened 2026-01-06
Refactor CLI to make commands more logical

### LINKED ISSUE #828: Refactor the top level interface
## Background
After years of use and new functionality being strapped on to the archiver, the top level interface has become cumbersome and confusing. To make the archiver easier to use and develop, we need to refactor this interface to make it more streamlined and less complex.

####  The CLI is trying to do too much with one command
- The `zenodo`/`fsspec` commands have different arguments and its confusing which to use. These should be split into two commands. We should also have separate settings objects for storing only the relevant settings for each backend.
- There should be a separate command for retrying a run. #782 adds retry functionality, but retries should always inherit the settings from the run it's retrying, so this could be a simple command that just points to a run summary file and takes no other arguments
- There should be another standalone command for publishing an open "draft deposition". With the zenodo backend we have typically relied on their web interface, but we don't have that option with the fsspec backend. This can be worked around by using the `--auto-publish` argument, but this does not leave us with an opportunity to manually review the status before publishing.

### Unused functionality causing unnecessary complexity
There's certain functionality that we've added over the years that we no longer use, but still exists and makes things unnecessarily complex. This includes:

- The ability to archive more than one dataset at a time. In practice we generally run archivers one at a time during local development, or use a separate github runner for each dataset in it's own run, so this functionality is never used.
- Unused CLI args. I'm not sure if we still use the `--only-years` arg, but @e-belfer might know. In general, we should review all the args and decide which are still necessary.

## Tasks
- [x] Create separate CLI commands for `zenodo`/`fsspec` backends with only necessary args and store these in separate settings objects
- [x] Create standalone command to retry a run
- [x] Create standalone command to publish a draft deposition
- [x] Remove ability to run multiple archivers at once
- [x] Review all existing CLI args and remove any that are no longer needed + consolidate args if possible

zschira:
# Overview

Closes #828 .

## What problem does this address?
This PR refactors the CLI to make the top level commands more logical and easier to maintain. Specifically, it splits the `zenodo` / `fsspec` backends to use different commands so it's clear which arguments apply to each backend. It also creates a new standalone `retry-run` command that can be used to retry a failed run, and to publish the outputs from a successful run that didn't set `auto-publish`.  The new commands also only support running one archiver at a time since that's all we do in practice, which also simplifies much of the high-level orchestration logic.

# Testing

I've used the new CLI with both backends, to do test runs, and I've tried the `retry-run` command locally with the `fsspec` depositor. I've also updated all of the tests to work with this new backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,16 +356,14 @@ Once you've written your archiver, it's time to test that it works as expected! 
 the archiver locally, run the following commands in your terminal:
 
 ```bash
-pudl_archiver --datasets {new_dataset_name} --initialize --summary-file {new_dataset_name}-summary.json --depositor fsspec --deposition-path {file://local/path/to/folder}
+pudl_archiver archive fsspec {new_dataset_name} {file://local/path/to/folder} --initialize
 ```
 
 - `--initialize` creates a new deposition, and is used when creating a brand new archive
-- `--summary-file` will save a .json file summarizing the results of all
-validation tests, which is useful for reviewing your dataset.
-- `--depositor` selects the backend engine used for archive storage - in this case,
-we save files locally, but by default this uploads files to Zenodo.
-- `--depositor-path`: the path to the folder where you want to download local files for
-inspection.
+- A JSON file file summarizing the results of all validation tests, will be saved at
+`{new_dataset_name}_run_summary.json`. This file can be useful for reviewing your dataset.
+- Using the `fsspec` backend with the argument `{file://local/path/to/folder}` will save
+the results of the run to this local path.
 
 Run the archiver and review the output in the specified folder, iterating as needed to
 ensure that all files download as expected.


src/pudl_archiver/orchestrator.py
```diff
@@ -15,83 +13,27 @@
 logger = logging.getLogger(f"catalystcoop.{__name__}")
 
 
-def _load_failed_run_summary(
```
zschira: This is the logic that now exists in the top-level `retry-run` command, and is much simpler to do there.

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -227,27 +221,12 @@ async def get_file(self, filename: str) -> bytes:
...
-            fname: self.deposition.get_deposition_path(DepositionDirectory.WORKSPACE)
-            / fname
-            for fname in self.deposition.deposition_files[DepositionDirectory.WORKSPACE]
-        }
         return FsspecDraftDeposition(
             deposition=self.deposition,
             settings=self.settings,
             api_client=self.api_client,
             dataset_id=self.dataset_id,
             # When we open a new draft we assume it starts with all files from previous version
```
e-belfer: Is this comment still relevant now that the code has been removed?

src/pudl_archiver/depositors/fsspec.py
```diff
@@ -277,6 +256,25 @@ class FsspecDraftDeposition(DraftDeposition):
...
+        # Draft starts with all files in previous published deposition
+        draft_files = {
+            fname: self.deposition.get_deposition_path(DepositionDirectory.PUBLISHED)
+            / fname
+            for fname in self.deposition.deposition_files[DepositionDirectory.PUBLISHED]
+        }
+
+        # If we are retrying a previous run there may already be files in the workspace directory
+        # Add these to the draft as well as files in the current published archive
+        # TODO: Check if we're doing a retry as we don't expect files here unless we're doing a retry
```
e-belfer: Looks like this is still an open todo?

src/pudl_archiver/cli.py
```diff
@@ -1,128 +1,199 @@
...
+    help="Years to download data for. Supported datasets: censusdp1tract, censuspep, "
+    "eia176, eia191, eia757a, eia860, eia860m, eia861, eia923, eia930, "
+    "eiaaeo, eiamecs, eiawater, eiasteo, epacamd_eia, epacems, epaegrid, ferc1, ferc2, "
+    "ferc6, ferc60, ferc714, mshamines, nrelatb, phmsagas, usgsuswtdb",
+)
+dataset_argument = click.argument("dataset", type=str)
+
+
+@archive.command
+@click.option("--sandbox", is_flag=True, help="Use Zenodo sandbox server")
```
e-belfer: Would structurally be helpful to keep the order between method only and shared options the same for Zenodo and fsspec. Here it's the first one in the list, below it's the last - visually it'd be helpful to keep them both either at the top or bottom so it's quick to note that these are the method-specific options.

PR #1077, opened 2026-04-13
Add `fsspec` publication workflow

zschira:
# Overview

This PR adds a workflow for publishing `fsspec` archives. It also modifies the `run-fsspec-archiver` to turn off auto-publish, so we can enable a review process like we have for zenodo backed archives. Now, after `run-fsspec-archiver` is complete, and new or updated data will be left in the `{ARCHIVE_BASE_DIRECTORY}/workspace` where the changes can be inspected. Once the inspection is complete, we can run `publish-fsspec-archive` with the run ID from the first workflow to publish the results.

### DIFF
--- a/.github/workflows/publish-fsspec-archive.yml
+++ b/.github/workflows/publish-fsspec-archive.yml
@@ -40,11 +40,6 @@ jobs:
           pixi info
           pixi list
 
-      - name: Additional Playwright setup
-        run: |
-          pixi run playwright install --with-deps webkit
-          pixi run playwright install --with-deps chromium
-
       - name: Download run summary
         uses: actions/download-artifact@v8
         with:
@@ -59,6 +54,5 @@ jobs:
           set -o pipefail
           pixi run pudl_archiver \
             archive \
-            retry-run \
-            --auto-publish \
+            publish-run \
             ${{ inputs.dataset }}_run_summary.json 2>&1 | tee ${{ inputs.dataset }}_log.txt
--- a/.github/workflows/publish-fsspec-archive.yml
+++ b/.github/workflows/publish-fsspec-archive.yml
@@ -42,6 +42,13 @@ jobs:
           pixi run playwright install --with-deps webkit
           pixi run playwright install --with-deps chromium
 
+      - name: Download run summary
+        uses: actions/download-artifact@v8
+        with:
+          run-id: ${{ inputs.run_id }}
+          pattern: run-summaries-${{ inputs.dataset }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Attempt to publish deposition for ${{ inputs.dataset }}
         # set -o pipefail passes on any failure codes to the next step,
         # enabling us to see whether or not the pudl_archiver run failed.
@@ -51,11 +58,4 @@ jobs:
             archive \
             retry-run \
             --auto-publish \
-            2>&1 | tee ${{ inputs.dataset }}_log.txt
-
-      - name: Download run summary
-        uses: actions/download-artifact@v8
-        with:
-          run-id: ${{ inputs.run_id }}
-          pattern: run-summaries-${{ inputs.dataset }}
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+            ${{ inputs.dataset }}_run_summary.json 2>&1 | tee ${{ inputs.dataset }}_log.txt
--- a/.github/workflows/publish-fsspec-archive.yml
+++ b/.github/workflows/publish-fsspec-archive.yml
@@ -1,0 +1,61 @@
+---
+name: publish-fsspec-archive
+
+on:
+  workflow_dispatch:
+    inputs:
+      run_id:
+        description: Run ID pointing to run of run-fsspec-archiver action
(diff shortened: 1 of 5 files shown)


src/pudl_archiver/cli.py
```diff
@@ -153,6 +153,53 @@ def fsspec(
     )
 
 
+@archive.command
+@click.argument(
+    "summary-file",
+    type=str,
+)
+def publish_run(summary_file: str):
```
zschira, line 159: I added a new `publish-run` command that does basically the exact same thing as the `retry-run` command, except it first checks to make sure there were no failed partitions before continuing a run. I did this to make it clear that the `publish-fsspec-archive` workflow will only publish results from a successful run, and will not retry any failed partitions, which could be misleading.

src/pudl_archiver/cli.py
```diff
@@ -184,6 +230,12 @@ def retry_run(summary_file: str, auto_publish: bool):
     failed_partitions = failed_run_summary.failed_partitions
     successful_partitions = failed_run_summary.successful_partitions
 
+    if failed_partitions == {}:
```
zschira, line 220: Also added a check to `retry-run` so it will only work if there are actually failed partitions. This is to make the functionality of `retry-run` vs `publish-run` clearly distinct.

src/pudl_archiver/cli.py
```diff
@@ -153,6 +153,53 @@ def fsspec(
...
 
+@archive.command
+@click.argument(
+    "summary-file",
+    type=str,
+)
+def publish_run(summary_file: str):
+    """Publish the results from a successful run that did not have ``auto-publish`` set.
+
+    SUMMARY_FILE Points to a Run Summary JSON file output by a previous run which
```
e-belfer:
```suggestion
   Args:
         summary_file: A string pointing to a Run Summary JSON file output by a previous run which
```
String and not a path?

src/pudl_archiver/cli.py
```diff
@@ -153,6 +153,53 @@ def fsspec(
...
+    set to ``True``.
+    """
+    # Load run summary file and parse
+    with Path(summary_file).open() as f:
+        previous_run_summary = RunSummary.model_validate(json.load(f))
+
+    # Extract settings from failed run
+    run_settings = previous_run_summary.run_settings.model_copy(
+        update={
+            "retry_run": summary_file,
```
e-belfer: Mind adding a note in the docstring about why you update this one?

src/pudl_archiver/cli.py
```diff
@@ -153,6 +153,53 @@ def fsspec(
...
+
+    if failed_partitions != {}:
+        raise RuntimeError(
+            "publish-run should not be called on a run with failed partitions."
+            "Instead, use the retry-run command."
+            f"Found the following failed partitions: {failed_partitions.keys()}"
+        )
+
+    # Retry archiver run
+    asyncio.run(
```
e-belfer, line 187: Will this not re-run the archive? Maybe I'm getting a bit tripped up by the `archive_dataset()` function. Would be helpful to clarify in the docstring here.

README.md
```diff
@@ -143,12 +143,11 @@ successfully added to the archive in the previous run. Basic usage looks like th
 following:
 
 ```bash
-pudl_archiver retry-run {run_summary_json_file}
+pudl_archiver archive retry-run {run_summary_json_file}
```
e-belfer: So conceptually I think having `retry-run` and `publish-run` separately from the `archive` command helps clarify that the archiver is not being run again as part of the publication process (if that's indeed true?). Is there a practical reason for reorganizing this?

.github/workflows/publish-fsspec-archive.yml
e-belfer: Github run ID? Might be helpful to clarify a bit more in case anyone other than you wants to run this some day :)

e-belfer (approved):
I see from your test runs that https://github.com/catalyst-cooperative/pudl-archiver/actions/workflows/publish-fsspec-archive.yml and the corresponding `run-fsspec-archiver.yml` work as expected. I have a few requests for more docs and one non-blocking structural question, but in general this looks good to me and I don't think anything is blocking.

src/pudl_archiver/cli.py
zschira: I added a dedicated method for loading `RunSummary` objects with some detail in the docstring.

src/pudl_archiver/cli.py
zschira: Added some expanded comments. In this case since we're passing `failed_partitions` and `successful_partitions` from the previous run, but all of the partitions were successful, it won't try to download anything and will just try to publish the results of the previous run.

README.md
zschira: `retry-run` was actually always within the `archive` command, and this example was incorrect, but I agree that it makes more sense for these to be standalone commands, so I've changed that.

.github/workflows/publish-fsspec-archive.yml
zschira: Added an expanded description. It might be good for us to find a place in the docs to include instructions for the monthly review process.

PR #1105, opened 2026-05-24
Goodbye PUDL

### LINKED ISSUE #499: Re-design `pudl-archiver` to better separate dependencies from main `pudl` repository

zaneselvans:
# Overview

- Obtain PUDL data source and contributor metadata from the nightly PUDL `datapackage.json` descriptor instead of depending on the the PUDL repository directly.
- This totally seems to work! I had to futz with the datapackage.json outputs from the `better-datapackage` pudl repo branch a little bit to include more of our custom metadata, which we were relying on in here.
- Remove the `catalystcoop.pudl` dependency.
- Add in some missing direct dependencies we were previously getting through the PUDL install.
- Remove some derelict dependencies that weren't actually being used any more.
- Bump python version to 3.14+
- Switch from using `pre-commit` to `prek` as we have in our other repos.

## Review Questions

- One thing that we lose with this transition is access to the PUDL metadata classes like `DataSource` and `Contributor` which we were using in this repo to do data validation. Now it's all just plain Python dictionaries being passed around without any declared structure.
- Is this okay? On the PUDL side the outputs that end up in the `datapackage.json` are still being validated. But we also define a bunch of non-PUDL sources in the `pudl-archive` repo, and now they aren't being validated. Should we define some Pydantic models or typed dictionaries or data classes in here to make sure that we're using a uniform data model?
- For the moment I've added a static `datapackage.json` file in a testing bucket (since we don't have a nightly build with the real thing yet) -- 

Closes #499 

# Testing

- Ran the unit and integration tests locall and in GitHub CI.

# Followup Issues

- Add lightweight data validation to metadata structures in `pudl-archiver`

# To-do list

- [x] Do a PR self-review
- [x] Try a [normal archiver run](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/26354310372) on the `goodbye-pudl` branch
- [x] Try a [FERC EQR archiver run](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/26357571778/job/77587015152) on the `goodbye-pudl` branch
- [x] Rename `get_sources()` --> `get_pudl_sources()`
- [x] Rename `get_contributors()` --> `get_pudl_contributors()`
- [x] Switch to the real PUDL datapackage.json URL once we have a nightly build output.
- [x] Update the documentation / README


.pre-commit-config.yaml
```diff
@@ -29,7 +29,7 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.15.12
     hooks:
-      - id: ruff
```
zaneselvans, line 32: Update legacy name

pyproject.toml
```diff
@@ -177,44 +174,40 @@ ci = { depends-on = [
 ], description = "Run full CI suite" }
 
 [tool.pixi.dependencies]
+aiobotocore = ">=3.7.0,<4"
```
zaneselvans, line 173: There were a number of dependencies which we use directly in this repo that weren't listed here explicitly, and were getting included as transitive dependencies from `catalystcoop.pudl` They've been added directly now.  These included:

- `aiobotocore`
- `botocore`
- `fsspec`
- `google-cloud-storage`
- `python-dateutil`
- `requests`

There were also several dependencies listed that weren't actually being used, and they've been removed, including:

- `dask`
- `numba`
- `numpy`
- `psycopg2-binary`
- `pyproj`

But if there was something subtle going on that makes any of these guys necessary let me know!

tests/unit/conftest.py
```diff
@@ -0,0 +1,10 @@
+"""Unit test configuration."""
+
+import os
+from pathlib import Path
+
+# Use the checked-in fixture descriptor so unit tests never fetch from S3.
+os.environ.setdefault(
+    "PUDL_DATAPACKAGE_PATH",
+    str(Path(__file__).parent.parent / "fixtures" / "datapackage.json"),
```
zaneselvans, line 9: Use a minimal datapackage stand-in to avoid the unit tests trying to pull anything from S3.

tests/fixtures/datapackage.json
zaneselvans, line 1: A minimal datapackage.json that the unit tests can rely on so they don't need to touch the network.

src/pudl_archiver/__init__.py
zaneselvans, line 1: Removed some unused imports in here.  Was there a reason they were actually needed?

src/pudl_archiver/metadata/sources.py
zaneselvans, line 1: There's a lot of lines touched in here, but it's really just about the indentation, and conversion from a module-level constant to a function, which allows us to avoid pulling data from S3 at import.

src/pudl_archiver/metadata/pudl.py
```diff
@@ -0,0 +1,108 @@
+"""Lazy loader for the PUDL datapackage descriptor."""
+
+import functools
+import json
+import os
+
+from upath import UPath
+
+# PUDL_DATAPACKAGE_URL = "s3://pudl.catalyst.coop/nightly/pudl_parquet_datapackage.json"
+PUDL_DATAPACKAGE_URL = "gs://test.catalyst.coop/datapackage.json"
```
zaneselvans: TEMPORARY FOR TESTING. Swap it back before merging or after we've got a real nightly output.

src/pudl_archiver/metadata/pudl.py
```diff
@@ -0,0 +1,108 @@
...
+
+# Static license descriptors (frictionless License fields: name, title, path).
+# Keyed by the short identifiers used throughout the archiver.
+LICENSES: dict[str, dict[str, str]] = {
```
zaneselvans: These are totally static and there's not many of them, so it seemed silly to pull them from the compiled list of licenses in the datapackage.

src/pudl_archiver/metadata/pudl.py
```diff
@@ -0,0 +1,108 @@
...
+        "path": "https://creativecommons.org/publicdomain/zero/1.0/",
+    },
+    "us-govt": {
+        "name": "other-pd",
+        "title": "U.S. Government Work",
+        "path": "http://www.usa.gov/publicdomain/label/1.0/",
+    },
+}
+
+KEYWORDS: dict[str, list[str]] = {
```
zaneselvans: These were the only KEYWORDS elements we were actually referencing so I vendored them. Not ideal but better than the deep entanglement before!

src/pudl_archiver/metadata/pudl.py
```diff
@@ -0,0 +1,108 @@
...
+
+@functools.lru_cache(maxsize=1)
+def get_sources() -> dict[str, dict]:
+    """Return PUDL data sources keyed by name.
+
+    Replaces ``pudl.metadata.sources.SOURCES``.
+    Each value is a dict with at minimum: ``name``, ``title``, ``path``,
+    ``description``, ``keywords``.
+    """
+    return {source["name"]: source for source in get_datapackage()["sources"]}
```
zaneselvans: Need to use comprehensions here and below because we use a dictionary rather than an array in our metadata.

src/pudl_archiver/metadata/pudl.py
```diff
@@ -0,0 +1,108 @@
...
+    Checks the ``PUDL_DATAPACKAGE_PATH`` environment variable first; if set,
+    reads from that local path.  Otherwise fetches from the public S3 URL
+    :data:`PUDL_DATAPACKAGE_URL`.
+
+    Raises:
+        RuntimeError: if the descriptor cannot be read from either source.
+    """
+    path = UPath(os.getenv("PUDL_DATAPACKAGE_PATH", PUDL_DATAPACKAGE_URL))
+    try:
+        return json.loads(path.read_text())
```
zaneselvans, line 29: Right now just treating this as a big nested python dictionary, since we no longer have access to the PUDL metadata classes.  I'd kinda like to have more data validation since not all of the data source metadata in this repo comes from PUDL, but also not thrilled about having two parallel data validation systems that could drift apart.  This seemed like the simpler option for the moment.

src/pudl_archiver/metadata/pudl.py
```diff
@@ -0,0 +1,108 @@
...
+        "united states",
+        "us",
+        "usa",
+        "government",
+        "federal",
+    ],
+}
+
+
+@functools.lru_cache(maxsize=1)
```
zaneselvans, line 12: Caching all of these so we aren't constantly pulling the `datapackage.json` from S3.  But it'll still get pulled once per process, which is gonna be a lot. Alternatively we could cache it on disk in the repo here and update it once a week or on demand with a little github action.

src/pudl_archiver/archivers/ferc/xbrl.py
zaneselvans, line 1: Type annotation changes due to Python 3.14 update.

src/pudl_archiver/archivers/validate.py
zaneselvans, line 1: Type annotation changes due to Python 3.14 update.

src/pudl_archiver/depositors/zenodo/depositor.py
zaneselvans, line 1: Type annotation changes due to Python 3.14 update.

src/pudl_archiver/depositors/depositor.py
zaneselvans, line 1: Type annotation changes due to Python 3.14 update.

src/pudl_archiver/depositors/fsspec.py
zaneselvans, line 1: Type annotation changes due to Python 3.14 update.

pyproject.toml
```diff
@@ -120,33 +120,29 @@ multiline-quotes = "double"
...
 channels = ["conda-forge"]
-platforms = ["linux-64", "osx-64", "osx-arm64"]
+platforms = ["linux-64", "linux-aarch64", "osx-64", "osx-arm64"]
 channel-priority = "strict"
 preview = ["pixi-build"]
 
 [tool.pixi.pypi-dependencies]
 pudl_archiver = { path = ".", editable = true }
-# furo version on conda-forge is very out of date and it is pure python.
-furo = ">=2025.12.19, <2026"
```
zaneselvans, line 130: Sadly, we have no documentation in here yet, so no furo required.

pyproject.toml
```diff
@@ -120,33 +120,29 @@ multiline-quotes = "double"
...
-# semantic-version is not available on conda-forge.
-semantic-version = ">=2.10.0, <3"
```
zaneselvans, line 132: `semantic_version` is now available through `conda-forge`

pyproject.toml
```diff
@@ -120,33 +120,29 @@ multiline-quotes = "double"
 
 [tool.pixi.workspace]
 channels = ["conda-forge"]
-platforms = ["linux-64", "osx-64", "osx-arm64"]
+platforms = ["linux-64", "linux-aarch64", "osx-64", "osx-arm64"]
```
zaneselvans, line 123: Added `linux-aarch64` to the environment to see if we'd be able to run this in an Orbstack docker container, and yes!

pyproject.toml
```diff
@@ -2,7 +2,7 @@
 name = "pudl_archiver"
 version = "0.2.0"
 authors = [{ name = "Catalyst Cooperative", email = "pudl@catalyst.coop" }]
-requires-python = ">=3.13"
+requires-python = ">=3.14"
```
zaneselvans, line 5: Bump to Python 3.14 since we're no longer constrained by `catalystcoop.pudl` and it turns out all of our dependencies in here are fine with it.

pyproject.toml
```diff
@@ -177,44 +170,42 @@ ci = { depends-on = [
...
-numpy = ">=2.4.3,<3"
-pandas = ">=2.3.3,<3"
-pre-commit = ">=4.5.1,<5"
-psycopg2-binary = ">=2.9.10,<3"
-pyarrow = ">=20.0.0,<21"
-pydantic = ">=2.13.2,<3"
+frictionless = ">=5.19.0,<6"
+fsspec = ">=2026.4.0,<2027"
+google-cloud-storage = ">=3.10.1,<4"
+pandas = ">=3.0.3,<4"
```
zaneselvans, line 188: I was nervous about bumping the pandas version, but we don't use it very heavily in here and it seems to work fine.

pyproject.toml
e-belfer: I think some of these implicit dependencies got explicitly added to resolve some crankiness when PUDL was still using mamba and pudl-archiver was migrating to pixi, so if this hasn't broken anything that should be fine.

src/pudl_archiver/depositors/zenodo/entities.py
```diff
@@ -8,12 +8,11 @@
 import re
 from typing import Annotated, Literal
 
-from pudl.metadata.classes import Contributor, DataSource
-from pudl.metadata.sources import SOURCES
 from pydantic import BaseModel, Field, StringConstraints, field_validator
 
 from pudl_archiver.depositors.depositor import DepositionState
-from pudl_archiver.metadata.sources import NON_PUDL_SOURCES
+from pudl_archiver.metadata.pudl import get_sources
```
e-belfer: Calling this `get_pudl_sources()` would make the difference clearer throughout.

src/pudl_archiver/metadata/pudl.py
e-belfer: That makes sense, but to me these make more sense to live in the now jettisoned `constants.py` or in `sources.py` as they aren't PUDL specific. Same with the keywords below.

src/pudl_archiver/metadata/pudl.py
e-belfer: I seem to remember using many other of the keywords in other data archives - do you mean that these are the only ones used in non-PUDL data sources?

src/pudl_archiver/metadata/pudl.py
e-belfer: It's uncommon that we add a new archiver to PUDL, but when we do we want to basically immediately run the archiver after the metadata merges into PUDL. This method avoids the user having to remember to update the file after run. If it becomes a slog, I could also see building in the cache update check into the weekly GHA run. For now, if this isn't a performance barrier this seems ok to me.

@zschira might also have thoughts here?

src/pudl_archiver/metadata/pudl.py
e-belfer: Yes, I don't love this. If there's a more elegant solution that comes to mind I'm all ears, otherwise I think I'd propose splitting out the follow-up validation work into a separate issue given the time sensitivity of making this fix.

src/pudl_archiver/metadata/sources.py
```diff
@@ -1,11 +1,12 @@
 """Metadata and operational constants."""
 
+import functools
 from typing import Any
 
 import pandas as pd
-from pudl.metadata.constants import CONTRIBUTORS, KEYWORDS, LICENSES
```
e-belfer, line 6: See my comment above re: this reorganization.

src/pudl_archiver/metadata/sources.py
e-belfer: Why would moving this constant to a function help avoid importing data from S3? Even if you want to cache the function call (which I'm still a bit confused about why that's necessary in this case), you could still call `return NON_PUDL_SOURCES` and I think it'd be a more legible function than defining a massive dictionary in `return`.

e-belfer (changes requested):
Some organizational questions and hesitations. ~~However, my big concern is that at the moment running this on production every archiver is failing: https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/26404556729/job/77724685591~~

Update: the rerun with the interim datapackage file added to the nightly builds (see https://github.com/catalyst-cooperative/pudl/issues/5270) has worked as expected. The remaining request changes are organizational at this point: https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/26408037211/job/77735969178

The unit tests are also expecting an exact message for the failure which has now changed.

pyproject.toml
zaneselvans: It seems to be working fine so I assume nothing is missing! I also ended up removing all the Sphinx related packages since we do not yet have a docs build in here (though we should set one up!)

src/pudl_archiver/metadata/sources.py
```diff
@@ -23,527 +24,542 @@
...
-            "years": [
-                2003,
-                2012,
-                2018,
-            ]  # there are PDF only versions for 1999, 1995 and 1992
+
+@functools.lru_cache(maxsize=1)
+def get_non_pudl_sources() -> dict[str, Any]:
+    """Return metadata for datasets archived by PUDL but not processed into it."""
+    catalyst = get_contributors()["catalyst-cooperative"]
```
zaneselvans: @e-belfer The `KEYWORDS` and `LICENSES` are imported constants but we're still pulling the list of contributors from the PUDL `datapackage.json` which is being pulled with this `get_contributors()` call, which is why this becomes a function. Admittedly, we're only referring to the `catalyst-cooperative` contributor, which we could hard-code as a constant too.  But we have access to the other PUDL metadata now so I went that route so we can reference other data contributors in here easily.

src/pudl_archiver/metadata/sources.py
zaneselvans: Ah, sorry there's one call of `get_contributors()` inside the function below, which is what necessitates the functionalization. I'll comment on it down there.

src/pudl_archiver/depositors/zenodo/entities.py
zaneselvans: I'll switch `get_contributors()` to `get_pudl_contributors()` too.

src/pudl_archiver/metadata/pudl.py
zaneselvans: There's a big collection of thematic keyword groupings in the PUDL repo, which we use to annotate the datasets over there, but because the thematic groups aren't part of the exported metadata / datapackage.json we don't have access to them.  These were the only ones that we were using in here.

src/pudl_archiver/metadata/pudl.py
zaneselvans: The `@lru_cache()` is caching function calls within a particular Python process, not persisting the datapackage anywhere. This lets us download the metadata once per process, and then refer to it anywhere, but without initializing something at import-time, which would be annoying given that it requires network access. The PUDL `datapackage.json` is like 3-4MB.

src/pudl_archiver/metadata/pudl.py
zaneselvans: I think some much lighter-weight Pydantic classes or dataclasses or typed dictionaries that declare the shape, types, and required fields is probably plenty. Will make a followup issue.

src/pudl_archiver/metadata/pudl.py
zaneselvans: Yeah they don't really belong here... should I go ahead and add a constants.py?  That must have been jettisoned before this PR.

src/pudl_archiver/metadata/sources.py
e-belfer: I'm still not loving this giant return call as an outcome of the reorganization, and I wonder whether even a simple key replace or redefining the catalyst cooperative contributor in this repository is a better solution. @zschira maybe you have thoughts here?

e-belfer (commented):
The run produced expected changes, so it looks like things are working as expected here: https://github.com/catalyst-cooperative/pudl-archiver/issues/1118

My one big hesitation at this point is to do with the `get_non_pudl_sources()` call. I'd lean towards redefining the Catalyst contributor in this repository or doing some strategic search and replace of a placeholder "catalyst-cooperative" contributor key in the function rather than creating a massive dictionary in the return call.

Also note that these changes will definitely impact the README instructions on how to add a new dataset, so once everything's settled here you'll want to make a cleanup pass there to be sure nothing is out of date.

Adding the additional lines to the error report means that **every** archive validation failure is now also marked as an **other issue** in #1118 - this is definitely not a desired outcome of this PR and should get fixed.

src/pudl_archiver/metadata/pudl.py
zaneselvans: Moved to constants.py

zaneselvans:
@e-belfer Can you say more about what you don't like about having this metadata inside of a function rather than in a module level constant (which precludes any information coming from the PUDL metadata without additional processing of the constant)?

Having a constant that gets munged with some `dict.update()` or search & replace seems like it could be conceptually confusing -- it would be easy to see the constant, import it elsewhere, and assume that you've got the metadata, only to later realize that it's not valid or complete without the post processing.  If we really need the metadata to live in a module-level constant I think just redefining the `catalyst-cooperative` contributor as a constant feels simpler and would definitely work for now, but if we ever wanted to pull in any metadata fields that are available from PUDL in the future we'd run into the same issue again.

On the error types / extra lines I was trying to surface the nature of the validation failures in the logs that are visible on GitHub for debugging purposes -- before I was just getting "Validation failed" with no information about what dataset or partition resulted in the failure or what kind of failure it was.  Is it literally the additional newlines that's causing the problem? Or including the additional information at all? Is it actually parsing the logs rather than structured run summaries?

I could also `logger.error()` the validation failure information rather than putting it in the exception. Would that work better?

zaneselvans:
@e-belfer Based on [this run](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/26481876400) and [the resulting issue](https://github.com/catalyst-cooperative/pudl-archiver/issues/1121) it looks like the validation failures are being categorized correctly now with the right matching string, so hopefully this is ready to merge!

src/pudl_archiver/depositors/zenodo/entities.py
```diff
@@ -112,57 +111,48 @@ def check_empty_string(cls, doi: str):  # noqa: N805
...
-            ]
-
         if data_source_id in ["gridpathratoolkit", "vcerare"]:
-            # If data source was manually archived by us, specify that the
-            # data_source.path is a documentation link, rather than where we archived
-            # the data from. These datasets are highly processed, so we also drop the
-            # 'PUDL Raw' prefix from the title to avoid confusion.
-            title = f"{data_source.title}"
+            # data_source.path is a documentation link, not the archive source.
+            # Drop the 'PUDL Raw' prefix — these datasets are highly processed.
```
e-belfer: Non-blocking: not sure why this comment got re-written, but the new comment is I think less clear than the one that existed before.

src/pudl_archiver/metadata/constants.py
```diff
@@ -0,0 +1,70 @@
...
+    ],
+    "us_govt": [
+        "united states",
+        "us",
+        "usa",
+        "government",
+        "federal",
+    ],
+}
+"""Topical lists of keywords not availble in this from from the PUDL datapackage metadata."""
```
e-belfer:
```suggestion
"""Topical lists of keywords used in this repository (copied from the PUDL datapackage metadata)."""
```
Non-blocking: This sentence seems unfinished - is this more like what you meant?

src/pudl_archiver/metadata/pudl.py
e-belfer: @zaneselvans Reminder to make a follow-up issue and link it here.

README.md
```diff
@@ -37,24 +37,25 @@ uploaded to Zenodo.
 ## Installation
 
 We use `pixi` to create and manage the `pudl-archiver` development environment.
+The project uses Python 3.14+, which is managed automatically by pixi.
```
e-belfer: Any reason to specify the version? It'll certainly go out of date at some point.

e-belfer (approved):
Little comments about documentation that you can take or leave, but the issue created looks good. Thanks for fixing this!

src/pudl_archiver/metadata/pudl.py
zaneselvans: See #1122

PR #1164, opened 2026-06-15
Switch from using Slack to Zulip for notifications

zaneselvans:
# Overview

- Update the `run-archiver`, `run-fsspec-archiver`, and `initialize-archiver` workflows to notify Zulip not Slack.
- Also send Zulip notifications at workflow launch.
- Use a single `make_markdown_notification.py` script since both GitHub and Zulip want markdown.
- Fix some `actionlint` pre-commit hook complaints in the workflow files.
- Removed the overly noisy codemonkeys slack notification about `run-pytest.yml` outcomes.

# Testing

- I did several test runs with the `run-archiver` action and a single dataset, with notifications going to Slack or creating issues on GitHub.

# To-do list


.github/workflows/initialize-archiver.yml
```diff
@@ -19,7 +19,38 @@ on:
...
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          TARGET_SERVER: ${{ inputs.server }}
+          WORKFLOW_NAME: ${{ github.workflow }}
+        run: |
+          {
+            echo "LAUNCH_CONTENT<<EOF"
+            echo "# :rocket: Launching \`$WORKFLOW_NAME\` $TARGET_SERVER workflow ([view run]($RUN_URL))"
+            echo
+            echo "Datasets:"
+            printf '%s\n' "$DATASETS_JSON" | jq -r '.[] | "- `\(.)`"'
```
zaneselvans, line 38: Makes a list of the datasets being archived.

.github/workflows/run-archiver.yml
```diff
@@ -174,42 +207,45 @@ jobs:
         run: |
           {
             echo 'ERRORS<<DELIM1'
-            pixi run ./scripts/make_github_notification.py --summary-files *_run_summary.json --error-files *_log.txt --summary-type error
+            pixi run ./scripts/make_markdown_notification.py --summary-files ./*_run_summary.json --error-files ./*_log.txt --summary-type error
```
zaneselvans, line 210: I consolidated the github + slack message generation scripts, since both GitHub + Zulip want markdown now.

.github/workflows/run-pytest.yml
```diff
@@ -48,27 +48,3 @@ jobs:
 
       - name: Upload test coverage report to CodeCov
         uses: codecov/codecov-action@v7
-
-  ci-notify:
```
zaneselvans, line 52: This was an overly spammy notification in Slack, so I discontinued it.

scripts/make_slack_notification_message.py
zaneselvans, line 1: Both Zulip and GitHub want Markdown, so we can just use one script now.

scripts/make_markdown_notification.py
```diff
@@ -208,6 +307,17 @@ def main(summary_files: list[Path], error_files: list[Path], summary_type: str)
         print(failed_blocks)
     elif summary_type == "unchanged":
         print(unchanged_blocks)
```
zaneselvans, line 306: These per-type outputs remain broken out so they can be inserted into the GitHub issue template, which has additional text / instructions beyond what shows up in the Zulip notification.

.github/workflows/initialize-archiver.yml
krivard: super cute!

.github/workflows/initialize-archiver.yml
```diff
@@ -97,20 +128,25 @@ jobs:
...
       - name: show summaries
         run: ls -R
-      - name: Munge summaries together
-        id: all_summaries
+      - name: Load pixi from cache
+        uses: prefix-dev/setup-pixi@v0.9.6
+      - name: Create Zulip notification content
+        id: zulip_summaries
+        env:
+          WORKFLOW_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
```
krivard: nice addition

.github/workflows/run-archiver.yml
```diff
@@ -60,9 +60,42 @@ jobs:
...
+  notify-start:
+    runs-on: ubuntu-latest
+    needs: determine-datasets
+    steps:
+      - name: Build launch notification
+        id: launch_message
+        env:
+          DATASETS_JSON: ${{ needs.determine-datasets.outputs.datasets }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+          TARGET_SERVER: ${{ github.event_name == 'schedule' && 'production' || inputs.server }}
```
krivard, line 72: a little brainbendy but i'll allow it

.github/workflows/run-archiver.yml
```diff
@@ -60,9 +60,42 @@ jobs:
...
 
+  notify-start:
+    runs-on: ubuntu-latest
+    needs: determine-datasets
+    steps:
+      - name: Build launch notification
+        id: launch_message
+        env:
+          DATASETS_JSON: ${{ needs.determine-datasets.outputs.datasets }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
```
krivard, line 71: should we make the variable name for this the same as it is in `initialize-archiver.yml` to avoid confusion?

scripts/make_markdown_notification.py
```diff
@@ -47,24 +42,34 @@ def _parse_args():
...
-    a checkbox.
+    When archives fail, they may not have a URL.
     """
-    header = f"[**{name}**]({url})" if url else f"**{name}**"
+    header = f"### [{name}]({url})" if url else f"### {name}"
 
-    if action:
-        return f"{header}\n\n{content}\n\n- [ ] {action}"
+    if action and include_action:
+        action_prefix = "- [ ]" if include_action else "-"
```
krivard: when do you expect the `else` clause to be used? is `include_action` not guaranteed to be truthy inside this block?

scripts/make_markdown_notification.py
```diff
@@ -47,24 +42,34 @@ def _parse_args():
...
     return parser.parse_args()
 
 
 def _format_message(
-    url: str | None, name: str, content: str, action: str | None = None
+    url: str | None,
+    name: str,
+    content: str,
+    action: str | None = None,
+    include_action: bool = True,
```
krivard: hmm I'm not convinced `include_action` is needed, since `False and "something"` and `None and "something"` are both falsy and `True and "something"` yields `"something"`. I don't feel strongly about it though so if you find it more clear to separate them, keep it

scripts/make_markdown_notification.py
```diff
@@ -144,7 +172,7 @@ def _format_errors(log: str) -> str | None:
         r"(?:catalystcoop.pudl_archiver.archivers.classes:\d+ Archiving )([a-z0-9]*)",
         log,
     )
-    name = name_re.group(1)
+    name = name_re.group(1) if name_re else "Unknown"
```
krivard, line 172: nice 👍

scripts/make_markdown_notification.py
```diff
@@ -157,10 +185,55 @@ def _format_errors(log: str) -> str | None:
...
+    if changed_blocks:
+        parts.append("## Changed")
+        parts.append(changed_blocks)
+
+    if unchanged_blocks:
+        parts.append("## Unchanged")
+        parts.append(unchanged_blocks)
+
+    if not any([error_blocks, failed_blocks, changed_blocks, unchanged_blocks]):
+        parts.append("_No downloaded summary or error artifacts were found._")
```
krivard: Zulip-flavored markdown doesn't recognize underscores -- if you want italics try a single star; that works in both GitHub and Zulip

krivard (approved):
This can merge as-is; all recs optional:

There are two different variable names being used to refer to the same string: RUN_URL and WORKFLOW_URL -- it might be better to settle on just one.

Underscores in markdown have variable support -- it might be better to stick to asterisk notation.

I have a preference for boolean short-circuiting over creating additional arguments, but admittedly it's a slightly unhinged preference and it's okay if using explicit arguments is clearer for everybody else

There's a ternary in here we might not need

.github/workflows/run-archiver.yml
zaneselvans: Okay, I standardized on `RUN_URL` and `--run-url` because it's actually a link to a particular run, not the workflow overall.

scripts/make_markdown_notification.py
zaneselvans: Ah. I think probably this didn't get exercised in the test runs so I never saw that output.

.github/workflows/run-archiver.yml
zaneselvans: I found all of these little logical naming snippets annoying. :(

scripts/make_markdown_notification.py
zaneselvans: I simplified this by moving the "should we include an action?" decision up to the individual message type formatters -- now the checkboxes are included if we have an action at all, and there's no list of actions in the Zulip case, which is what we really want.

scripts/make_markdown_notification.py
zaneselvans: Eliminated.

PR #1190, opened 2026-07-01
Archive both monthly and annual EIA 191 data

### LINKED ISSUE #1117: Update EIA 191 archiver to grab annual data
Currently we only archive EIA 191 monthly report data. However, December snapshots are available back to 2005 (an additional 9 years of data). To support integration of them into PUDL (https://github.com/catalyst-cooperative/pudl/issues/5276), we need to archive this data first.

A starter redefinition of this archiver could look something like this (with redefinition of the `get_year_resource()` still to be implemented):

```
        """Download EIA 191 resources, keeping monthly data only."""
        reports_list = await self.get_reports(url=self.base_url, form=self.form)

        # Annual and monthly data have different date ranges, but we want to end up
        # with one zipfile per year with one file per data format.
        # To do this, we iterate through the reports to get the total date range
        # and frequencies available for each report.
        freq_dict = defaultdict(set)
        for report in reports_list:
            for year in report.available_years:
                freq_dict[year.ayear].add(report.report_code)

        for year in freq_dict:
            yield self.get_year_resource(year, reports_list, freq_dict[year])
```

e-belfer:
# Overview

Closes #1117 

Right now we're only archiving EIA 191 monthly data, but annual data goes further back in time. For some years, both datasets are reported. This archiver adds all December snapshot data files to the yearly zipfile to enable downstream integration of more complete timeseries.

Adapt the EIA 191 archiver to check to see what reports are available for each year, then adding both of those reports into one zipfile.

# Testing

https://github.com/catalyst-cooperative/pudl-archiver/issues/1212

# To-do list

- [x] Once oracle failure is resolved, test this locally and on Zenodo.
- [x] Once reviewed, create prod archive.


cmgosnell (approved):
looks good! I built a sandbox archive of this locally and got expected errors that the archive got bigger. But spot checked the yearly resources and it all looks good to me.

PR #1208, opened 2026-07-06
Grab missing versions of usgsuswtdb

e-belfer:
# Overview

An incomplete regex query missed versions on https://www.sciencebase.gov/catalog/item/5e99a01082ce172707f6fd2a that used V instead of v.

Updated the regex query to include both v and V.

# Testing

See #1209 and https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/28800063953

# To-do list


cmgosnell (approved):
V|very good

PR #1213, opened 2026-07-07
Improve ability to retry `fsspec` backed archivers

zschira:
# Overview

The EQR archiver has proven to be quite flaky. Despite our existing retries / backout system for file downloads, we often encounter a persistent timeout when downloading EQR zipfiles. We do have a `retry-run` command that theoretically would allow us to pick back up from failures, but because the timeouts were causing an exception, we ended up not actually writing the run summary JSON file that we would need to do so. This PR improves our error handling so we fail gracefully and produce a run summary JSON file in these cases and provides a builtin way to retry runs directly through github.

# What did you change in this PR?
- Handle failures during scraping / uploads so we can retry more failed runs
- Migrate `publish-fsspec-archive` workflow to `publish-or-retry-fsspec-archive` so it can be used to retry runs directly through github
- Make `run-fsspec-archiver` only run a single archiver at a time (that's all we do in practice right now anyway) to make retries more straightforward
- Make `publish-or-retry-fsspec-archive` automatically attempt a retry if `run-fsspec-archiver` fails

# Testing

Tested a retrying an EQR run through github [here](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/28819569477)


.github/workflows/publish-or-retry-fsspec-archive.yml
zschira, line 1: Unfortunately, github doesn't like to show nice diffs for renamed files. To view the exact diff you can use the following command:

```
git diff main:.github/workflows/publish-fsspec-archive.yml eqr-retries:.github/workflows/publish-or-retry-fsspec-archive.yml
```

The changes in this file include:
1. Make this file capable of retrying runs and publishing through the `command` input
2. Make the workflow trigger a retry when the run-fsspec-archiver` workflow fails
3. Upload run summaries so we can run this workflow again to publish a deposition after a retry

.github/workflows/run-fsspec-archiver.yml
```diff
@@ -4,9 +4,9 @@ name: run-fsspec-archiver
 on:
   workflow_dispatch:
     inputs:
-      datasets:
-        description: 'Comma-separated list of datasets to archive (e.g., "ferceqr","sec10k").'
-        default: '"ferceqr"'
+      dataset:
```
zschira, line 7: It makes it a lot simpler to kick off a retry automatically if we just run one `dataset` at a time.

src/pudl_archiver/archivers/classes.py
```diff
@@ -781,42 +781,37 @@ async def _unpack_resources(
 
     async def _filter_resources(
         self,
-        retry_parts: list[Partitions] | None = None,
+        skip_partitions: list[Partitions],
```
zschira, line 784: If a run fails early, then we don't always know which partitions didn't complete successfully, so it's easier / better to specify the partitions NOT to retry instead of the ones to retry.

There are changes throughout switching to use the new `skip_partitions`.

src/pudl_archiver/archivers/validate.py
```diff
@@ -48,6 +48,21 @@ class FileUniversalValidation(ValidationTestResult):
     resource_name: Path
 
 
+def exception_validation(e: Exception | None) -> DatasetUniversalValidation:
```
zschira, line 51: This is a new validation test that I added to guarantee that when we catch an exception during a run that we mark that run as having failed.

src/pudl_archiver/archivers/validate.py
```diff
@@ -220,7 +235,7 @@ def load_previous_run(
                 "run_settings": previous_run_summary.run_settings.model_copy(
                     update={
                         "retry_run": summary_file,
-                        "auto_publish": True,
+                        "auto_publish": auto_publish,
```
zschira, line 238: This was a bug where we weren't properly setting `auto_publish` for retries.

src/pudl_archiver/cli.py
```diff
@@ -174,7 +174,7 @@ def publish_run(summary_file: str):
     failed_partitions = previous_run_summary.failed_partitions
     successful_partitions = previous_run_summary.successful_partitions
 
-    if failed_partitions != {}:
+    if (failed_partitions != {}) or (not previous_run_summary.success):
```
zschira, line 177: Sometimes in a failed run we don't record any `failed_partitions` because we actually caught an exception so many partitions weren't tried at all.

src/pudl_archiver/orchestrator.py
```diff
@@ -18,32 +18,31 @@ async def orchestrate_run(
...
     # Download resources and add to archive
-    async for name, resource in downloader.download_all_resources(
-        retry_parts=failed_partitions
-        if failed_partitions is None
-        else list(failed_partitions.values()),
-    ):
-        resources[name] = resource
-        draft = await draft.add_resource(name, resource)
+    run_exception = None
+    try:
```
zschira, line 30: This is where we're actually catching exceptions to fail more gracefully.

.github/workflows/publish-or-retry-fsspec-archive.yml
zaneselvans: I've had good luck getting it to track moved files using `git mv oldname newname` which preserves the commit history and gives nice diffs.

.github/workflows/publish-or-retry-fsspec-archive.yml
```diff
@@ -0,0 +1,99 @@
...
+
+      - name: Additional Playwright setup
+        run: |
+          pixi run playwright install --with-deps webkit
+          pixi run playwright install --with-deps chromium
+
+      - name: Download run summary
+        uses: actions/download-artifact@v8
+        with:
+          run-id: "$RUN_ID"
```
zaneselvans: Have you tried running a `publish-run` job? I don't think the bare envvar will work here, will it? I think you need something inside the doublecurlies like this maybe:

```suggestion
          run-id: ${{ env.RUN_ID }}
```

Otherwise I think you get a literal `$RUN_ID` But I could be wrong. I get all these different interpolation contexts confused in the workflow files.

.github/workflows/publish-or-retry-fsspec-archive.yml
```diff
@@ -0,0 +1,99 @@
...
+          pattern: run-summaries-*
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Get dataset from run summary filename
+        id: dataset_name
+        run: |
+          file=$(find summaries -name '*_run_summary.json')
```
zaneselvans: Can we guarantee that there will always be one and only one run summary? Or do we need to check the results of the glob?  Where is the `summaries` directory being created?

src/pudl_archiver/archivers/classes.py
zaneselvans, line 824: Is it ever possible to get here with chunksize == 0?

E.g. in an archiver with no `self.concurrency_limit` can `len(resources)` ever be zero? If so then do we get a 0 / 0 error here? It looks like the default concurrency limit is `None` and many archivers don't have one set explicitly. It seems like previously this wouldn't have been a problems because if `len(retry_parts) == 0` then we would have returned early.

.github/workflows/run-fsspec-archiver.yml
zaneselvans: Yeah I think we've generally come around to having each dataset run independently on its own runner right?

src/pudl_archiver/orchestrator.py
```diff
@@ -18,32 +18,31 @@ async def orchestrate_run(
...
+        async for name, resource in downloader.download_all_resources(
+            skip_partitions.values(),
+        ):
+            resources[name] = resource
+            draft = await draft.add_resource(name, resource)
+    except Exception as e:
+        run_exception = e
+        logger.error(f"download_all_resources failed!\n{e}")
```
zaneselvans: I recently discovered `logger.exception(e)` which automatically includes the error message and traceback and logs at ERROR level, which is convenient and informative.

.github/workflows/publish-or-retry-fsspec-archive.yml
```diff
@@ -0,0 +1,99 @@
...
+    types:
+      - completed
+  workflow_dispatch:
+    inputs:
+      run_id:
+        description: Github action run ID, which is the number at the end of the URL when looking at a run
+        required: true
+        type: string
+      command:
+        description: Switch between the retry-run and publish-run command
```
zaneselvans: Can you be more explicit about what happens on a retry? If a retry is successful will it also publish? Or do we need to run retry, and then also run publish?

.github/workflows/publish-or-retry-fsspec-archive.yml
```diff
@@ -0,0 +1,99 @@
+---
+name: publish-or-retry-fsspec-archive
+
+on:
+  # Retry failed runs of fsspec archiver
+  workflow_run:
+    workflows: ["run-fsspec-archiver"]
```
zaneselvans, line 7: If this is kicked off by the completion of `run-fsspec-archiver` and that workflow failed, then it will do a `retry-run` one time, trying to complete the previous run, and if that also fails then it's done and we have to come try a manual re-run.  Is that right?

zaneselvans (commented):
What kinds of failed runs have you tried recovering with the new setup?

Things that should be tested if they haven't been already:

* We allow archivers / datasets to be unpartitioned, with `"working_partitions": {}`  Which I think means you could have empty `skip_partitions` regardless of whether the archiving run succeeded or failed. How do these changes affect those datasets/archivers, if at all?
* Does the `$RUN_ID` in the `publish-run` job actually expand to the value of that envvar?
* Are we sure there's never any way to get a divide by zero error with the new chunksizse calculation?
* Have you been able to successfully obtain a run summary artifact in `publish-or-retry-fsspec-archive`?  The find command references a `summaries` directory and I'm not seeing where that directory is created. Is it ever possible to get more than one run summary file in that find? If so I think it'll break. I'm not seeing that step reflected in the logs of the successful `publish-run` so I'm suspicious it may not have been exercised.

.github/workflows/publish-or-retry-fsspec-archive.yml
zschira: It used to be possible that it would generate more than one run summary if we ran multiple archivers at once, but since I changed the `run-fsspec-archiver` workflow to only run one archiver at a time, it will only produce one summary file now.

src/pudl_archiver/archivers/classes.py
zschira: Ah yeah good catch. This function still gets called in the case where we are trying to publish a run, in which case we don't expect to be downloading any new resources. I've fixed this to return immediately if there are no resources to download.

src/pudl_archiver/orchestrator.py
zschira: Nice! Switched to using this

.github/workflows/publish-or-retry-fsspec-archive.yml
zschira: Added more detail to this description

zschira:
> We allow archivers / datasets to be unpartitioned, with "working_partitions": {} Which I think means you could have empty skip_partitions regardless of whether the archiving run succeeded or failed. How do these changes affect those datasets/archivers, if at all?

In the case that `skip_partitions` is empty it will retry every partition, so in this case it would just retry the empty partition `{}`. I've added a couple extra integration tests to validate the retry behavior, including one that tests an unpartitioned dataset and it works as expected.

> Have you been able to successfully obtain a run summary artifact in publish-or-retry-fsspec-archive? The find command references a summaries directory and I'm not seeing where that directory is created. Is it ever possible to get more than one run summary file in that find? If so I think it'll break. I'm not seeing that step reflected in the logs of the successful publish-run so I'm suspicious it may not have been exercised.

Ah yeah the `summaries` directory was a holdover and is removed now. I had previously tested using the old `publish-fsspec-archiver` workflow before I changed the name, so I wouldn't have to add a stub workflow, but there's been some minor deviations since then. I've now added a stub for testing and have a successful `publish` run [here](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/29838495878).

tests/integration/fsspec_depositor_test.py
zaneselvans, line 1: Thanks for adding these new integration tests.  We might also want to test the `publish-run` case. Adapting the pattern you've already got in here maybe that would look like this?

```python
@pytest.mark.asyncio
async def test_publish_run(
    good_zipfile,
    fixed_bad_zipfile,
    tmp_path,
    datasource: dict,
    mocker,
):
    """Test the ``publish-run`` path for a fully successful, unpublished run.

    This mimics the ``pudl_archiver publish-run`` CLI command: a first run downloads
    every partition successfully with ``auto_publish=False``, so the new files land in
    the workspace directory but are never published. A second run with
    ``skip_partitions`` set to *all* of the first run's ``successful_partitions``,
    like ``publish-run`` does when it calls the archiver again using the previous
    run summary. Since every partition gets skipped, ``download_all_resources`` should
    filter down to zero resources and take its early-return path and the already-downloaded
    workspace files should get published.
    """
    deposition_path = tmp_path / "deposition"
    deposition_path.mkdir()

    settings = RunSettings(
        clobber_unchanged=True,
        auto_publish=False,
        refresh_metadata=False,
        initialize=True,
        depositor="fsspec",
        depositor_args={"deposition_path": str(deposition_path)},
    )
    good_part = {"part": "good_part"}
    other_part = {"part": "other_part"}

    class TestDownloader(AbstractDatasetArchiver):
        name = "Test Downloader"

        def __init__(self, should_download: bool, **kwargs):
            super().__init__(**kwargs)
            self.should_download = should_download

        async def get_resources(self):
            yield self.get_zipfile(good_zipfile, parts=good_part), good_part
            yield self.get_zipfile(fixed_bad_zipfile, parts=other_part), other_part

        async def get_zipfile(self, zip_path, parts):
            if not self.should_download:
                # publish-run should never actually download a partition that was
                # already downloaded and left in the workspace by a previous run.
                raise RuntimeError(
                    f"Unexpectedly tried to (re)download {zip_path.name} during a "
                    "publish-run."
                )
            return ResourceInfo(local_path=zip_path, partitions=parts)

    v1_summary, _ = await orchestrate_run(
        dataset="pudl_test",
        downloader=TestDownloader(should_download=True, session="session"),
        run_settings=settings,
        session="session",
    )
    with (tmp_path / "run_summary.json").open("w") as f:
        f.write(json.dumps([v1_summary.model_dump()], indent=2))

    assert v1_summary.success
    assert v1_summary.failed_partitions == {}
    assert good_part == v1_summary.successful_partitions["good.zip"]
    assert other_part == v1_summary.successful_partitions["bad.zip"]
    assert not (deposition_path / "published" / "good.zip").exists()
    assert not (deposition_path / "published" / "bad.zip").exists()
    assert (deposition_path / "workspace" / "good.zip").exists()
    assert (deposition_path / "workspace" / "bad.zip").exists()

    # Mimic `publish-run`: rerun with auto_publish forced on and every partition from
    # the previous successful run passed in as `skip_partitions`.
    settings = settings.model_copy(update={"auto_publish": True})
    v2_downloader = TestDownloader(should_download=False, session="session")
    v2_summary, _ = await orchestrate_run(
        dataset="pudl_test",
        downloader=v2_downloader,
        run_settings=settings,
        session="session",
        skip_partitions=v1_summary.successful_partitions,
    )
    assert v2_summary.success
    assert (deposition_path / "published" / "good.zip").exists()
    assert (deposition_path / "published" / "bad.zip").exists()
```

It looks like we aren't already testing an attempted publishing run on a failed archive. That's generic to all the archivers. Maybe that goes in the CLI test module and looks like:

```python
@pytest.mark.parametrize(
    "failed_partitions,validation_success",
    [
        ({"eia860-2020.zip": {"year": 2020}}, True),
        ({}, False),
    ],
    ids=["failed_partitions", "failed_validation"],
)
def test_publish_run_rejects_unsuccessful_run(
    failed_partitions, validation_success, tmp_path, mocker
):
    """``publish-run`` should refuse to run against a summary with failures.

    ``publish-run`` is only meant to publish the results of a run that fully
    succeeded but wasn't auto-published. If the previous run has any failed
    partitions, or otherwise didn't pass validation, it should raise rather than
    silently publish an incomplete archive -- the user is expected to use
    ``retry-run`` instead.
    """
    summary = RunSummary(
        dataset_name="test_dataset",
        validation_tests=[
            ValidationTestResult(
                name="test_test",
                description="test of the tests",
                success=validation_success,
            )
        ],
        file_changes=[],
        date="right_now",
        previous_version_date="before_right_now",
        record_url="https://test.com",
        datapackage_changed=True,
        failed_partitions=failed_partitions,
        successful_partitions={},
        run_settings=RunSettings(),
    )
    summary_file = tmp_path / "run_summary.json"
    summary_file.write_text(summary.model_dump_json())

    archive_dataset_mock = unittest.mock.AsyncMock()
    mocker.patch("pudl_archiver.cli.archive_dataset", new=archive_dataset_mock)

    runner = CliRunner()
    with pytest.raises(RuntimeError, match="publish-run should not be called"):
        runner.invoke(
            pudl_archiver,
            ["publish-run", str(summary_file)],
            catch_exceptions=False,
        )

    archive_dataset_mock.assert_not_awaited()
```

src/pudl_archiver/archivers/classes.py
```diff
@@ -812,6 +812,10 @@ async def download_all_resources(
         coordinates downloading all resources concurrently.
         """
         resources = await self._filter_resources(skip_partitions)
+        # When running the publish-run command we should end up with no resources to download
+        if len(resources) == 0:
+            logger.info("Found no resources to download, returning immediately.")
+            return
```
zaneselvans, line 818: Does the fact that this early return comes after the call above to `_filter_resources()` (which calls `_unpack_resources()` which calls `get_resources()`) mean that even in the case of a `publish-run` we have to download all the EQR data anew? If so then it seems like this would mean the `publish-run` would still suffer from network flakiness, rather than just publishing what we've already downloaded. I feel like I must be misunderstanding something about the logic.

zaneselvans (changes requested):
One clarifying question which I suspect has to do with me not understanding async.

:x: Request for some testing of the `publish-run` case. I left possible tests in the comments, created in collaboration with 🤖 (they already pass)

src/pudl_archiver/archivers/classes.py
zschira: When you call an `async` function it returns an `awaitable`, but it doesn't actually do anything until it is awaited. So when we call the download functions in `get_resources` it doesn't actually run them until the `asyncio.as_completed` call later on.

tests/integration/fsspec_depositor_test.py
zschira: Thanks for the test suggestions, added!

zaneselvans (approved):

PR #1226, opened 2026-07-15
Update FERC archivers to work with FERC Online web app

### LINKED ISSUE #1197: `ferc`: FercOnline website requires playwright to access download links for DBF data
See https://forms.ferc.gov/ and https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/28514349658/job/84610120312

zschira:
# Overview

Closes #1197. FERC DBF data is now only available through a JS web-app.

# What did you change in this PR?
This PR uses `playwright` to navigate FERC online and download DBF data. It also adds a way to run the archiver with docker because I have long had issues using playwright on my desktop.

# Testing


scripts/run-unit-tests.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
zschira, line 1: This wrapper script makes it possible to switch between running unit tests with/without docker in our pre-commit hooks

src/pudl_archiver/archivers/ferc/ferc2.py
```diff
@@ -47,51 +29,3 @@ async def get_resources(self) -> ArchiveAwaitable:
...
-
-    async def get_year_dbf(
-        self, year: int, part: int | None = None
-    ) -> tuple[Path, dict]:
-        """Download a single DBF of historical FERC Form 2 data from 1991-2021.
-
-        Source page:
-            https://www.ferc.gov/industries-data/natural-gas/industry-forms/form-2-2a-3-q-gas-historical-vfp-data
-        """
-        early_urls: dict[tuple(int, int), str] = {
```
zschira, line 59: DBF data is no longer split into 2 partitions per year. We'll probably have to make some minor changes in PUDL to accommodate, but it should make it more consistent with other datasets

src/pudl_archiver/archivers/ferc/ferc_online_helpers.py
zschira, line 1: The good thing about the change to this new app is that every dataset is handled exactly the same, so we can use a single helper function for all datasets (including pre-2013 EQR although I've not added that to the EQR archiver yet).

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+WORKDIR /app
+
+COPY pyproject.toml .
+COPY pixi.lock .
+COPY README.md .
+COPY LICENSE.txt .
+
+# Install the pixi environment (includes the editable local package)
+RUN pixi install --locked --environment docker
```
zschira, line 11: Installing the full pixi environment requires that the `src` directory is available when building the docker image, but if we copy `src` in here, then the `pixi install` has to run every time we make any changes, and that's really slow. I've created this `docker` environment (this is defined in `pyproject.toml`) that doesn't actually install `pudl_archiver` so we don't need `src`, but we can still run the full `playwright` install in here.

pyproject.toml
```diff
@@ -124,13 +124,16 @@ platforms = ["linux-64", "linux-aarch64", "osx-64", "osx-arm64"]
 channel-priority = "strict"
 preview = ["pixi-build"]
 
-[tool.pixi.pypi-dependencies]
+[tool.pixi.feature.dev.pypi-dependencies]
```
zschira, line 130: This is where we make `pudl_archiver` installed in the `default` environment, but not the `docker` environment.

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+services:
+  pudl_archiver:
+    build: .
+    env_file:
+      - .env
+    volumes:
+      - .:/app
```
zschira, line 7: Mount the base of the repo so we have access to `src` when actually running this

src/pudl_archiver/archivers/ferc/ferc6.py
```diff
@@ -18,10 +18,13 @@ class Ferc6Archiver(AbstractDatasetArchiver):
 
     async def get_resources(self) -> ArchiveAwaitable:
         """Download FERC 6 resources."""
-        for year in range(2000, 2022):
-            if not self.valid_year(year):
-                continue
-            yield self.get_year_dbf(year)
+        dbf_years = [year for year in range(2000, 2022) if self.valid_year(year)]
+        yield ferc_online_helpers.get_resources_for_form(
+            ferc_form="1",
```
krivard: ?
```suggestion
            ferc_form="6",
```

src/pudl_archiver/archivers/ferc/ferc_online_helpers.py
```diff
@@ -0,0 +1,47 @@
...
+
+from playwright.async_api import async_playwright
+
+from pudl_archiver.archivers.classes import ResourceInfo
+
+logger = logging.getLogger(f"catalystcoop.{__name__}")
+
+
+async def get_resources_for_form(
+    ferc_form: Literal["1", "2", "60", "EQR"],
```
krivard: is ferc6 not intended to go through this helper?

src/pudl_archiver/archivers/ferc/ferc6.py
krivard: is this function still needed?

src/pudl_archiver/archivers/ferc/ferc_online_helpers.py
```diff
@@ -0,0 +1,47 @@
...
+    """Download a resource corresponding to a single year/form combo from ferc online app."""
+    logger.info(f"Downloading the following years for ferc{ferc_form}: {years}")
+    async with async_playwright() as pw:
+        browser = await pw.webkit.launch(headless=True)
+        context = await browser.new_context()
+        page = await context.new_page()
+        await page.goto("https://forms.ferc.gov/")
+
+        # Navigate to form specific page
+        await page.locator(f"#lnkFormData{ferc_form}").click()
```
krivard: This will work for now, but [playwright generally recommends role-based locators whenever possible](https://playwright.dev/python/docs/locators#locating-elements), since those are more robust to code changes that don't otherwise affect the GUI. For this locator that would be

```suggestion
        await page.get_by_role("link", name=f"Form {ferc_form} Data").click()
```

src/pudl_archiver/archivers/ferc/ferc_online_helpers.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # Loop through all years and download
+        resources = []
+        for year in years:
+            logging.info(f"Attempting to download ferc{ferc_form} {year}")
+            async with page.expect_download() as download_info:
+                await page.locator(f"#Content1_lnk{year}f{ferc_form}").click()
```
krivard: here the role-based locator would be


```suggestion
                await page.get_by_role("link", name=str(year)).click()
```

this even works as-is for the links that say something like "2021 Q1 & Q2" -- `name="2021"` will fetch it correctly.

src/pudl_archiver/archivers/ferc/ferc60.py
krivard: is this function still needed?

tests/unit/ferc_archiver_test.py
```diff
@@ -27,85 +26,6 @@
...
-    )
-    mock_session = mocker.AsyncMock()
-    archiver = form_class(mock_session, only_years=only_years)
-    resources = [res async for res in archiver.get_resources()]
-    # don't await these, just check to make sure they have right intention
-    dbfs = [r for r in resources if r.__name__ == "get_year_dbf"]
-    xbrls = [r for r in resources if r.__name__ == "archive_xbrl_for_form"]
-    assert len(dbfs) == num_dbf
-    # Always just one method that archives all xbrl resources
-    assert len(xbrls) == 1
```
krivard, line 106: checking my understanding: we don't need this test anymore, because none of the ferc archivers are actually using the only_years/valid_year machinery at the moment?

if so, we should get rid of the valid_year calls in the archivers themselves

docker-compose.yml
```diff
@@ -0,0 +1,17 @@
...
+  # fsspec archives use GCS for production backend
+  pudl_archiver_fsspec:
+    extends:
+      service: pudl_archiver
+    volumes:
+      - ~/.config/gcloud:/home/appuser/.config/gcloud:ro
```
krivard, line 13: what does this line do?

README.md
```diff
@@ -88,6 +88,13 @@ For datasets that use PUDL source metadata, `pudl-archiver` loads a PUDL
 To develop or test against a local descriptor, set `PUDL_DATAPACKAGE_PATH` to be the
 absolute path to your local `datapackage.json` file.
 
+### Using Docker
+`playwright` doesn't always play right with every environment, so we've provided a
+`docker` based workflow to help working around this. To install the docker environment,
+run `docker compose build`. Then, you will need to set the environment variable
+`PUDL_ARCHIVER_USE_DOCKER=true`. This will tell the pre-commit unit test hook to run
+unit tests through docker.
```
krivard, line 96: yes good!

README.md
```diff
@@ -98,6 +105,14 @@ of these commands, run:
 pudl_archiver --help
 ```
 
+### Using Docker
+When using docker to run the archiver, you prefix each command with `docker compose
+run --rm`, so the above command would become:
```
krivard: is this what you mean, or is the prefix only for certain commands?
```suggestion
When using docker to run the archiver, prefix each command with `docker compose
run --rm`. For example, the above command would become:
```

krivard (changes requested):
Information request: what's going on with the valid_year test?

Required change: 

* double-check the ferc6 archiver; there might be a copy-paste remnant there.

Optional but strongly recommended changes: 

* switch to role-based playwright locators
* remove dead code

Optional: documentation edits for clarity

tests/unit/ferc_archiver_test.py
zschira: The ferc archivers do still use the `valid_years` thing, but I changed exactly how it works and it would break this test. The gist of it is that they all used to `yield` a `get_year_dbf` awaitable for each year of DBF data, which is what this test was looking for, and I changed it so now they just yield a single `get_resources_for_form` awaitable that will handle all years. I've added a new updated function that will check that `get_resources_for_form` gets called with the correct years.

src/pudl_archiver/archivers/ferc/ferc6.py
zschira: No, removed!

src/pudl_archiver/archivers/ferc/ferc_online_helpers.py
zschira: Ah yes it definitely is, fixed!

src/pudl_archiver/archivers/ferc/ferc60.py
zschira: Removed!

docker-compose.yml
zschira: This will mount `gcloud` credentials within the docker container so if you're running the `fsspec` archiver that uses GCS for storage then you will be able to authenticate from within docker. I decided to make this a separate service since you typically don't need `gcloud` credentials to run the normal archiver.

zschira:
@krivard I think this is ready for another look!

src/pudl_archiver/archivers/ferc/ferc2.py
```diff
@@ -1,44 +1,26 @@
...
-                continue
-            for part in [1, 2]:
-                yield self.get_year_dbf(year, part)
-
-        # Get annually partitioned DBF data
-        for year in range(1996, 2022):
-            if not self.valid_year(year):
-                continue
-            yield self.get_year_dbf(year)
+        dbf_years = [year for year in range(1994, 2022) if self.valid_year(year)]
```
zaneselvans: Is the earlier 1991-1993 data we were previously able to archive no longer available?

src/pudl_archiver/archivers/ferc/ferc2.py
zaneselvans: We were never able to parse the split years, so this might mean we can push the coverage back in time now.

src/pudl_archiver/archivers/ferc/ferc_online_helpers.py
```diff
@@ -0,0 +1,47 @@
...
+
+from playwright.async_api import async_playwright
+
+from pudl_archiver.archivers.classes import ResourceInfo
+
+logger = logging.getLogger(f"catalystcoop.{__name__}")
+
+
+async def get_resources_for_form(
+    ferc_form: Literal["1", "2", "6", "60", "EQR"],
```
zaneselvans: Should `EQR` be capitalized here? Won't that result in filenames like `fercEQR-2012.zip`

src/pudl_archiver/archivers/ferc/ferc6.py
```diff
@@ -18,10 +15,13 @@ class Ferc6Archiver(AbstractDatasetArchiver):
...
         """Download FERC 6 resources."""
-        for year in range(2000, 2022):
-            if not self.valid_year(year):
-                continue
-            yield self.get_year_dbf(year)
+        dbf_years = [year for year in range(2000, 2022) if self.valid_year(year)]
+        yield ferc_online_helpers.get_resources_for_form(
+            ferc_form="6",
+            years=dbf_years,
+            partitions_base={"data_format": "DBF"},
```
zaneselvans: You've changed the capitalization on the `data_format` which I think will have downstream impacts.  We should be using all lower case for all of these.

src/pudl_archiver/archivers/ferc/ferc2.py
```diff
@@ -1,44 +1,26 @@
...
+        yield ferc_online_helpers.get_resources_for_form(
+            ferc_form="2",
+            years=dbf_years,
+            partitions_base={"data_format": "DBF"},
```
zaneselvans: Change in `data_format` capitalization -- in general we have standardized on all lower case all the time for the archive partitions.

src/pudl_archiver/archivers/ferc/ferc2.py
zschira: Double checked this and there's actually only 1996-2022 for FERC 2, unfortunately.

src/pudl_archiver/archivers/ferc/ferc2.py
zschira: I've switched to using lowercase everywhere. The `datastore` already converts to lowercase so this shouldn't cause any issues in PUDL

src/pudl_archiver/archivers/ferc/ferc_online_helpers.py
zschira: I went ahead and removed EQR altogether for now, because I noticed that some years are split in parts so we'll need some adjustments to accommodate if we want to archive pre-2013 EQR

zaneselvans: We've got a static archive of those guys that I hand-assembled and it'd be wild if they actually touched that data / I'm not sure I'd trust it... so this seems fine.

https://zenodo.org/records/10086109

krivard (approved):
Nice fixes!

PR #1270, opened 2026-08-01
Propagate default ferceqr dataset in scheduled runs

zaneselvans:
# Overview

- The scheduled EQR update on 2026-08-01 failed because the `inputs.dataset` value only gets the default `ferceqr` value when run `on: workflow_dispatch`. In the `on: schedule` case it remains unset, leading to an invalid script invocation.
- I think the `default` value is really just about pre-filling the workflow dispatch UI, so I added a fallback to `ferceqr` in the environments of the jobs that need datasets in the `run-fsspec-archiver.yml` workflow.
- Ideally there would be a way to specify the default in just a single place, rather than several... I think I can get it down to 2 but the UI default has to be a static string.
- Also added a little more detail to the logging that happens during the run so that the "no change" "updated data" and "new file" cases are all clear in the logs.

# Testing

Since `on: schedule` only works from `main` I tested this with the `gh api` CLI, passing in a null dataset:

```console
gh api repos/catalyst-cooperative/pudl-archiver/actions/workflows/run-fsspec-archiver.yml/dispatches \
  -f ref=fix-ferceqr-default -f 'inputs[dataset]='
```

It successfully fell back to the `ferceqr` default resulting in [this archiving job](https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/30720484780/job/91423337410)

# To-do list

### DIFF
--- a/.github/workflows/run-fsspec-archiver.yml
+++ b/.github/workflows/run-fsspec-archiver.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Build launch notification
         id: launch_message
         env:
-          DATASET: ${{ inputs.dataset }}
+          DATASET: ${{ inputs.dataset || 'ferceqr' }}
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
           WORKFLOW_NAME: ${{ github.workflow }}
         run: |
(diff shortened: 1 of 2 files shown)


krivard (approved):
🚀

PR #1283, opened 2026-08-19
Archive EIA930A data

### LINKED ISSUE #1258: Add EIA 930A to EIA 930 archive
In #1235, we added the reference table to the EIA 930. There's an additional dataset, EIA 930A, which is also reported on the same data page for 2023 and 2024 at the moment. We should add these files into the appropriate zip file years of data in order to be able to extract them into PUDL.

The archiver should:

- [ ] Find all URLs on [this page](https://www.eia.gov/electricity/gridmonitor/about) that reference EIA 930A
- [ ] Strip the year from the link
- [ ] Add these files to the correct year zipfile
- [ ] Be able to handle new years of data inputs flexibly

e-belfer:
# Overview

Closes #1258 

- Add EIA 930A to archive for all years of available data.

- Add EIA 930A files as separate zips
- Update partitions to add form for all files
- Use playwright for EIA 930A because links do not immediately appear using existing request based methods.

# Testing

Run the archiver:

# To-do list

### DIFF
--- a/src/pudl_archiver/archivers/eia/eia930.py
+++ b/src/pudl_archiver/archivers/eia/eia930.py
@@ -1,8 +1,11 @@
 """Download EIA-930 data."""
 
+import re
 from pathlib import Path
+from urllib.parse import urljoin
 
 import pandas as pd
+from playwright.async_api import async_playwright, expect
 
 from pudl_archiver.archivers.classes import (
     AbstractDatasetArchiver,
@@ -16,6 +19,7 @@ FILE_LIST_URL = "https://www.eia.gov/electricity/gridmonitor/sixMonthFiles/EIA93
 REFERENCE_URL = (
     "https://www.eia.gov/electricity/930-content/EIA930_Reference_Tables.xlsx"
 )
+ABOUT_URL = "https://www.eia.gov/electricity/gridmonitor/about"
 
 
 class Eia930Archiver(AbstractDatasetArchiver):
@@ -23,7 +27,7 @@ class Eia930Archiver(AbstractDatasetArchiver):
 
     name = "eia930"
 
-    async def get_file_list(self) -> pd.DataFrame:
+    async def get_eia930_file_list(self) -> pd.DataFrame:
         """Get EIA 930 file list dataframe."""
         return pd.read_csv(FILE_LIST_URL)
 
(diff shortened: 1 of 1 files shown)


src/pudl_archiver/archivers/eia/eia930.py
```diff
@@ -36,28 +40,66 @@ async def get_reference_table(self) -> pd.DataFrame:
...
+        await self.browser.close()
+        await self.playwright.stop()
+
+    async def get_eia930a_files(self) -> dict[str, str]:
+        """Get a dictionary of EIA 930A file download URLs indexed by year."""
+        link_dict = {}
+        self.playwright = await async_playwright().start()
+        self.browser = await self.playwright.webkit.launch()
+
+        link_pattern = re.compile(r"EIA_930A_(\d{4})_with layout.xlsx")
```
zaneselvans, line 54: Huh I wonder if/when they'll add the 2025 data.

src/pudl_archiver/archivers/eia/eia930.py
```diff
@@ -80,6 +122,28 @@ async def get_year_resource(
 
         return ResourceInfo(
             local_path=zip_path,
-            partitions={"half_year": f"{year}half{half_year}"},
+            partitions={"half_year": f"{year}half{half_year}", "form": "eia930"},
```
zaneselvans, line 125: What necessitates adding the `form` as a new partition key? Do we have a mix of `eia930` and `eia930a` partitions in here now?

src/pudl_archiver/archivers/eia/eia930.py
```diff
@@ -80,6 +122,28 @@ async def get_year_resource(
...
+        file_type = url.split(".")[
+            -1
+        ]  # Infer filetype based on url, rather than assigning
+        filename = f"eia930a-{year}.{file_type}"
+        await self.download_and_zip_file(url, filename, zip_path)
+        data_paths_in_archive.add(filename)
+
+        return ResourceInfo(
+            local_path=zip_path,
+            partitions={"year": year, "form": "eia930a"},
```
zaneselvans, line 147: Ah okay. So now we have both annually partitioned `eia930a` and half-year partitioned `eia930` resources in here. I guess that makes sense.

src/pudl_archiver/archivers/eia/eia930.py
zaneselvans, line 34: Ah, I didn't realize we were already archiving this guy -- I think we have a hand-compiled version of this data (copied from the spreadsheet) in PUDL now -- maybe we should integrate this into the EIA-930 transform.

zaneselvans: Running the archiver locally I note that we're zipping up the individual EIA-930A XLSX files, but we're zipping this stand-alone XLSX file.  Do we want to standardize between them all?

zaneselvans (approved):
Nothing blocking -- I ran it and it seems to work fine locally.

One question about whether we want to standardize on either zipping or not zipping all of the individual XLSX files across the archive. Right now we seem to do a mix.

src/pudl_archiver/archivers/eia/eia930.py
zaneselvans: Also, given that we now have bothe year and half_year, I wonder if it makes sense to have either of them associated with this guy as a partition -- or if we should just have the form: reference.

PR #1284, opened 2026-08-19
Add oldest data (1991-1995) back to FERC Form 2 archiver

### LINKED ISSUE #1259: FERC 2 1991-1995 no longer available
With the move to [forms.ferc.gov](https://forms.ferc.gov/), the FERC Form 2 is no longer available for the time period from 1991 to 1995, inclusive.

We [do not extract 1991-1995 in PUDL](https://github.com/catalyst-cooperative/pudl/blob/db8124173b44574f4874b414bdcf740b1a9cf7fe/src/pudl/metadata/sources.py#L604-L608), but we [historically have archived it](https://zenodo.org/records/20489452).

We should decide:

* Whether to seek 1991-1995 data elsewhere on the FERC site
* Whether to publish new FERC 2 archives that do not include 1991-1995
* Whether to (for applications that need coverage for all years) split FERC 2 1991-1995 data into its own zenodo archive versioned separately, or use multiple versions of the full FERC 2 archive, and pull only the subset of files covering 1991-1995 from the older version

e-belfer:
# Overview

Closes #1259 

- New FERC Online archive doesn't contain oldest FERC Form 2 data.

- Add back the old archiving method, just for the oldest years of data.

# Question for reviewer
- Do we want to add the `parts = all` partition back for newer DBF and XBRL data?

# Testing

Run the archiver and verify the oldest data is not deleted: https://github.com/catalyst-cooperative/pudl-archiver/actions/runs/32282792533

# To-do list


zschira (approved):
Looks good! I had just been meaning to do a test run locally before approving, which I've done now and seems to be working fine

e-belfer:
@zschira any opinions about this?

> Do we want to add the parts = all partition back for newer DBF and XBRL data?

zschira:
> Do we want to add the parts = all partition back for newer DBF and XBRL data?

I think it would make sense to add in for the DBF data since that is a meaningful distinction there, but since XBRL data isn't split into multiple partitions like this, it's probably fine to leave out there.

